PR #12, opened 2019-10-31
Updates Kotlin to latest EAP (1.3.60-eap-76) and bumps KTlint

### LINKED ISSUE #8: Kapt not working with the sample
If we add the kotlin kapt plugin 

```apply plugin: 'kotlin-kapt'```

Android studio throws an error 

```e: java.lang.IllegalStateException: Backend Internal error: Exception during code generation```

This might be an issue with the new android gradle plugin. I have tried it outside the project and same error persists.

rafaeltoledo:
Should fix #8

### DIFF
--- a/JetNews/app/build.gradle.kts
+++ b/JetNews/app/build.gradle.kts
@@ -50,7 +50,7 @@ android {
 val ktlint by configurations.creating
 
 dependencies {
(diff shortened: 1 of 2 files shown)


nickbutcher:
Thanks for the contribution 🙌

Issue #11, opened 2019-10-31
Toolbar - navigation button hardly clickable


mice777:
The button in top Toolbar is very hardly clickable. It uses size of drawable for click area, which is 24x24 dp for ic_back.xml. 48dp is probably recommended minimum for buttons.

Ideally the button uses entire toolbar height, and has predefined width, starting from leftmost pixels. The drawable icon may be whatever size it wishes, but clickable area should be large enough. 

See visualization of current clickable area:
![image](https://user-images.githubusercontent.com/3145438/67989513-53255d00-fc33-11e9-8d76-b5e476bb06da.png)

rock3r:
This was fixed by #45 // CC @nickbutcher

JoseAlcerreca:
This has been fixed

PR #13, opened 2019-11-01
Fix toolbar navigation button.

mice777:
Fix for issue #11


mice777:
@googlebot I fixed it.

PR #14, opened 2019-11-01
Fix toolbar navigation button.

mice777:
Fix issue #11

### DIFF
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/Vectors.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/Vectors.kt
@@ -31,7 +31,9 @@ import androidx.ui.res.vectorResource
 fun VectorImageButton(@DrawableRes id: Int, onClick: () -> Unit) {
     Ripple(bounded = false) {
         Clickable(onClick = onClick) {
-            VectorImage(id)
+            Container(expanded = true) {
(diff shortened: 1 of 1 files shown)


JetNews/app/src/main/java/com/example/jetnews/ui/Vectors.kt
```diff
@@ -31,7 +31,9 @@ import androidx.ui.res.vectorResource
 fun VectorImageButton(@DrawableRes id: Int, onClick: () -> Unit) {
     Ripple(bounded = false) {
         Clickable(onClick = onClick) {
-            VectorImage(id)
+            Container(expanded = true) {
```
nickbutcher: This changes the rendering of the app – the icon is pushed to the right.

mice777: I don't get visual difference, the icon is placed exactly on same location, it's just clickable properly in entire area.

I get this:
image

mice777: True, with this fix the icon is pushed a bit to right, but actually it's correctly placed to center of `navigationIcon` in the `TopAppBar`, because without originally it was aligned to left side of it.

nickbutcher:
Fixed by #45

PR #17, opened 2019-11-06
Fix links in Jetnews README.

nickbutcher:


nickbutcher:
Thanks @tiembo, resolved this and also made project links relative so that they work on different branches. PTAL.

Issue #26, opened 2019-11-19
TopAppBar title aligned bottom when system text font size is small


Veronikapj:
Device is Google pixel 3a, Os version is android 10.
System Setting > Display > font size is small 
Top App bar title looks bottom aligned.

![image](https://user-images.githubusercontent.com/360685/69161686-e36a0a00-0b2e-11ea-8a9c-e7d2adb1ce20.png)
![image](https://user-images.githubusercontent.com/360685/69161843-17452f80-0b2f-11ea-810a-d4580b5bc3ef.png)

nickbutcher:
Thank you for filing! However we consider this as working as intended. The Material spec specifies that Top App Bar titles should be bottom aligned (with the title's baseline 20dp from the bottom) so I'll close this as working as intended. Thanks again for the issue.

PR #27, opened 2019-11-19
TopAppBar title alignment to CenterLeft on each screen

Veronikapj:
For fix the issue #26


nickbutcher:
Thank you for the PR! However we consider this as working as intended. The Material spec specifies that Top App Bar titles should be bottom aligned (with the title's baseline 20dp from the bottom) so I'll close this as working as intended. Thanks again for the fix.

PR #32, opened 2019-12-14
Adds UI tests setup

manuelvicnt:


JetNews/app/src/androidTest/java/com/example/jetnews/TestHelper.kt
```diff
@@ -0,0 +1,74 @@
...
+            Surface(children = children)
+        }
+    }
+}
+
+/**
+ * Workarounds, these functions should be removed when UI testing improves
+ */
+
+fun workForComposeToBeIdle() {
```
nickbutcher:
```suggestion
fun waitForComposeToBeIdle() {
```

JetNews/app/src/androidTest/java/com/example/jetnews/TestHelper.kt
```diff
@@ -0,0 +1,74 @@
...
+    selectedTopics.clear()
+}
+
+/**
+ * Helper method that can be used to test Jetnews UI Composables in isolation
+ */
+fun ComposeTestRule.setMaterialContent(children: @Composable() () -> Unit) {
+    setContent {
+        MaterialTheme {
+            Surface(children = children)
```
nickbutcher:
```suggestion
            Surface {
                children()
            }
```

JetNews/app/src/androidTest/java/com/example/jetnews/JetnewsUiTest.kt
```diff
@@ -0,0 +1,53 @@
...
+import androidx.ui.test.doClick
+import androidx.ui.test.findByText
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+
+@MediumTest
+@RunWith(JUnit4::class)
```
mrk-han: Does this not need to be `@RunWith(AndroidJUnit4::class)`? If so, why?

manuelvicnt: This is what the Jetpack Compose team uses internally and we wanted to keep the same format for now. They use it because `AndroidJUnit` runner does not work with `ParametrizedTests`.

Also, removing this line was causing a lot of issues when running the tests from Android Studio, giving a `java.lang.AssertionError: The component does not exist!` error.

JetNews/app/src/androidTest/java/com/example/jetnews/TestHelper.kt
```diff
@@ -0,0 +1,76 @@
...
+    }
+}
+
+/**
+ * Workarounds, these functions should be removed when UI testing improves
+ */
+
+fun waitForComposeToBeIdle() {
+    // Temporary workaround - use waitForIdle in dev04
+    Thread.sleep(500)
```
manuelvicnt: Please notice that this is a **_temporary_** workaround as we weren't able to wait properly with the existing tools. This will go away in the next release, don't ever use `Thread.sleep` in your UI tests.

PR #33, opened 2019-12-14
Adds CI with Github actions

manuelvicnt:


JetNews/.github/workflows/ci.yaml
```diff
@@ -0,0 +1,32 @@
...
+#      - name: Ktlint
+#        run: ./gradlew ktlint
+
+      - name: Build project
+        run: ./gradlew assemble --stacktrace
+
+      - name: Run instrumentation tests
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: 21
```
chrisbanes: Any reason for 21? 28/29 has a lot of performance improvements for the emulator iirc

JetNews/.github/workflows/ci.yaml
```diff
@@ -0,0 +1,32 @@
...
+          arch: x86_64
```
chrisbanes: Nit: x86 will use less RAM

JetNews/.github/workflows/ci.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - master
+  pull_request:
+
+jobs:
+  ci:
+    name: Build + Test
+    runs-on: macOS-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
```
chrisbanes: It's good idea to use `fetch-depth: 1` here, especially for repos which uses contain prebuilts (or have previously contained prebuilts).

JetNews/.github/workflows/ci.yaml
chrisbanes: You use a matrix and run on various API levels too.

JetNews/.github/workflows/ci.yaml
manuelvicnt: `fetch-depth: 1` is the default value but can add it (just in case in changes in the future?). Thanks!

JetNews/.github/workflows/ci.yaml
manuelvicnt: Good idea! Adding the matrix with API 21 and 29! Thanks :) 

I generally like to cover the UI tests with the min API supported

JetNews/.github/workflows/ci.yaml
chrisbanes: Ah, TIL 👍

PR #35, opened 2019-12-16
Adds Gradle cache and Ktlint to Github actions

manuelvicnt:


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+      - uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches/build-cache-1
+          key: ${{ runner.os }}-gradlebuildcache-${{ hashFiles('checksum.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-gradlebuildcache-
+
+      - name: Ktlint
+        run: cd JetNews && ./gradlew ktlint --stacktrace
```
chrisbanes: Could this not be:

`./gradlew JetNews:ktlint`?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Build project
+        run: cd JetNews && ./gradlew assemble --stacktrace
```
chrisbanes: Ditto:

Could this not be:

`./gradlew JetNews:assemble`?

.github/workflows/ci.yaml
manuelvicnt: Tried this before and now again (because maybe I was doing something wrong before) but still getting the following error: `Project 'JetNews' not found in root project 'compose-samples'`

manuelvicnt:
The Gradle cache actions have been taken from @chrisbanes 's Tivi project:
https://github.com/chrisbanes/tivi/blob/master/.github/workflows/build.yml

PR #44, opened 2020-01-07
Removes wait workaround in tests

manuelvicnt:


JetNews/app/src/androidTest/java/com/example/jetnews/TestHelper.kt
```diff
@@ -61,14 +61,9 @@ fun ComposeTestRule.setMaterialContent(children: @Composable() () -> Unit) {
 }
 
 /**
- * Workarounds, these functions should be removed when UI testing improves
+ * Workaround, this function should be removed when UI testing improves
```
JoseAlcerreca: Remove comment?

manuelvicnt: Removed it and added a comment more relevant to the method

PR #45, opened 2020-01-13
Make drawer buttons take up all horizontal space

rock3r:
This tiny PR makes the drawer buttons take up all horizontal space instead of the awkward left-aligned behaviour they have currently.

 Before | After
 --- | ---
 Screenshot_1578933836 | Screenshot_1578933266


rock3r:
I knew you'd appreciate, Nick ;)

PR #66, opened 2020-03-25
Update dependencies

rock3r:
Update AGP 4.1 to alpha 04, Gradle to 6.3-rc1, and Kotlin to 1.3.71. App smoke tested and working.

CC @objcode


JetNews/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Thu Feb 27 20:17:30 CET 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-rc-1-all.zip
```
nickbutcher: 6.3 is out (i.e. no need for RC) and we tend to use `bin` releases.

rock3r: Ah, didn't know it was out. Just went with what AS told me to in the quickfix. Will update and switch to bin. Out of curiosity, any reason why you use `bin`? AFAIK `all` is preferred for usage in IDEs, and AS itself suggests to switch to `all`.

nickbutcher: IIRC it's just because it's a smaller download if you don't need source.

nickbutcher:
Ok, now this one **does** need a rebase.

rock3r:
Done :)

PR #68, opened 2020-04-03
Bump to dev08.

objcode:
Changes:
 * Refactored modifiers to use fluent style throughout
 * Switched to new use Icon and Image composables
 * Switch to using IconButton where appropriate
 * Updated Scaffolding to use IconButton for navigationIcon
 * Removed vector helpers from sample that are now included in the library
 * Update typography of code paraghraphs to expand, and avoid setting background
    on text in code blocks as alpha is not respected – use the background of the    Box
 * Replaced all Container with Box
 * General code cleanup to remove nesting that can be replaced with modifiers
 * Moved Modifiers to *first default argument* to follow style
 * Aligned bullets with baseline of text
 * Sized bullets based on font scaling
 * Removed many named arguments that were not required for readibility
* Added note for bugfix related to font coloring requiring contentColor w/ link to b/143626708
 * Use ColorFilter.tint to semantically tint vectors
 * Continue using Modifier.None for default arugments which is the style in dev08, this is likely to be replaced with `Modifier` in dev09


objcode:
@nickbutcher This is the PR for develop -> master merge for dev08.

PR #70, opened 2020-04-04
Favorites Screen Implementation

pandubaraja:
Hello there 👋,

This PR is intented to implement Favorite Feature in JetNews. So the user can see what his/her favorite articles in `FavoriteScreen`

Because now we can open `ArticleScreen` from `FavoriteScreen` I had to modify `ArticleScreen`  arrow back button implementation because the current implementation is always go to `HomeScreen`, I want it go back to **last screen** instead. So I implement a simple Screen backstack there.

Thank you 🙏 

Preview:
ezgif com-resize


pandubaraja:
@googlebot I fixed it.

PR #82, opened 2020-04-24
Using Kotlin Result class instead of a custom one

patxibocos:
Kotlin JVM comes with a built-in [`Result`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/) class similar to the one introduced recently in this app.
It is a different approach as it is not based on a sealed class, but it provides functions such as fold to run functions either for successful or failed results.
This PR is replacing the custom Result class with the one that comes in the stdlib 🙂


manuelvicnt:
Hi! Thanks for the PR. We decided not to use Kotlin's Result type because it cannot be returned from a function and we didn't want to limit ourselves because of that. 

This problem will get fixed in Kotlin 1.4, meanwhile, we prefer to use our own type.

patxibocos:
> Hi! Thanks for the PR. We decided not to use Kotlin's Result type because it cannot be returned from a function and we didn't want to limit ourselves because of that.
> 
> This problem will get fixed in Kotlin 1.4, meanwhile, we prefer to use our own type.

That makes a lot of sense, I didn't realize about it. Thanks Manuel!

PR #96, opened 2020-06-11
Implimenting Back Propagation from Articles and Interests

NkoroiEric:
In MainActivity override onBackPressed 
`override fun onBackPressed() {
        onBackPressed { super.onBackPressed() }
  }
`
 

This is handles the navigation and its in JetnewsApp:
`fun onBackPressed(f:()->Unit){
    when(JetnewsStatus.currentScreen){
        is Screen.Home -> f()
        else -> navigateTo(Screen.Home)
    }
}
`


nickbutcher:
Thanks so much for the PR but this is now implemented by #98

PR #99, opened 2020-06-16
Switch snackbar auto-dismal to use coroutines.

nickbutcher:


JoseAlcerreca:
I'm more interested in the test than the impl :)

PR #115, opened 2020-08-13
Update to snapshot 6759487.

nickbutcher:


JetNews/app/src/main/java/com/example/jetnews/ui/SwipeToRefresh.kt
```diff
@@ -16,212 +16,77 @@
...
-    val size = with(DensityAmbient.current) { SWIPE_DISTANCE_SIZE.toPx() }
-    // min is below negative to hide
-    val min = -size
-    val max = size * SWIPE_DOWN_OFFSET
-    val dragPosition = state { max }
+    val refreshDistance = with(DensityAmbient.current) { RefreshDistance.toPx() }
+    val state = rememberRefreshSwipeableState(refreshingState)
+
+    // When complete the swipe-to-refresh, kick off the action
+    onCommit(state.value) {
```
nickbutcher, line 52: Not sure about this — can you think of a better way to invoke the refresh action when completing the swipe?

manuelvicnt: Not sure as the API doesn't offer APIs to know when a state has been reached. Feature request?

JetNews/app/src/main/java/com/example/jetnews/ui/SwipeToRefresh.kt
```diff
@@ -16,212 +16,77 @@
...
+    refreshIndicator: @Composable () -> Unit,
+    content: @Composable () -> Unit
 ) {
-    val size = with(DensityAmbient.current) { SWIPE_DISTANCE_SIZE.toPx() }
-    // min is below negative to hide
-    val min = -size
-    val max = size * SWIPE_DOWN_OFFSET
-    val dragPosition = state { max }
+    val refreshDistance = with(DensityAmbient.current) { RefreshDistance.toPx() }
+    val state = rememberRefreshSwipeableState(refreshingState)
```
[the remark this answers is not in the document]
nickbutcher: The issue is that this needs to update if the `refreshingState` argument changes which is why I didn't use `rememberSwipeableState` so that I could guard it's `remember` with that arg.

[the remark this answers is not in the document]
manuelvicnt: Ah! I see, makes sense. Let's leave it like this until we have a better alternative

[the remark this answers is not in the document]
nickbutcher: https://issuetracker.google.com/issues/164113833

JetNews/app/src/main/java/com/example/jetnews/ui/SwipeToRefresh.kt
nickbutcher: https://issuetracker.google.com/issues/164113834

manuelvicnt: Thanks! Should we add a TODO to improve this when that's fixed?

PR #123, opened 2020-08-27
[CI] Sets up a workflow per project

chrisbanes:


.github/workflows/JetNews.yaml
```diff
@@ -0,0 +1,113 @@
+name: JetNews
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'JetNews/**'
```
JoseAlcerreca, line 8: I assume you can't use SAMPLE_PATH here yet?

chrisbanes: Nope, `env` is a runtime thing and these are statically read.

chrisbanes: I'll add some comments once I've got it working

JoseAlcerreca:
Runs JetNews only :+1: Now let's trigger for another project

JoseAlcerreca:
Runs both projects correctly :+1:

JoseAlcerreca (approved):

JoseAlcerreca:
image

PR #126, opened 2020-08-28
Fix grammatical errors

androiddevnotes:


androiddevnotes:
@googlebot I signed it!

JoseAlcerreca:
Oh great, a discussion on the Oxford comma.  😁

This is not an error as the serial comma is optional* in US and considered unnecessary in UK. I would only use it to disambiguate. 


*Obviously it depends on the source as English has no official guidelines

PR #158, opened 2020-09-09
[Jetchat] Fix UserInputText composable overlap

chathil:
When the text in UserInputText is too long it will overlap.

before:

after:


JoseAlcerreca:
Good catch and looks good! I'm not sure if this should be supported in the original TextField though (maybe it is with one of the wrappers). Let me ask around.

JoseAlcerreca:
Created https://issuetracker.google.com/168187755 to move the scrolling behavior from TextField to BaseTextField.

TODO: Add a TODO in code

JoseAlcerreca:
This workaround is too complex for a small benefit as it's already tracked to be fixed. Closing for now but thank you very much!

PR #163, opened 2020-09-14
[Jetsnack] Update to alpha03

nickbutcher:


.github/workflows/Crane.yaml
```diff
@@ -2,8 +2,6 @@ name: Crane
 
 on:
   push:
-    branches:
```
chrisbanes, line 5: Why the removal? (ditto other projects)

Crane/buildSrc/src/main/java/com/example/crane/buildsrc/Dependencies.kt
```diff
@@ -21,7 +21,7 @@ object Versions {
 }
 
 object Libs {
-    const val androidGradlePlugin = "com.android.tools.build:gradle:4.2.0-alpha08"
+    const val androidGradlePlugin = "com.android.tools.build:gradle:4.2.0-alpha09"
```
chrisbanes, line 24: alpha10 (ditto other projs)

chrisbanes (changes requested):

.github/workflows/Crane.yaml
nickbutcher: So that it builds this branch. This is pulled in from @JoseAlcerreca 's [change](https://github.com/android/compose-samples/commit/5f9e1a83007a86680e37bafb26a970fce1c36b29).

Crane/buildSrc/src/main/java/com/example/crane/buildsrc/Dependencies.kt
nickbutcher: This is just for Jetsnack. Each sample is a separate PR.

.github/workflows/Crane.yaml
chrisbanes: You don't need that, the `pr` block means that it will build as soon as you create a PR (and update the PR).

I added this to all of my projects because I find that having actions run on draft branches annoying, and wasteful.

Crane/buildSrc/src/main/java/com/example/crane/buildsrc/Dependencies.kt
chrisbanes: If they'll be done in separate PRs, remove all changes for the other samples?

chrisbanes (approved):
Looks like the build is failing, but changes lgtm

PR #141, opened 2020-09-02
[Jetsnack] Provide default value for `InsetsAmbient`. 

nickbutcher:
Fixes #130

PR #169, opened 2020-09-15
[Jetcaster] Update to Compose SNAPSHOT 6824694

chrisbanes:


florina-muntenescu (approved):

Jetcaster/app/src/main/java/com/example/jetcaster/util/Insets.kt
```diff
@@ -438,9 +438,9 @@ fun Insets.toInnerPadding(
     top: Boolean = true,
     end: Boolean = true,
     bottom: Boolean = true
-): InnerPadding = with(DensityAmbient.current) {
+): PaddingValues = with(DensityAmbient.current) {
```
nickbutcher, line 441: I also renamed this `toPaddingValues` and updated the comments in Jetsnack/Owl. Oh and added the default to the ambient per #141 . Would be good to align (and update the goo.gle/compose-insets gist).

chrisbanes: Yeah agreed. I'll update https://goo.gle/compose-insets next week, and then make sure everything is consistent

PR #176, opened 2020-09-16
Update JetNews to alpha03.

objcode:
Update snackbar in [HomeScreen] to not restart when posts changes but
the error remains false (previously it would flicker due to a restart)

Change workaround in SwipeToRefresh to not crash in alpha03. This was
caused by the behavior change of [onCommit].

Various renames from alpha01-alpha03

### DIFF
--- a/JetNews/.gitignore
+++ b/JetNews/.gitignore
@@ -1,8 +1,15 @@
 *.iml
 .gradle
 /local.properties
-/.idea
+/.idea/caches
+/.idea/libraries
+/.idea/modules.xml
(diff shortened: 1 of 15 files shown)


JetNews/.gitignore
```diff
@@ -1,8 +1,15 @@
 *.iml
 .gradle
 /local.properties
-/.idea
+/.idea/caches
```
JoseAlcerreca, line 4: revert

JoseAlcerreca (approved):

PR #177, opened 2020-09-17
[Jetsurvey] Using alpha03 Snackbar APIs

florina-muntenescu:
Updating Sign In and Sign Up screens to use `Scaffold`

### DIFF
--- a/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInScreen.kt
+++ b/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInScreen.kt
@@ -108,7 +108,7 @@ fun SignIn(onNavigationEvent: (SignInEvent) -> Unit) {
         ErrorSnackbar(
             snackbarHostState = snackbarHostState,
             onDismiss = { snackbarHostState.currentSnackbarData?.dismiss() },
-            modifier = Modifier.align(Alignment.TopCenter)
+            modifier = Modifier.align(Alignment.BottomCenter)
         )
     }
 }
--- a/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInScreen.kt
+++ b/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInScreen.kt
@@ -152,38 +152,36 @@ fun ErrorSnackbar(
     modifier: Modifier = Modifier,
     onDismiss: () -> Unit = { }
 ) {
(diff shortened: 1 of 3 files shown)


Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInScreen.kt
```diff
@@ -122,46 +145,44 @@ fun SignInContent(
...
-                        text = errorText,
-                        style = MaterialTheme.typography.body2
-                    )
-                },
-                action = {
-                    TextButton(
-                        onClick = onDismiss,
-                        contentColor = contentColor()
-                    ) {
+    snackbarHostState.currentSnackbarData?.let { data ->
```
chrisbanes, line 155: Do you need this? The `snackbarHostState` should handle this for you.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInScreen.kt
```diff
@@ -56,36 +57,58 @@ sealed class SignInEvent {
...
-                    onClick = { showSnackbar.value = true },
-                    modifier = Modifier.fillMaxWidth()
-                ) {
-                    Text(text = stringResource(id = R.string.forgot_password))
                 }
             }
         }
+    )
+
+    Stack(modifier = Modifier.fillMaxSize()) {
```
chrisbanes, line 107: Just checking, should this be at the bottom of the screen?

florina-muntenescu: Yes. good catch.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInScreen.kt
florina-muntenescu: `currentSnackbarData` is nullable so when I try to set the `message` I'd end up with a nullable string. 
I could do `snackbarHostState.currentSnackbarData?.message?.let { ... }` in the snackbar text lambda, but what's the point of showing a snackbar without a message?

chrisbanes: Ahhh I see. Yeah remove this, and then use the `data` pass into the `snackbar` lambda below

chrisbanes (approved):

PR #186, opened 2020-09-22
[Jetsurvey] Password validation errors fixed

florina-muntenescu:
* Sign up - password validation not showing
* Sign in - enabling the sign in button when email AND password are valid


JoseAlcerreca (approved):

JoseAlcerreca:
I would target master since you're not targeting alpha04 yet.

PR #187, opened 2020-09-22
Automatically create GitHub release with APKs

chrisbanes:
This new action will trigger when we create a tag. I'll test this on `1.0.0-alpha03` once merged.

### DIFF
--- a/.github/workflows/Release.yml
+++ b/.github/workflows/Release.yml
@@ -1,0 +1,118 @@
+name: GitHub Release with APKs
+
+on:
+  push:
+    # tags:
+    # - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2    
+
+      - name: Copy CI gradle.properties
+        run: mkdir -p ~/.gradle ; cp .github/ci-gradle.properties ~/.gradle/gradle.properties
+
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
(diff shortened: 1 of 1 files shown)


.github/workflows/Release.yml
```diff
@@ -0,0 +1,118 @@
...
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          #tag_name: "${{ github.ref }}"
+          #release_name: ${{ github.ref }}
+          draft: true # TODO update
+          release_name: "Test"
+          prerelease: false
```
JoseAlcerreca, line 38: should this be true until stable? does it do anything apart from showing the label?

chrisbanes: I don't use pre-release personally. It doesn't do anything other than display a small note under the release name.

.github/workflows/Release.yml
```diff
@@ -0,0 +1,117 @@
...
+      - name: Build all projects
+        run: ./scripts/gradlew_recursive.sh assembleDebug
+
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: "${{ github.ref }}"
```
JoseAlcerreca, line 34: Since you just uncommented this, can you reenable the build again to see how this looks like (and if it works)

chrisbanes: It won't work, the ref will be a commit id here. Only for the tag will the ref be correct.

chrisbanes: ...and this definitely works. It's taken from the official action docs: https://github.com/actions/create-release

nickbutcher (approved):

PR #191, opened 2020-09-24
Convert b/ links to public issuetracker URLs

JoseAlcerreca:


Jetcaster/app/src/main/java/com/example/jetcaster/util/ViewModel.kt
```diff
@@ -31,7 +31,8 @@ fun <VM : ViewModel> viewModelProviderFactoryOf(
 ): ViewModelProvider.Factory = SimpleFactory(create)
 
 /**
- * This needs to be a named class currently to workaround a compiler issue: b/163807311
+ * This needs to be a named class currently to workaround a compiler issue:
+ *  https://issuetracker.google.com/163807311
```
nickbutcher, line 35: I get access denied for this. Is it public?

nickbutcher: Doesn't seem so. I'd revert.

JoseAlcerreca: Done

nickbutcher (approved):

PR #197, opened 2020-09-25
[Jetchat] Updates to alpha04 snapshot

JoseAlcerreca:


Jetchat/buildSrc/src/main/java/com/example/compose/jetchat/buildsrc/dependencies.kt
```diff
@@ -21,7 +21,7 @@ object Versions {
 }
```
nickbutcher, line 21: ktlint 0.39.0

chrisbanes (approved):

PR #198, opened 2020-09-25
[Rally] Updates to alpha04 snapshot

JoseAlcerreca:


Rally/buildSrc/src/main/java/com/example/compose/rally/buildsrc/dependencies.kt
```diff
@@ -21,7 +21,7 @@ object Versions {
 }
```
nickbutcher, line 21: Bump ktlint to 0.39.0 as well

nickbutcher (approved):

PR #199, opened 2020-09-25
[Jetcaster] Remove AppCompat + MDC

chrisbanes:
They're unnecessary since we're a Compose-only app

### DIFF
--- a/Jetcaster/app/src/main/res/values/themes.xml
+++ b/Jetcaster/app/src/main/res/values/themes.xml
@@ -16,7 +16,7 @@
   -->
 <resources>
 
-    <style name="Theme.Jetcaster" parent="Theme.MaterialComponents.NoActionBar">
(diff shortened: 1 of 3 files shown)


Jetcaster/app/src/main/res/values/themes.xml
```diff
@@ -16,7 +16,7 @@
   -->
 <resources>
 
-    <style name="Theme.Jetcaster" parent="Theme.MaterialComponents.NoActionBar">
+    <style name="Theme.Jetcaster" parent="android:Theme.Material.NoActionBar">
         <item name="colorPrimary">#ff00ff</item>
```
nickbutcher, line 20: Where do these attrs come from? Shouldn't they have android ns?

chrisbanes: Hmmm, no idea how this worked for me. Fixed.

JoseAlcerreca (approved):

PR #200, opened 2020-09-28
[Crane] Update to snapshot 6860046

manuelvicnt:


Crane/buildSrc/src/main/java/com/example/crane/buildsrc/Dependencies.kt
```diff
@@ -41,9 +41,10 @@ object Libs {
         const val appcompat = "androidx.appcompat:appcompat:1.3.0-alpha02"
```
nickbutcher, line 41: Do you still need AppCompat dep?

PR #204, opened 2020-09-29
[Jetsnack] Use lazy row/columns.

nickbutcher:


Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Feed.kt
```diff
@@ -47,55 +46,62 @@ import com.example.jetsnack.ui.components.JetsnackSurface
...
+        snackCollections,
+        onSnackClick,
+        modifier
+    )
+}
+
+@Composable
+private fun Feed(
+    filters: List<Filter>,
+    snackCollections: List<SnackCollection>,
```
manuelvicnt, line 70: Nit: Feels like passing `snackCollections` as an argument first and then `filters` reads better when calling this function.

manuelvicnt (approved):
Very small nit, LGTM anyway :D

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Feed.kt
chrisbanes: +1

JoseAlcerreca (approved):

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Feed.kt
nickbutcher: Yeah, agreed.

chrisbanes (approved):

PR #217, opened 2020-10-08
[Crane] Update to SNAPSHOT 6891475

manuelvicnt:


Crane/app/src/main/java/androidx/compose/samples/crane/calendar/CalendarActivity.kt
```diff
@@ -42,7 +42,7 @@ fun launchCalendarActivity(context: Context) {
     context.startActivity(intent)
 }
 
-class CalendarActivity : AppCompatActivity() {
```
chrisbanes, line 45: Remove the AppCompat dependency too

manuelvicnt: Thanks! Good catch, just removed it

chrisbanes (approved):

PR #219, opened 2020-10-12
[Rally] Update to SNAPSHOT 6891475

JoseAlcerreca:


Rally/buildSrc/src/main/java/com/example/compose/rally/buildsrc/dependencies.kt
```diff
@@ -89,3 +89,9 @@ object Libs {
         }
     }
 }
+
+object Urls {
+    const val mavenCentralSnapshotRepo = "https://oss.sonatype.org/content/repositories/snapshots/"
```
manuelvicnt, line 94: Nit: not needed if accompanist is not used

manuelvicnt (approved):

manuelvicnt (approved):

PR #221, opened 2020-10-13
Bump JetNews to alpha05

objcode:
Changes:
- Use new produceState API in launchUiStateProducer
- Rename launchUiStateProducer to produceUiState to match naming
- Change launchInComposition to new API LaunchedTask
- Replace Box/Surface with new Box import
- Update modifiers for Text composable to work with new layout rules
- Update to use new AmbientContentColor and AmbientEmphasisLevels

Bugfix:
- Fix size of error case in [HomeScreenErrorAndContent], avoiding
  the swipe to refresh drawing in the incorrect place.

### DIFF
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
@@ -253,15 +253,14 @@ private fun HomeScreenErrorAndContent(
 ) {
     if (posts.data != null) {
         PostList(posts.data, navigateTo, favorites, onToggleFavorite, modifier)
-    } else {
-        Box(Modifier.fillMaxSize()) {
-            if (!posts.hasError) {
-                // if there are no posts, and no error, let the user refresh manually
-                TextButton(onClick = onRefresh, modifier.fillMaxSize()) {
-                    Text("Tap to load content", textAlign = TextAlign.Center)
-                }
-            }
+    } else if (!posts.hasError) {
+        // if there are no posts, and no error, let the user refresh manually
+        TextButton(onClick = onRefresh, Modifier.fillMaxSize()) {
+            Text("Tap to load content", textAlign = TextAlign.Center)
         }
+    } else {
+        // there's currently an error showing, don't show any content
+        Box(modifier.fillMaxSize()) { /* empty screen */ }
     }
 }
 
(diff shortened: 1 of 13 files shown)


JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -253,10 +253,14 @@ private fun HomeScreenErrorAndContent(
 ) {
     if (posts.data != null) {
         PostList(posts.data, navigateTo, favorites, onToggleFavorite, modifier)
-    } else if (!posts.hasError) {
-        // if there are no posts, and no error, let the user refresh manually
-        TextButton(onClick = onRefresh, modifier.fillMaxSize()) {
-            Text("Tap to load content", textAlign = TextAlign.Center)
+    } else {
+        Box(Modifier.fillMaxSize()) {
```
manuelvicnt, line 257: Nit: Is this needed considering the `TextButton` below has the same modifier? 
Also, if `Box` is necessary here, it should use the `modifier` passed as a param instead of `TextButton`

manuelvicnt (approved):
Small nit but LGTM

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
objcode: The flow control is fairly confusing here, let me rework it.

objcode: Simplified it to a three-way branch instead of nested ifs. Please take a look again!

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -253,15 +253,14 @@ private fun HomeScreenErrorAndContent(
...
-        Box(Modifier.fillMaxSize()) {
-            if (!posts.hasError) {
-                // if there are no posts, and no error, let the user refresh manually
-                TextButton(onClick = onRefresh, modifier.fillMaxSize()) {
-                    Text("Tap to load content", textAlign = TextAlign.Center)
-                }
-            }
+    } else if (!posts.hasError) {
+        // if there are no posts, and no error, let the user refresh manually
+        TextButton(onClick = onRefresh, Modifier.fillMaxSize()) {
```
manuelvicnt, line 258: Nit: what about using `modifier` here?

```suggestion
        TextButton(onClick = onRefresh, modifier.fillMaxSize()) {
```

manuelvicnt (approved):

PR #237, opened 2020-10-20
[Crane] Adds Hilt to the project

manuelvicnt:
Changes made:
- Introduces a basic data layer in the app Repository and DataSource classes.
- Uses [Hilt](https://developer.android.com/hilt) as the DI solution to manage dependencies. It works perfectly because the `viewModel()` function used in Compose code gets the default ViewModelFactory created by Hilt with the `@ViewModelInject` annotation.
- Tests also use Hilt as now we heavily rely on ViewModels in our composables.


Crane/README.md
```diff
@@ -32,6 +32,37 @@ implemented using a different Activity will be displayed. In there, you can see
...
+) : ViewModel() { ... }
+
+@Composable
+fun CraneHomeContent(...) {
+    val viewModel: MainViewModel = viewModel()
+    ...
+}
+```
+
+Disclaimer: Passing dependencies that are not available at compile time to a ViewModel
```
chrisbanes, line 56: Add in something like:

> which is sometimes called 'assisted injection'

Crane/README.md
```diff
@@ -32,6 +32,37 @@ implemented using a different Activity will be displayed. In there, you can see
...
+(as it's the case of the [DetailsViewModel](detailsViewModel) that takes as a dependency the name of
```
chrisbanes, line 57: I think this paragraph could be shorter and simplified. It's a bit wordy and long. Let me know if you'd like help re-writing.

Crane/app/src/main/java/androidx/compose/samples/crane/data/DestinationsLocalDataSource.kt
```diff
@@ -0,0 +1,153 @@
...
+ * Annotated with Singleton as the class created a lot of objects.
+ */
+@Singleton
+class DestinationsLocalDataSource @Inject constructor() {
+
+    val craneRestaurants = listOf(
+        ExploreModel(
+            city = NAPLES,
+            description = "1286 Restaurants",
+            imageUrl = "https://images.unsplash.com/photo-1534308983496-4fabb1a015ee?ixlib=rb-1.2.1&auto=format&fit=crop&w=$DEFAULT_IMAGE_WIDTH"
```
chrisbanes, line 34: Don't bother now, but this can be improved so that the width is automatically applied from the actual size of the component. See Owl for an [example](https://github.com/android/compose-samples/blob/7da64ff119990df23bd5a026be8c2757ebbd2a82/Owl/app/src/main/java/com/example/owl/ui/utils/NetworkImage.kt#L87).

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsViewModel.kt
```diff
@@ -0,0 +1,57 @@
...
+import dagger.Module
+import dagger.hilt.InstallIn
+import dagger.hilt.android.components.ActivityRetainedComponent
+
+class DetailsViewModel @AssistedInject constructor(
+    destinationsRepository: DestinationsRepository,
+    @Assisted cityName: String
+) : ViewModel() {
+
+    val cityDetails = destinationsRepository.getDestination(cityName)
```
chrisbanes, line 34: Can you add the type here? Type inference isn't great for public fields/functions.

Is this a LiveData? If so, this the UI state stream should never be `null`. The ViewModel should take care of handling the null case, not the UI.

Crane/app/src/main/java/androidx/compose/samples/crane/home/MainViewModel.kt
```diff
@@ -16,42 +16,61 @@
...
-class MainViewModel : ViewModel() {
+class MainViewModel @ViewModelInject constructor(
+    private val destinationsRepository: DestinationsRepository,
+    @DefaultDispatcher private val defaultDispatcher: CoroutineDispatcher,
+    datesRepository: DatesRepository
+) : ViewModel() {
 
-    val hotels = craneHotels
-    val restaurants = craneRestaurants
+    val hotels = destinationsRepository.hotels
```
chrisbanes, line 41: Ditto: explicit type and make non-null

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
```diff
@@ -59,56 +58,62 @@ fun launchDetailsActivity(context: Context, item: ExploreModel) {
...
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-
-        val args = DetailsActivityArg(
-            name = intent.getStringExtra(DETAILS_NAME)!!,
-            description = intent.getStringExtra(DETAILS_DESCRIPTION)!!,
-            latitude = intent.getStringExtra(DETAILS_LATITUDE)!!,
-            longitude = intent.getStringExtra(DETAILS_LONGITUDE)!!
-        )
+        val args = DetailsActivityArg(intent.getStringExtra(DETAILS_CITY_NAME)!!)
```
chrisbanes, line 77: Nit: throw a better error message if the arg isn't passed in

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
```diff
@@ -46,11 +46,10 @@ import com.google.android.libraries.maps.CameraUpdateFactory
...
 import com.google.android.libraries.maps.model.LatLng
 import com.google.android.libraries.maps.model.MarkerOptions
+import dagger.hilt.android.AndroidEntryPoint
+import javax.inject.Inject
 
-private const val DETAILS_NAME = "DETAILS_NAME"
-private const val DETAILS_DESCRIPTION = "DETAILS_DESCRIPTION"
-private const val DETAILS_LATITUDE = "DETAILS_LATITUDE"
-private const val DETAILS_LONGITUDE = "DETAILS_LONGITUDE"
+private const val DETAILS_CITY_NAME = "DETAILS_CITY_NAME"
```
chrisbanes, line 52: KEY_ARG_DETAILS...

Crane/app/src/main/java/androidx/compose/samples/crane/data/DestinationsLocalDataSource.kt
manuelvicnt: Doing this in a follow-up PR! Thanks, didn't know about that

chrisbanes: Cool, I'm thinking about adding the `AmbientImageLoader` to Accompanist-Coil itself

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsViewModel.kt
manuelvicnt: Done! Thank you

chrisbanes (approved):

PR #241, opened 2020-10-21
SwipeToRefreshLayout bug reproducible example

pedrovgs:
While using a copy-pasted version of ``SwipeToRefreshLayout`` in my app I've noticed it does not work with ``LazyColumnFor`` and ``LazyColumnForIndexed`` components as expected. The swipe down gesture is ignored if the content inside ``SwipeToRefreshLayout`` is a ``LazyColumn``. However, works like a charm with other components.

I've created 3 tests to show you how to reproduce this bug. As you can see, the first test using a ``ScrollableColumn`` is passing, but the last 2 tests using lazy columns are failing. This component was working with Jetpack Compose alpha-04 but after updating to alpha-05 it stopped working.

Honestly, I have no idea how to fix it. Looks like something related to the touch management in the ``LazyColum`` implementation is not forwarding the touch event to the view wrapping the column, but I think there is no way I can fix it without having access to ``LazyColum`` code.

I hope this helps you to reproduce the bug and fix it 😃

### DIFF
--- a/JetNews/app/src/androidTest/java/com/example/jetnews/ui/SwipeToRefreshLayoutTest.kt
+++ b/JetNews/app/src/androidTest/java/com/example/jetnews/ui/SwipeToRefreshLayoutTest.kt
@@ -1,0 +1,147 @@
+/*
+ * Copyright 2020 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


pedrovgs:
You can see the tests reproducing the error failing [here](https://github.com/android/compose-samples/pull/241/checks?check_run_id=1287714734)

objcode:
Thanks!

We're planning on revisiting the gesture APIs soon and I'll leave this open until we have a chance to refactor swipe to refresh to the new APIs and make these pass!

Thanks,
Sean

pedrovgs:
Hey @objcode thanks for the quick response. Is there any workaround we could apply? Without swipe to refresh working, I can't update my app to the latest compose release 😢

JorgeCastilloPrz:
Having the same issue on my side @objcode

objcode:
Update, there's no known workaround on the current alpha, and we expect to patch this soon so that it works as expected.

I'll close this github issue for now, please follow changes on https://issuetracker.google.com/issues/171389491

PR #249, opened 2020-10-23
[JetChat] Update to snapshot 6922857

JoseAlcerreca:


Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/UserInput.kt
```diff
@@ -309,27 +307,23 @@ private fun UserInputSelector(
...
+            disabledContentColor = disabledContentColor
         )
+
         // Send button
         Button(
             modifier = Modifier
                 .padding(horizontal = 16.dp)
                 .preferredHeight(36.dp),
-            elevation = 0.dp,
+            elevation = ButtonConstants.defaultElevation(defaultElevation = 0.dp),
```
manuelvicnt, line 323: Default is `0.dp` already

```suggestion
            elevation = ButtonConstants.defaultElevation(),
```

manuelvicnt (approved):
Small nit, but LGTM anyway

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/UserInput.kt
JoseAlcerreca: The default is the default which is 0 so apparently we can remove the whole thing. Strange that I needed to add it in the first place

PR #262, opened 2020-10-29
[Jetsnack] Add navigation library

JolandaVerhoef:


Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -16,33 +16,35 @@
...
     ProvideDisplayInsets {
         JetsnackTheme {
-            Crossfade(navigator.current) { destination ->
-                when (destination) {
-                    Destination.Home -> Home(actions.selectSnack)
-                    is Destination.SnackDetail -> SnackDetail(
-                        snackId = destination.snackId,
-                        upPress = actions.upPress
+            val navController = rememberNavController()
+            NavHost(navController = navController, startDestination = "home") {
```
JoseAlcerreca, line 36: Let's extract strings to an object or something like HomeSections

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -16,33 +16,35 @@
...
+                composable("home") {
+                    Home(onSnackSelected = { navController.navigate("snackDetail/$it") })
+                }
+                composable(
+                    route = "snackDetail/{snackId}",
+                    arguments = listOf(navArgument("snackId") { type = NavType.LongType })
+                ) { backStackEntry ->
+                    SnackDetail(
+                        snackId = backStackEntry.arguments?.getLong("snackId")
+                            ?: throw IllegalArgumentException("Snack Detail screen needs a snack id"),
```
JoseAlcerreca, line 46: Use requireNotNull (see Owl PR)

msfjarvis:
I've been struggling with a [backstack issue with Navigation](https://github.com/msfjarvis/compose-lobsters/issues/66) that I can reproduce here as well. To reproduce it on Jetsnack:

- Install a build from this branch
- Tap the Profile bottom nav item
- Tap the Home bottom nav item
- Press the back button/perform the back gesture and observe that a recomposition of the Home tab's UI occurs, indicating that a navigation event was triggered.
- Repeat the back button press/gesture, and observe that the app now correctly exits.

I'm not sure if we're all doing things incorrectly or this is a bug with navigation-compose.

I have been unable to repro it on Owl with #264 because it doesn't try to return the user to the start destination on back press.

JolandaVerhoef:
> I've been struggling with a [backstack issue with Navigation](https://github.com/msfjarvis/compose-lobsters/issues/66) that I can reproduce here as well. To reproduce it on Jetsnack:
> 
> * Install a build from this branch
> * Tap the Profile bottom nav item
> * Tap the Home bottom nav item
> * Press the back button/perform the back gesture and observe that a recomposition of the Home tab's UI occurs, indicating that a navigation event was triggered.
> * Repeat the back button press/gesture, and observe that the app now correctly exits.
> 
> I'm not sure if we're all doing things incorrectly or this is a bug with navigation-compose.
> 
> I have been unable to repro it on Owl with #264 because it doesn't try to return the user to the start destination on back press.

I created a bug on Buganizer for this behavior, see https://b.corp.google.com/issues/173181032

msfjarvis:
Public link for non-Googlers: https://issuetracker.google.com/issues/173181032

Thanks for filing the issue!

msfjarvis:
Thanks for the fix @JolandaVerhoef! I've been able to replicate it for my project as well :)

PR #263, opened 2020-10-30
[Jetchat] Moves to Scaffold and Compose drawer content 

JoseAlcerreca:


Jetchat/app/src/main/java/com/example/compose/jetchat/MainViewModel.kt
```diff
@@ -0,0 +1,37 @@
...
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+
+/**
+ * Used to communicate between screens.
+ */
+class MainViewModel : ViewModel() {
+
+    private val _drawerShouldBeOpened = MutableLiveData(false)
```
manuelvicnt, line 28: StateFlow? If LiveData is not used in other places, we could get away with it and remove its dependency

Jetchat/app/src/main/java/com/example/compose/jetchat/NavActivity.kt
```diff
@@ -17,72 +17,98 @@
...
-     * Back closes drawer if open.
-     */
-    override fun onBackPressed() {
-        if (drawerLayout.isDrawerOpen(GravityCompat.START)) {
-            drawerLayout.closeDrawer(GravityCompat.START)
-        } else {
-            super.onBackPressed()
+    @Composable
+    private fun JetchatScaffold(
+        scaffoldState: ScaffoldState,
```
manuelvicnt, line 76: Nit: Give a default value

```suggestion
        scaffoldState: ScaffoldState = rememberScaffoldState(),
```

Jetchat/app/src/main/java/com/example/compose/jetchat/NavActivity.kt
```diff
@@ -17,72 +17,98 @@
...
+                            scaffoldState.drawerState.close()
+                        }
+                    )
+                }
+            ) {
+                val context = ContextAmbient.current
+
+                // Draw the
+                AndroidView({
+                    LayoutInflater.from(context)
```
manuelvicnt, line 103: Can this layout use ViewBinding? eng is really against the pattern of calling `LayoutInflater` inside the `AndroidView` lambda

Jetchat/app/src/main/java/com/example/compose/jetchat/NavActivity.kt
```diff
@@ -17,72 +17,98 @@
...
+                                bundle
+                            )
+                            scaffoldState.drawerState.close()
+                        }
+                    )
+                }
+            ) {
+                val context = ContextAmbient.current
+
+                // Draw the
```
manuelvicnt, line 101: Incomplete comment (:

Jetchat/app/src/main/java/com/example/compose/jetchat/NavActivity.kt
```diff
@@ -17,72 +17,98 @@
...
+                            findNavController(R.id.nav_host_fragment).navigate(
+                                R.id.nav_profile,
+                                bundle
+                            )
+                            scaffoldState.drawerState.close()
+                        }
+                    )
+                }
+            ) {
+                val context = ContextAmbient.current
```
manuelvicnt, line 99: You shouldn't need this `context`, the `AndroidView` lambda should already give you one but try to avoid using it in that way. See comment below

Jetchat/app/src/main/java/com/example/compose/jetchat/NavActivity.kt
```diff
@@ -17,72 +17,98 @@
...
-    /**
-     * Back closes drawer if open.
-     */
-    override fun onBackPressed() {
-        if (drawerLayout.isDrawerOpen(GravityCompat.START)) {
-            drawerLayout.closeDrawer(GravityCompat.START)
-        } else {
-            super.onBackPressed()
+    @Composable
+    private fun JetchatScaffold(
```
manuelvicnt, line 75: Make this a top level function, shouldn't be included in the activity. If you need it to access the `findNavController`, hoist that logic

Jetchat/app/src/main/java/com/example/compose/jetchat/components/JetchatDrawer.kt
```diff
@@ -0,0 +1,173 @@
...
+    ProvideEmphasis(emphasis = AmbientEmphasisLevels.current.medium) {
+        Text(text, style = MaterialTheme.typography.caption, modifier = Modifier.padding(16.dp))
+    }
+}
+
+@Composable
+private fun ChatItem(text: String, selected: Boolean, onChatClicked: () -> Unit) {
+    val background = if (selected) {
+        Modifier
+            .background(MaterialTheme.colors.primary.copy(alpha = 0.08f))
```
manuelvicnt, line 91: Nit: inline with above

Jetchat/app/src/main/java/com/example/compose/jetchat/components/JetchatDrawer.kt
```diff
@@ -0,0 +1,173 @@
...
+    Row(
+        modifier = Modifier
+            .preferredHeight(48.dp)
+            .fillMaxWidth()
+            .padding(horizontal = 8.dp, vertical = 4.dp)
+            .clip(MaterialTheme.shapes.medium)
+            .clickable(onClick = onProfileClicked),
+        verticalAlignment = CenterVertically
+    ) {
+        ProvideEmphasis(emphasis = AmbientEmphasisLevels.current.medium) {
```
manuelvicnt, line 134: `ProvideEmphasis` should be apply only to `Text`

Jetchat/app/src/main/java/com/example/compose/jetchat/components/JetchatDrawer.kt
```diff
@@ -0,0 +1,173 @@
...
+                Image(
+                    imageResource(id = profilePic),
+                    modifier = Modifier
+                        .padding(8.dp)
+                        .preferredSize(24.dp)
+                        .clip(CircleShape),
+                    contentScale = ContentScale.Crop
+                )
+            } else {
+                Spacer(modifier = Modifier.preferredWidth(24.dp).padding(8.dp))
```
manuelvicnt, line 145: This modifier is shared with the `Image` above. I'd assign it to a variable so that the logic is easier to follow

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/BackHandler.kt
```diff
@@ -44,6 +53,12 @@ fun backPressHandler(onBackPressed: () -> Unit, enabled: Boolean = true) {
     // Using onCommit guarantees that failed transactions don't incorrectly toggle the
     // remembered callback.
     onCommit(enabled) {
```
manuelvicnt, line 55: As `highPriority` could also potentially change, it should be included in `onCommit`

```suggestion
    onCommit(enabled, highPriority) {
```

Jetchat/app/src/main/java/com/example/compose/jetchat/profile/ProfileViewModel.kt
```diff
@@ -35,8 +35,8 @@ class ProfileViewModel : ViewModel() {
         _userData.value = if (userId == meProfile.userId) meProfile else colleagueProfile
     }
 
-    private val _userData = MutableLiveData<ProfileScreenState>()
-    val userData: LiveData<ProfileScreenState> = _userData
+    private val _userData = MutableStateFlow<ProfileScreenState?>(null)
```
manuelvicnt, line 38: This pattern is very common when migrating from LD to StateFlow but I'd discourage having a StateFlow of a nullable type. What about creating an empty ProfileScreenState for this?

```suggestion
    private val _userData = MutableStateFlow<ProfileScreenState>(
        ProfileScreenState("", "", "", "", "") // Empty user to begin with
    )
```

The view then should check `userId.isEmpty()` for showing the error state, or even better, have that function as an extension function on `ProfileScreenState` to indicate if it's the object is properly constructed or not.

The benefits of doing this might not be obvious for this example, but it's the right pattern to use. In this way, you avoid that you can emit `null` again for this `StateFlow` and handling nullability when collecting

manuelvicnt (commented):
Thanks for making the changes! This looks great

Added a comment about the usage of `StateFlow`. Thanks!

Jetchat/app/src/main/java/com/example/compose/jetchat/profile/ProfileViewModel.kt
JoseAlcerreca: I think we're trying to avoid null too much, this is an acceptable case for it. An empty screen state doesn't mean the same thing as null. In the view I can do:
```
                        if (userData == null) {
                            ProfileError()
```
and display better data to the user than an empty profile. Or I would have to create an extension `ProfileScreenState.isEmpty()` or something similar which is wasteful.

In a real app you would probably have this data wrapped into a resource sealed class. If there's a big problem with nullable types I would do that first.

JoseAlcerreca: ping!

manuelvicnt: Fine, leave it like this 🤷‍♂️

manuelvicnt (approved):

PR #264, opened 2020-10-30
[Owl] Migrates to navigation-compose

JoseAlcerreca:


Owl/app/src/main/java/com/example/owl/ui/OwlApp.kt
```diff
@@ -17,38 +17,40 @@
...
+                val actions = remember(navController) { Actions(navController) }
+                NavHost(navController = navController, startDestination = Destinations.Onboarding) {
+                    composable(Destinations.Onboarding) {
+                        Onboarding(actions.onboardingComplete)
+                    }
+                    composable(Destinations.Courses) {
+                        Courses(actions.selectCourse)
+                    }
+                    composable(Destinations.CourseDetails + "/{courseId}") { backStackEntry ->
+                        val courseId = backStackEntry.arguments?.getLong("courseId")
```
nickbutcher, line 50: use `require`? https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require.html

alchitry:
This change removes the crossfade effect when switching views. Is there a way to animate changes with navigation-compose?

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
```diff
@@ -16,37 +16,26 @@
...
-    @Parcelize
-    object Courses : Destination()
-
-    @Immutable
-    @Parcelize
-    data class Course(val courseId: Long) : Destination()
+object Destinations {
+    const val Onboarding = "Onboarding_dest_key"
+    const val Courses = "Courses_dest_key"
+    const val CourseDetails = "Courses_dest_key"
```
JolandaVerhoef, line 25: Instead of using the same value for `Courses` and `CourseDetails`, maybe just use the `Courses` constant in the actions?

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
```diff
@@ -16,37 +16,26 @@
...
-    object Onboarding : Destination()
-
-    @Parcelize
-    object Courses : Destination()
-
-    @Immutable
-    @Parcelize
-    data class Course(val courseId: Long) : Destination()
+object Destinations {
+    const val Onboarding = "Onboarding_dest_key"
```
JolandaVerhoef, line 23: I would suggest using simply `"Onboarding"` as a value here. That way, deeplinks will be more readable and sensible.

Owl/app/src/main/java/com/example/owl/ui/OwlApp.kt
```diff
@@ -17,38 +17,40 @@
...
+
+                val actions = remember(navController) { Actions(navController) }
+                NavHost(navController = navController, startDestination = Destinations.Onboarding) {
+                    composable(Destinations.Onboarding) {
+                        Onboarding(actions.onboardingComplete)
+                    }
+                    composable(Destinations.Courses) {
+                        Courses(actions.selectCourse)
+                    }
+                    composable(Destinations.CourseDetails + "/{courseId}") { backStackEntry ->
```
JolandaVerhoef, line 49: The argument string `courseId` is reused twice. Would it make sense to extract this into the NavGraph class somehow?

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
JoseAlcerreca: I used the same by mistake! But since it seems to work, let's just use one.

Owl/app/src/main/java/com/example/owl/ui/OwlApp.kt
JoseAlcerreca: What do you think about this? A little verbose and a lot of curly braces

JolandaVerhoef: Better. I'll create something similar for Jetsnack!

JolandaVerhoef (approved):

Owl/app/src/main/java/com/example/owl/ui/OwlApp.kt
```diff
@@ -17,38 +17,42 @@
...
+                        Onboarding(actions.onboardingComplete)
+                    }
+                    composable(Destinations.Courses) {
+                        Courses(actions.selectCourse)
+                    }
+                    composable(
+                        Destinations.Courses + "/{${Destinations.CoursesArgs.CourseId}}"
+                    ) { backStackEntry ->
+                        val courseId = requireNotNull(backStackEntry.arguments)
+                            .getLong(Destinations.CoursesArgs.CourseId)
```
JolandaVerhoef, line 53: You have to explicitly mention that this argument is of type Long, by passing it to the `composable` function. Otherwise it will be added as a String argument by default.

JolandaVerhoef (changes requested):

Owl/app/src/main/java/com/example/owl/ui/OwlApp.kt
```diff
@@ -17,38 +17,42 @@
...
+                    composable(Destinations.Onboarding) {
+                        Onboarding(actions.onboardingComplete)
+                    }
+                    composable(Destinations.Courses) {
+                        Courses(actions.selectCourse)
+                    }
+                    composable(
+                        Destinations.Courses + "/{${Destinations.CoursesArgs.CourseId}}"
+                    ) { backStackEntry ->
+                        val courseId = requireNotNull(backStackEntry.arguments)
```
JolandaVerhoef, line 52: You want the `requireNotNull` to apply to the argument itself, not to the list of arguments.

```
requireNotNull(backStackEntry.arguments?.getLong(Destinations.CoursesArgs.CourseId))
```

JoseAlcerreca: The nullable object is `arguments`. An invalid long is another story (added some comments)

JoseAlcerreca:
WIP until I add nav tests

JoseAlcerreca:
WIP, I need to make tests a bit more solid and discuss API with eng

Owl/app/src/androidTest/java/com/example/owl/ui/NavigationTest.kt
```diff
@@ -0,0 +1,125 @@
...
+        // The first course should be shown
+        composeTestRule.onNodeWithSubstring(courses.first().name).assertExists()
+    }
+
+    @Test
+    fun coursesToDetail() {
+        // Given the app in the courses screen
+        startActivity(MainDestinations.Courses.route)
+
+        composeTestRule.onRoot().printToLog("jalc")
```
JolandaVerhoef, line 90: This seems like a left-over debug statement. Is it needed?

Owl/app/src/main/java/com/example/owl/ui/courses/FeaturedCourses.kt
```diff
@@ -78,10 +82,15 @@ fun FeaturedCourse(
...
         ConstraintLayout(
-            modifier = Modifier.clickable(
-                onClick = { selectCourse(course.id) }
-            )
+            modifier = Modifier
+                .clickable(
+                    onClick = { selectCourse(course.id) }
+                )
+                .semantics {
+                    accessibilityLabel = featuredString
```
JolandaVerhoef, line 92: By explicitly setting the accessibility label, you are overriding the accessibility label that's generated for this constraint layout by default. Is that your intention? If you only want to have a reference to this in your test, you can use the testTag property instead.

florina-muntenescu:
@JoseAlcerreca GitHub closed this automatically when `dev_alpha07` was merged and deleted. 
Sorry for that. Please reopen this, targeting `dev_alpha08`

PR #268, opened 2020-11-06
[Crane] Updates to snapshot 6953474

manuelvicnt:


Crane/app/src/main/java/androidx/compose/samples/crane/home/LandingScreen.kt
```diff
@@ -32,7 +32,7 @@ private const val SplashWaitTime: Long = 2000
 @Composable
 fun LandingScreen(modifier: Modifier = Modifier, onTimeout: () -> Unit) {
     Box(modifier = modifier.fillMaxSize(), alignment = Alignment.Center) {
-        LaunchedTask {
+        LaunchedEffect(Unit) {
```
objcode, line 35: This will call the `onTimeout` from initial composition even if it changes.

Suggested changes: Either pass onTimeout to LaunchedEffect to restart the timer, or use `rememberUpdatedState` to safely capture `onTimeout`

objcode (commented):
Minor note on `LaunchedEffect` usage.

Crane/app/src/main/java/androidx/compose/samples/crane/home/LandingScreen.kt
manuelvicnt: Sure! Done, thanks

Crane/app/src/main/java/androidx/compose/samples/crane/home/LandingScreen.kt
```diff
@@ -32,7 +32,7 @@ private const val SplashWaitTime: Long = 2000
 @Composable
 fun LandingScreen(modifier: Modifier = Modifier, onTimeout: () -> Unit) {
     Box(modifier = modifier.fillMaxSize(), alignment = Alignment.Center) {
-        LaunchedTask {
+        LaunchedEffect(onTimeout) {
```
nickbutcher, line 35: Is this what we want? To restart the delay if the `onTimeout` action changes? It seemed like `rememberUpdatedState` would be preferable here?

manuelvicnt: That's what @objcode suggested. If we use `rememberUpdatedState` we would also need `LaunchedEffect(Unit)` which makes `rememberUpdatedState` useless?

```
val currentOnTimeout by rememberUpdatedState(onTimeout)
LaunchedEffect(Unit) {
  ...
  currentOnTimeout() // this is equivalent to onTimeout() since this is never going to be called again on recomposition
}
```

nickbutcher (approved):
LGTM — thanks for the changes.

PR #276, opened 2020-11-08
Added Reply Theme

samsruti:
Added new theme: Reply

- List of Email Items
- Detail Screen


nickbutcher:
Thank you very much for the contribution! Unfortunately we will not be able to merge this into the repo as we are not looking to expand the number of samples at this time.

The team is dedicated to maintaining the samples in this repo (we update them every 2 weeks for each new Compose alpha release) and are not looking to expand this work right now. Also, per the contribution guide, it would have been nice to discuss on an issue before submitting this PR.

We're glad that you're excited about Compose and we would love contributions to the existing apps should you like to get involved in the future but adding a whole new sample isn't feasible right now.

PR #266, opened 2020-11-03
[JetNews] Move AppDrawer Composable from JetnewsApp.kt to AppDrawer.kt

takahirom:
Thanks for the many high quality samples!

I'm currently considering how to split files when building an app with Compose.
So I noticed that AppDrawer() used to be called by JetnewsApp() and now it's called by HomeScreen() and InterestsScreen(), and that AppDrawer() stays in JetnewsApp.kt. (https://github.com/android/compose-samples/commit/aafa9ad67bd13695582a4ececf244f42eb28e76c#diff-098e0ebfb55fd1512e2e230dca256aaab9e14bdee57e7317f04dace7061d5803L61 ) 
Therefore, I moved it. 
And there were previews such as PreviewJetnewsApp() and PreviewJetnewsAppDark(), but since it was a function for previewing AppDrawer, I changed it to PreviewAppDrawer().

PR #280, opened 2020-11-09
[JetNews] Bump to alpha07

yrezgui:
I'm currently running into a testing issue both in locally and in GitHub actions


yrezgui:
I'll merge #266 once this PR is merged

JetNews/app/src/main/java/com/example/jetnews/ui/home/PostCardTop.kt
```diff
@@ -54,8 +55,7 @@ fun PostCardTop(post: Post, modifier: Modifier = Modifier) {
         }
         Spacer(Modifier.preferredHeight(16.dp))
 
-        val emphasisLevels = AmbientEmphasisLevels.current
-        ProvideEmphasis(emphasisLevels.high) {
+        Providers(AmbientContentAlpha provides ContentAlpha.high) {
```
florina-muntenescu, line 58: the default is high, so this shouldn't be needed.

JetNews/app/src/main/java/com/example/jetnews/ui/home/PostCardYourNetwork.kt
```diff
@@ -62,8 +63,7 @@ fun PostCardPopular(
                     .fillMaxWidth()
             )
             Column(modifier = Modifier.padding(16.dp)) {
-                val emphasisLevels = AmbientEmphasisLevels.current
-                ProvideEmphasis(emphasisLevels.high) {
+                Providers(AmbientContentAlpha provides ContentAlpha.high) {
```
florina-muntenescu, line 66: the default is high, so this shouldn't be needed. same below

florina-muntenescu (changes requested):
Remove all the high emphasis, as that's the default.

JetNews/app/src/main/java/com/example/jetnews/ui/home/PostCards.kt
```diff
@@ -117,19 +116,17 @@ fun PostCardHistory(post: Post, navigateTo: (Screen) -> Unit) {
             modifier = Modifier.padding(end = 16.dp)
         )
         Column(Modifier.weight(1f)) {
-            ProvideEmphasis(AmbientEmphasisLevels.current.medium) {
```
florina-muntenescu, line 120: this medium emphasis was removed on purpose?

florina-muntenescu (approved):
one comment but otherwise looks good

JetNews/app/src/main/java/com/example/jetnews/ui/home/PostCards.kt
yrezgui: I should create a bug for that. The migration tool on `AmbientEmphasisLevels` set the value automatically to `ContentAlpha.high` and I removed it after cleaning up the project

PR #295, opened 2020-11-27
[Rally] Updates to snapshot 6995125

JoseAlcerreca:


Rally/app/src/androidTest/java/com/example/compose/rally/AnimatingCircleTests.kt
```diff
@@ -40,6 +41,7 @@ import org.junit.Test
  *
  * Note that different systems can produce slightly different screenshots making the test fail.
  */
+@ExperimentalTesting
```
yrezgui, line 44: Why is it required to add it now?

yrezgui (approved):

Rally/app/src/androidTest/java/com/example/compose/rally/AnimatingCircleTests.kt
JoseAlcerreca: It's very unstable right now

PR #315, opened 2020-12-10
Added prefix property to dependencies.kt (jetchat)

LartyHD:


LartyHD:
@googlebot I signed it!

JoseAlcerreca:
Personally, I don't think this helps too much. They don't change often.

chrisbanes:
I agree with @JoseAlcerreca. Thanks for the PR @LartyHD but I think this actually make the dependencies more difficult to read.

PR #319, opened 2020-12-14
[Jetnews] Update to snapshot 7033025

JolandaVerhoef:


JetNews/app/src/main/java/com/example/jetnews/ui/SwipeToRefresh.kt
```diff
@@ -73,3 +87,57 @@ fun SwipeToRefreshLayout(
         }
     }
 }
+
+/**
+ * Temporary workaround for nested scrolling behavior. See b/174756744
```
yrezgui, line 92: Bug isn't public, explain the issue if you can't set it to public

yrezgui (commented):
I would ask reviews from another person, who specifically used the `swipeable` modifier before

chrisbanes (approved):

PR #332, opened 2020-12-17
[All] Update to alpha09

JolandaVerhoef:


JetNews/build.gradle
```diff
@@ -37,7 +37,7 @@ plugins {
 subprojects {
     repositories {
         maven {
-            def snapshot = "7033025"
+            def snapshot = ""
             url  "https://androidx.dev/snapshots/builds/$snapshot/artifacts/repository/"
```
JoseAlcerreca, line 41: This would be a 404, in `main` we don't have the maven entry so I'd just remove it completely.

JoseAlcerreca (approved):

PR #357, opened 2021-01-15
Disable window insets animations on API < 30

chrisbanes:
There are currently issues in WindowInsetsAnimationCompat (likely https://issuetracker.google.com/175391720).
We'll re-enable at a later date once the issues are ironed out.


nickbutcher (approved):

JoseAlcerreca (changes requested):
-1 temporarily as I don't think this fixes the problem

PR #358, opened 2021-01-15
[Jetnews] Migrate to LazyColumn and LazyRow

JolandaVerhoef:


JetNews/app/src/main/java/com/example/jetnews/ui/article/PostContent.kt
```diff
@@ -75,27 +75,39 @@ private val defaultSpacerSize = 16.dp
...
-        PostMetadata(post.metadata)
-        Spacer(Modifier.preferredHeight(24.dp))
-        PostContents(post.paragraphs)
-        Spacer(Modifier.preferredHeight(48.dp))
+        item {
+            PostMetadata(post.metadata)
+            Spacer(Modifier.preferredHeight(24.dp))
+        }
+        items(post.paragraphs) {
+            Paragraph(paragraph = it)
```
yrezgui, line 106: `Paragraph` wasn't there before

JetNews/app/src/main/java/com/example/jetnews/ui/article/PostContent.kt
```diff
@@ -75,27 +75,39 @@ private val defaultSpacerSize = 16.dp
...
+                        lineHeight = 20.sp
+                    )
+                }
+                Spacer(Modifier.preferredHeight(defaultSpacerSize))
             }
-            Spacer(Modifier.preferredHeight(defaultSpacerSize))
         }
-        PostMetadata(post.metadata)
-        Spacer(Modifier.preferredHeight(24.dp))
-        PostContents(post.paragraphs)
```
yrezgui, line 97: `PostContents` seems to have disappeared

yrezgui (changes requested):

JetNews/app/src/main/java/com/example/jetnews/ui/article/PostContent.kt
```diff
@@ -75,27 +75,39 @@ private val defaultSpacerSize = 16.dp
...
-        Spacer(Modifier.preferredHeight(48.dp))
+        item {
+            PostMetadata(post.metadata)
+            Spacer(Modifier.preferredHeight(24.dp))
+        }
+        items(post.paragraphs) {
+            Paragraph(paragraph = it)
+        }
+        item {
+            Spacer(Modifier.preferredHeight(48.dp))
```
yrezgui, line 109: I'm not sure `Spacer` was supposed to be an independant `item`

JetNews/app/src/main/java/com/example/jetnews/ui/article/PostContent.kt
JolandaVerhoef: PostContents used to contain a `forEach`, adding a paragraph composable for each paragraph. I replaced that inline, so I could use `items`. But I forgot to remove the now unused PostContents composable. I added a new commit with that.

JetNews/app/src/main/java/com/example/jetnews/ui/article/PostContent.kt
JolandaVerhoef: How else would I do it? It's one spacer at the bottom of the lazy column. I can't combine it into the previous, as that's an `items` list.

JolandaVerhoef:
@yrezgui thanks for reviewing! I think I addressed all your comments.

JetNews/app/src/main/java/com/example/jetnews/ui/article/PostContent.kt
yrezgui: My bad I didn't understand it that way

yrezgui (approved):

PR #367, opened 2021-01-31
[Jetsnack] Update to alpha11.

nickbutcher:


Jetsnack/app/src/main/java/com/example/jetsnack/ui/theme/Theme.kt
```diff
@@ -85,7 +85,7 @@ fun JetsnackTheme(
     val colors = if (darkTheme) DarkColorPalette else LightColorPalette
 
     val sysUiController = SysUiController.current
-    onCommit(sysUiController, colors.uiBackground) {
+    SideEffect {
```
JoseAlcerreca, line 88: No DisposableEffect(sysUi...)? 
Is this because you don't need to run it multiple times or because SideEffect is smart enough?

JoseAlcerreca (approved):

PR #382, opened 2021-02-10
[Jetsnack] Update to snapshot 7123292.

nickbutcher:


Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/QuantitySelector.kt
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+@file:Suppress("DEPRECATION") // FIXME move to new ConstraintLayout dep once released
```
chrisbanes, line 17: Can we suppress just the function which uses ConstraintLayout? `file` suppressions tends to hide things which we can fix.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/cart/Cart.kt
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+@file:Suppress("DEPRECATION") // FIXME move to new ConstraintLayout dep once released
```
chrisbanes, line 17: Ditto

Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/QuantitySelector.kt
nickbutcher: As discussed, there's also a bunch of deprecated imports so leaving as file suppress. Hopefully we can remove this shortly when `constraintlayout-compose` `alpha02` comes out.

chrisbanes (approved):

PR #384, opened 2021-02-10
[Jetsurvey] Update to alpha12 snapshot

florina-muntenescu:


chrisbanes (approved):

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyScreen.kt
```diff
@@ -174,6 +174,7 @@ private fun TopAppBarTitle(
     )
 }
 
+@Suppress("DEPRECATION") // ConstraintLayout
```
yrezgui, line 177: Is the new constraintlayout artifact not working for you? In case, put a note that this is temporary

yrezgui (changes requested):

chrisbanes:
Let's just commit this for now. We'll fix the ConstraintLayout stuff when we move to alpha12

PR #387, opened 2021-02-11
[All] Update to Compose 1.0.0-alpha12

chrisbanes:


Jetchat/app/src/main/java/com/example/compose/jetchat/NavActivity.kt
```diff
@@ -82,8 +88,23 @@ class NavActivity : AppCompatActivity() {
                             scaffoldState.drawerState.close()
                         }
                     ) {
+                        // Workaround for b/178174718 and b/179181757
```
yrezgui, line 91: Better to put public bug links

Jetchat/app/src/main/java/com/example/compose/jetchat/NavActivity.kt
```diff
@@ -82,8 +88,23 @@ class NavActivity : AppCompatActivity() {
...
+                            if (bindingRef.value == null) {
+                                val binding: ViewBinding =
+                                    ContentMainBinding.inflate(inflater, parent, attachToParent)
+                                bindingRef.value = binding
+                                binding.root.compositionContext =
+                                    currentView.findViewTreeCompositionContext()
+                            }
+                            bindingRef.value as ViewBinding
+                        })
+                        // End of workaround for b/178174718
```
yrezgui, line 106: Same here

yrezgui (approved):

chrisbanes:
> I see that you already bumped compose to alpha12, which requires Kotlin 1.4.30 right? I can't seem to find the Kotlin Android Studio plugin which supports 1.4.30 right now, so I was wondering...

For now, you need to switch to the EAP Kotlin Plugin track:

Screenshot from 2021-02-11 14-47-07

PR #388, opened 2021-02-11
[Jetchat] Fix UI covered by navigation bars

chrisbanes:
Happens when the device is in landscape.

### Screenshots

Screenshot_1613042599
Screenshot_1613042610
Screenshot_1613042615


nickbutcher (approved):

JoseAlcerreca (changes requested):
this added an extra padding to the portrait screenshot

chrisbanes:
> this added an extra padding to the portrait screenshot

🤔 what padding? The left/right values will be 0 in portrait, so the change will no-op.

chrisbanes:
Definitely no extra padding with this PR in portrait. I think you might be seeing the padding on the GitHub comment? If you click the image, it will be shown larger.

PR #398, opened 2021-02-16
[Jetsurvey] Update to beta01 snapshot

florina-muntenescu:


Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyScreen.kt
```diff
@@ -174,43 +177,47 @@ private fun TopAppBarTitle(
     )
 }
 
-@Suppress("DEPRECATION") // ConstraintLayout
-@OptIn(ExperimentalLayout::class)
+@Preview
+@Composable
+fun TAPPreview() {
```
JolandaVerhoef, line 182: Is this name correct? If you're referring to TopAppBar, maybe it should be TABPreview instead? But I'd suggest writing it out anyway.

JolandaVerhoef (approved):
One small nit, looks good otherwise

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyScreen.kt
chrisbanes: It should be `Preview<NameOfComposable>` according to convention (I think there's a lint rule for this?)

Jetsurvey/build.gradle
```diff
@@ -71,4 +72,11 @@ subprojects {
             jvmTarget = "1.8"
         }
     }
+    configurations.configureEach {
```
chrisbanes, line 75: Add a `TODO` comment why this is needed

PR #400, opened 2021-02-16
[Rally] Content description, tabs, and grouping views

JolandaVerhoef:


Rally/app/src/main/java/com/example/compose/rally/ui/overview/OverviewScreen.kt
```diff
@@ -128,7 +130,9 @@ private fun AlertHeader(onClickSeeAll: () -> Unit) {
 @Composable
 private fun AlertItem(message: String) {
     Row(
-        modifier = Modifier.padding(RallyDefaultPadding),
+        modifier = Modifier
+            .padding(RallyDefaultPadding)
+            .semantics(mergeDescendants = true) {},
```
JoseAlcerreca, line 135: Add a comment explaining, also about the difference with  `Modifier.clearAndSetSemantics {}`

Rally/app/src/main/java/com/example/compose/rally/ui/overview/OverviewScreen.kt
```diff
@@ -138,9 +142,11 @@ private fun AlertItem(message: String) {
         )
         IconButton(
             onClick = {},
-            modifier = Modifier.align(Alignment.Top)
+            modifier = Modifier
+                .align(Alignment.Top)
+                .clearAndSetSemantics {}
         ) {
-            Icon(Icons.Filled.Sort, contentDescription = stringResource(id = R.string.sort))
+            Icon(Icons.Filled.Sort, contentDescription = null)
```
JoseAlcerreca, line 149: why wouldn't you want that?

JolandaVerhoef: I was trying to figure out what this button would actually do when you press it. Seemed weird to have a "sort" button on each alert. 

Also; having an identical id for a button on each row in a column is considered a bad practice, as voice access users won't be able to identify them, and focusing each of them would be a repeated effort. In this case, I would use a custom action on the row instead, that describes the action the user can take.

Rally/app/src/main/java/com/example/compose/rally/ui/overview/OverviewScreen.kt
JolandaVerhoef: There's barely any comments in this sample as it is. I'm adding these same properties to all sample apps. Is it really desired to add a comment "Merges the semantics properties of the descendants" to each of these? Or would it be best to have good documentation on these topics and leave it like this.

JoseAlcerreca: The comment shouldn't say what it does but why it does it in this case.

> There's barely any comments in this sample as it is. 
Not an excuse :)

JoseAlcerreca (approved):

PR #402, opened 2021-02-18
[Jetsnack] Correct creation of coroutines within composition.

nickbutcher:


Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
```diff
@@ -189,13 +187,10 @@ private fun JetsnackBottomNavLayout(
...
         }
     }
-    val scope = rememberCoroutineScope()
     selectionFractions.forEachIndexed { index, selectionFraction ->
         val target = if (index == selectedIndex) 1f else 0f
-        if (selectionFraction.targetValue != target) {
-            scope.launch {
-                selectionFraction.animateTo(target, animSpec)
-            }
+        LaunchedEffect(target) {
```
chrisbanes, line 192: Does this work? I think you need to add the `index` as one of the keys, so that you launch multiple concurrently.

nickbutcher: It does indeed work
js_nav

adamp: You might consider using the animSpec as an additional effect key depending on the behavior you want if it changes, finish the current animation in the current spec or change mid-animation.

adamp (approved):

chrisbanes (approved):

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
nickbutcher: Good call, done.

PR #438, opened 2021-03-12
[Owl] Nested navigation

JoseAlcerreca:


Owl/app/src/main/java/com/example/owl/ui/courses/Courses.kt
```diff
@@ -109,11 +93,21 @@ fun CoursesAppBar() {
...
-    SEARCH(R.string.search, R.drawable.ic_search)
+    MY_COURSES(R.string.my_courses, R.drawable.ic_grain, CoursesDestinations.MY_COURSES_ROUTE),
+    FEATURED(R.string.featured, R.drawable.ic_featured, CoursesDestinations.FEATURED_ROUTE),
+    SEARCH(R.string.search, R.drawable.ic_search, CoursesDestinations.SEARCH_COURSES_ROUTE)
+}
+
+/**
+ * Destinations used in the ([OwlApp]).
+ */
+private object CoursesDestinations {
```
nickbutcher, line 109: Could we use inline classes to make this type safe?

Owl/app/src/main/java/com/example/owl/ui/courses/Courses.kt
```diff
@@ -21,66 +21,50 @@ import androidx.annotation.StringRes
...
-            }
-        ) { innerPadding ->
-            val modifier = Modifier.padding(innerPadding)
-            when (selectedTab) {
-                CourseTabs.MY_COURSES -> MyCourses(courses, selectCourse, modifier)
-                CourseTabs.FEATURED -> FeaturedCourses(courses, selectCourse, modifier)
-                CourseTabs.SEARCH -> SearchCourses(topics, modifier)
+fun NavGraphBuilder.courses(
+    onCourseSelected: (Long) -> Unit,
+    onboardingComplete: State<Boolean>,
```
nickbutcher, line 48: Why does this need to be a `State`? Seems like if could just be a `Boolean`?

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
```diff
@@ -41,40 +65,126 @@ object MainDestinations {
...
-        navController = navController,
-        startDestination = startDestination
-    ) {
-        composable(MainDestinations.ONBOARDING_ROUTE) {
-            Onboarding(onboardingComplete = actions.onboardingComplete)
-        }
-        composable(MainDestinations.COURSES_ROUTE) {
-            Courses(selectCourse = actions.selectCourse)
+
+    BlueTheme {
```
nickbutcher, line 82: Seems unexpected to emit UI here. Can the be put back into `OwlApp`?

Owl/app/src/main/java/com/example/owl/ui/OwlApp.kt
```diff
@@ -16,21 +16,15 @@
...
 
 @Composable
-fun OwlApp(backDispatcher: OnBackPressedDispatcher) {
-
-    CompositionLocalProvider(LocalBackDispatcher provides backDispatcher) {
-        ProvideWindowInsets {
-            ProvideImageLoader {
-                NavGraph()
-            }
+fun OwlApp(finishActivity: () -> Unit) {
```
nickbutcher, line 24: Do we need to explicitly plumb this through? If we don't handle explicitly this will bubble up to the containing activity and finish it anyway?

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
```diff
@@ -41,40 +65,126 @@ object MainDestinations {
...
-fun NavGraph(startDestination: String = MainDestinations.ONBOARDING_ROUTE) {
+fun NavGraph(
+    finishActivity: () -> Unit,
+    startDestination: String = MainDestinations.COURSES_ROUTE,
+    showOnboardingInitially: Boolean = true
+) {
     val navController = rememberNavController()
 
+    // Onboarding could be read from shared preferences.
+    val onboardingComplete = remember { mutableStateOf(!showOnboardingInitially) }
```
nickbutcher, line 76: Shouldn't this be `remember(showOnboardingInitially)`?

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
```diff
@@ -41,40 +65,126 @@ object MainDestinations {
...
-            CourseDetails(
-                courseId = arguments.getLong(COURSE_DETAIL_ID_KEY),
-                selectCourse = actions.selectCourse,
-                upPress = actions.upPress
-            )
+    }
+}
+
+@Composable
+fun OwlBottomBar(navController: NavController, tabs: Array<CourseTabs>) {
```
nickbutcher, line 134: Ditto, feel like this should live in `OwlApp`

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
```diff
@@ -41,40 +65,126 @@ object MainDestinations {
...
+                    unselectedContentColor = LocalContentColor.current,
+                    modifier = Modifier.navigationBarsPadding()
+                )
+            }
         }
     }
 }
 
+@Composable
+fun OwlScaffold(
```
nickbutcher, line 169: Ditto, feel like this should live in `OwlApp`

Owl/app/src/main/java/com/example/owl/ui/courses/Courses.kt
JoseAlcerreca: Yeah this was the main blocker for this PR. I _think_ that the problem is that this is not a Composable so passing a boolean is not going to get recomposition magic as we expect.

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
JoseAlcerreca: Why? it has to be state in order to be passed to `courses`

Owl/app/src/main/java/com/example/owl/ui/OwlApp.kt
JoseAlcerreca: Not if the onboarding is being displayed, because the back stack has an additional destination in it at that point.

Owl/app/src/main/java/com/example/owl/ui/courses/Courses.kt
JoseAlcerreca: navigation-compose uses strings, can you expand on what you mean?

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
nickbutcher: Yes but shouldn't it be updated if the `showOnboardingInitially` param changes i.e.
```suggestion
    val onboardingComplete = remember(showOnboardingInitially) { mutableStateOf(!showOnboardingInitially) }
```

Owl/app/src/main/java/com/example/owl/ui/courses/Courses.kt
nickbutcher: I haven't played with inline classes but wondered if they could give us type safety i.e. define a `CourseDestination`type that becomes a raw string at runtime and kept Nav happy. Again, haven't tried it, just dislike stringly typeing.

JoseAlcerreca: It wouldn't add any benefit, as we can't tell navigation to use our class instead of String.

There's more safety coming so let's keep it stringly for now.

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
JoseAlcerreca: ah, I read it as `remember  { showOnboardingInitially }`. I don't think that param is ever going to change, but sure.

Owl/app/src/main/java/com/example/owl/ui/courses/Courses.kt
JoseAlcerreca: Added https://issuetracker.google.com/174783110 to the code

nickbutcher (approved):
Thanks for this and the changes!

PR #441, opened 2021-03-15
[All] Accompanist: dev.chrisbanes. -> com.google.

JoseAlcerreca:
Deprecate Chris Banes.


chrisbanes (approved):
Deprecation: confirmed.

PR #450, opened 2021-03-24
[Jetsnack] Add tests

manuelvicnt:
Adds simple navigation tests to Jetsnack.

Note: We don't have to worry about the animations in the bottom bar, the test framework takes care of that 👌

### DIFF
--- a/Jetsnack/app/src/androidTest/java/com/example/jetsnack/AppTest.kt
+++ b/Jetsnack/app/src/androidTest/java/com/example/jetsnack/AppTest.kt
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


Jetsnack/app/src/androidTest/java/com/example/jetsnack/AppTest.kt
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Test
+    fun app_canNavigateToAllScreens() {
+        // Check app launches at HOME
+        composeTestRule.onNodeWithText("HOME").assertIsDisplayed()
+        composeTestRule.onNodeWithText("Android's picks").assertIsDisplayed()
+
+        // Navigate to Search
+        composeTestRule.onNodeWithTag(HomeSectionTag.SEARCH.toString()).performClick()
```
JoseAlcerreca, line 58: This should be a last resort, did you try to match using hierarchy matchers?

manuelvicnt: Tell me more about that... do you have an example I could take a look at?

JoseAlcerreca: Hierarchical matchers or selectors: https://developer.android.com/jetpack/compose/testing

Even if you can't do this using a known element, you could add a single testTag for the bar and then use the hierarchical matchers to look for the specific buttons inside.

JoseAlcerreca (approved):

nickbutcher (approved):
Thanks!

PR #454, opened 2021-03-26
Add preview placeholder image to Owl app

pedrosax:
Extension of this change: https://github.com/android/compose-samples/pull/453/files


nickbutcher:
Does the preview image have to be in the main source set? What about moving it to debug so that it's not included in the APK?

pedrosax:
> Does the preview image have to be in the main source set? What about moving it to debug so that it's not included in the APK?

Done.

PR #463, opened 2021-04-08
[Owl] Add nested navigation

JoseAlcerreca:
This time using a non-snapshot build. It's similar to #438 which was reverted in #445


Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
```diff
@@ -41,23 +46,54 @@ object MainDestinations {
...
     ) {
         composable(MainDestinations.ONBOARDING_ROUTE) {
-            Onboarding(onboardingComplete = actions.onboardingComplete)
+            // Intercept back in Onboarding: make it finish the activity
+            BackHandler {
+                finishActivity()
+            }
+
+            Onboarding(
+                onboardingComplete = {
```
chrisbanes, line 74: Trailing lambda?

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
```diff
@@ -41,23 +46,54 @@ object MainDestinations {
...
-    val navController = rememberNavController()
+fun NavGraph(
+    modifier: Modifier = Modifier,
+    finishActivity: () -> Unit = {},
+    navController: NavHostController = rememberNavController(),
+    startDestination: String = MainDestinations.COURSES_ROUTE,
+    showOnboardingInitially: Boolean = true
+) {
+    // Onboarding could be read from shared preferences.
+    val onboardingComplete = remember(showOnboardingInitially) {
```
chrisbanes, line 57: This code messed with my head. It's correct, but it took me a minute to work it out.

Owl/app/src/main/java/com/example/owl/ui/OwlApp.kt
```diff
@@ -16,21 +16,98 @@
...
 
-    CompositionLocalProvider(LocalBackDispatcher provides backDispatcher) {
-        ProvideWindowInsets {
-            ProvideImageLoader {
-                NavGraph()
+    val navBackStackEntry by navController.currentBackStackEntryAsState()
+    val currentRoute = navBackStackEntry?.arguments?.getString(KEY_ROUTE)
+        ?: CourseTabs.FEATURED.route
+
+    if (currentRoute in CourseTabs.values().map { it.route }) {
```
chrisbanes, line 74: nit: extract `CourseTabs.values().map { it.route }` to  const or `remember` or something. `values()` creates a new array every call, and then so will `map`

Owl/app/src/main/java/com/example/owl/ui/OwlApp.kt
```diff
@@ -16,21 +16,98 @@
...
+                    unselectedContentColor = LocalContentColor.current,
+                    modifier = Modifier.navigationBarsPadding()
+                )
             }
         }
     }
 }
+
+@Composable
+fun OwlScaffold(
```
chrisbanes, line 102: Just use `Scaffold`? The only thing this is really doing is wrapping the content padding into a modifier.

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
JoseAlcerreca: When the function of the trailing lambda is not obvious like in this case I prefer to keep the name

Owl/app/src/main/java/com/example/owl/ui/OwlApp.kt
JoseAlcerreca: Remembered the array and the map, although I suspect this is less performant overall :)

Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt
JoseAlcerreca: Yeah... we need a better API for this

Owl/app/src/main/java/com/example/owl/ui/OwlApp.kt
chrisbanes: Remembering probably is, but storing the end map in a companion object would be optimal.

chrisbanes (approved):

JoseAlcerreca:
Thanks!

PR #466, opened 2021-04-13
[Jetnews] Migrate to Accompanist SwipeRefresh

chrisbanes:
Replaces the version built-in to Jetnews, with the more modern version available in Accompanist.


JetNews/build.gradle
```diff
@@ -38,6 +38,8 @@ subprojects {
     repositories {
         google()
         jcenter()
+
+        maven { url 'https://oss.sonatype.org/content/repositories/snapshots' }
```
manuelvicnt, line 42: Nit: Add a comment saying this is for the snapshot, so that we can remove when needed

manuelvicnt (approved):

JetNews/app/src/main/java/com/example/jetnews/ui/SwipeToRefresh.kt
```diff
@@ -16,124 +16,6 @@
...
-        //  workaround for a bug in the SwipableState API. Currently, state.value is a duplicated
-        //  source of truth of refreshingState.
-        LaunchedEffect(refreshingState) { state.animateTo(refreshingState) }
-    }
-}
-
 /**
- * Temporary workaround for nested scrolling behavior. There is no default implementation for
- * pull to refresh yet, this nested scroll connection mimics the behavior.
+ * Replaced with: https://google.github.io/accompanist/swiperefresh
```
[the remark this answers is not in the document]
chrisbanes: Because it's linked from lots of Slack comments, blog posts, StackOverflow posts, etc, etc. This is better than a 404 imo.

JetNews/build.gradle
chrisbanes: Done.

Issue #459, opened 2021-04-01
[Jetnews] TopAppBar doesn't match design spec


c5inco:
I believe the version implemented was due to API constraints back in the developer preview launch. With beta out, I believe we can fully realize the design spec that is in place.

This would impact the following screens:

- Home
- Article
- Interests

PR #467, opened 2021-04-13
[Jetnews] TopAppBar doesn't match design spec

angeles-bilbao6:
Added corrections for the TopAppBar in order to look like the most recent design

https://github.com/android/compose-samples/issues/459

**Home**

Before
image

After
image

**Article**

Before 
Article header

After
image

**Interests**

Before
image

After
image


JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -128,20 +135,45 @@ fun ArticleScreen(
         topBar = {
             TopAppBar(
                 title = {
-                    Text(
-                        text = "Published in: ${post.publication?.name}",
-                        style = MaterialTheme.typography.subtitle2,
-                        color = LocalContentColor.current
-                    )
+                    Column(modifier = Modifier.padding(top = 10.dp, bottom = 5.dp)) {
```
JoseAlcerreca, line 138: This column looks redundant

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -128,20 +135,45 @@ fun ArticleScreen(
...
+                        Row(
+                            modifier = Modifier
+                                .fillMaxWidth()
+                                .wrapContentWidth(align = Alignment.CenterHorizontally)
+                                .padding(end = 50.dp)
+                        ) {
+                            Surface(
+                                color = Color(0xff073042),
```
JoseAlcerreca, line 146: This should come from MaterialTheme, and ditto below

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -128,20 +135,45 @@ fun ArticleScreen(
...
+                                shape = CircleShape,
+                                modifier = Modifier.size(36.dp, 36.dp)
+                            ) {
+                                Image(
+                                    painter = painterResource(R.drawable.ic_article_icon),
+                                    contentDescription = "",
```
JoseAlcerreca, line 152: Normally we use null, is there a reason for empty string?

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -182,15 +187,34 @@ fun InterestsScreen(
         },
         topBar = {
             TopAppBar(
-                title = { Text("Interests") },
+                title = {
+                    Text(
+                        text = "Interests",
```
yrezgui, line 192: Can you extract it to the string file?

yrezgui (changes requested):

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
angeles-bilbao6: I removed the column and assign its padding values to the row instead

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
angeles-bilbao6: Changed, I created two new colors in the Material Theme

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
angeles-bilbao6: Changed to null

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
angeles-bilbao6: Changed

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -128,20 +135,43 @@ fun ArticleScreen(
...
+                        ) {
+                            Image(
+                                painter = painterResource(R.drawable.ic_article_icon),
+                                contentDescription = null,
+                                colorFilter = ColorFilter.tint(MaterialTheme.colors.articleIconColor),
+                                modifier = Modifier.padding(6.dp)
+                            )
+                        }
+                        Text(
+                            text = "Published in: \n${post.publication?.name}",
```
yrezgui, line 157: Extract in strings file

JetNews/app/src/main/java/com/example/jetnews/ui/theme/Theme.kt
```diff
@@ -39,9 +42,18 @@ private val DarkThemeColors = darkColors(
     onPrimary = Color.Black,
     secondary = Red300,
     onSecondary = Color.Black,
-    error = Red200
+    error = Red200,
+    onBackground = Color.White
 )
 
+val Colors.articleIconBackground: Color
```
yrezgui, line 49: @JoseAlcerreca @nickbutcher Do we commonly add extension colors to MaterialTheme or just add them as separate variables?

yrezgui (changes requested):

JetNews/app/src/main/java/com/example/jetnews/ui/theme/Theme.kt
JoseAlcerreca: No, we never add extension colors.

These colors should come from the theme. Why are they different? Is there a spec?

JoseAlcerreca (changes requested):

JetNews/app/src/main/java/com/example/jetnews/ui/theme/Theme.kt
angeles-bilbao6: Since these two colors are not considered within the Figma design Color Scheme I was unsure if I should place them with all the other colors, therefore, I looked in the other Jetpack compose examples to see how you were handling these declarations, I found the extension colors in the Jetsurvey theme and that is why I declared them this way. But I can make the changes, should I declare them as regular variables?

angeles-bilbao6: My other option would be to add the icon as a png drawable instead of a vector asset. Let me know which approach should I take.

JoseAlcerreca: My question still applies, why are they different from the Material Theme? isn't it the same red? What spec are you looking at?

angeles-bilbao6: I added the two colors  to create the Article header icon from a vector asset. 
image

None of them where considered in the Color Scheme because they were not actually a part of the theme, they were just the color for the icon and the background.
image

That is why I changed my approach and remove those two colors in my last commit and inserted the icon as a drawable png resource

nickbutcher: Can we use a `VectorDrawable` instead please so that this scales nicely on all screens?

angeles-bilbao6: @nickbutcher I used to have it as a Vector asset, but I was placing the background with a Surface component, but I wasn't sure where to create the color because of this conversation. Else I could also create an image asset which would allow me to set the background there instead of using a surface component . What do you think?

JoseAlcerreca: Oh I thought the "background color" was that red, I didn't see "icon". Yeah this should be a vector asset and the colors should be hardcoded in it, not in compose.

angeles-bilbao6: I just added the vector with the hardcoded colors.

JolandaVerhoef:
There are some conflicts, probably because of the work on Window Insets. Can you rebase and verify your changes work with both navigation types (3-button vs gesture), in landscape mode, in dark mode, and on large screens?

JolandaVerhoef:
Can you please add screenshots of the dark mode / landscape / 3 button nav / large screen use cases? I'd love to see if the app behaves correctly in all of those cases!

angeles-bilbao6:
image
image
image
image


image
image
image
image

image
image
image
image

JolandaVerhoef:
@angeles-bilbao6 I think there's an issue with the light theme, where the status bar icons are not visible. They should show up as dark icons when the theme is set to light. Otherwise, this looks good!

angeles-bilbao6:
@JolandaVerhoef you are right, the issue wasn't happening in the emulator but I was able reproduce it in a real device. 

I just uploaded the fix as you suggested with the dark icons.
image


Let me know if you are ok with this or if I may add another adjustment like a translucid white background to make them more visible

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -131,20 +137,45 @@ fun ArticleScreen(
...
+                            shape = CircleShape,
+                            modifier = Modifier.size(36.dp, 36.dp)
+                        ) {
+                            Image(
+                                painter = painterResource(R.drawable.ic_article_icon),
+                                contentDescription = "",
+                                colorFilter = ColorFilter.tint(Color(0xff32dd84)),
+                                modifier = Modifier.padding(4.dp)
+                            )
+                        }
```
JolandaVerhoef, line 157: Can you create an drawable xml for the whole icon, including the background, and set the colors inside that drawable xml? That makes this code a lot more readable and less magical.

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -131,20 +137,45 @@ fun ArticleScreen(
...
+                            modifier = Modifier
+                                .padding(start = 10.dp)
+                                .weight(1.5f)
+                        )
+                    }
                 },
                 navigationIcon = {
                     IconButton(onClick = onBack) {
-                        Icon(
+                        Image(
```
JolandaVerhoef, line 170: You can use an `Icon` here and set the `tint` parameter

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -131,20 +137,45 @@ fun ArticleScreen(
...
                             imageVector = Icons.Filled.ArrowBack,
-                            contentDescription = stringResource(R.string.cd_navigate_up)
+                            contentDescription = stringResource(R.string.cd_navigate_up),
+                            colorFilter = ColorFilter.tint(MaterialTheme.colors.primary)
                         )
                     }
-                }
+                },
+                elevation = 0.dp,
```
JolandaVerhoef, line 177: Setting this Elevation to 0dp, what happens when you start scrolling? Do you see the text of the screen content cut off? If so, that's not the expected behavior. I'd suggest sticking with the elevation, or better yet, adding elevation on scroll.

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -131,20 +137,45 @@ fun ArticleScreen(
...
+                backgroundColor = MaterialTheme.colors.onPrimary
```
JolandaVerhoef, line 178: `onPrimary` is a color meant for Text that is shown on top of a Primary surface. I don't think that's what you mean here?

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -166,15 +170,37 @@ fun HomeScreen(
...
+                        colorFilter = ColorFilter.tint(MaterialTheme.colors.onBackground),
+                        modifier = Modifier
+                            .fillMaxSize()
+                            .padding(bottom = 4.dp, top = 10.dp)
+                    )
+                },
                 navigationIcon = {
                     IconButton(onClick = { coroutineScope.launch { openDrawer() } }) {
-                        Icon(
+                        Image(
```
JolandaVerhoef, line 185: You can use an Icon here and set the tint

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -166,15 +170,37 @@ fun HomeScreen(
...
+                    ) {
+                        Icon(
+                            imageVector = Icons.Filled.Search,
+                            contentDescription = stringResource(R.string.cd_search)
                         )
                     }
-                }
+                },
+                backgroundColor = MaterialTheme.colors.onPrimary,
+                elevation = 0.dp
```
JolandaVerhoef, line 203: Same comment about elevation and background color

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -175,16 +180,36 @@ fun InterestsScreen(
...
+                    Text(
+                        text = stringResource(R.string.cd_interests),
+                        modifier = Modifier.fillMaxWidth(),
+                        textAlign = TextAlign.Center
+                    )
+                },
                 navigationIcon = {
                     IconButton(onClick = openDrawer) {
-                        Icon(
+                        Image(
```
JolandaVerhoef, line 194: Use Icon

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -175,16 +180,36 @@ fun InterestsScreen(
...
+                        onClick = { /* TODO: Open search */ }
+                    ) {
+                        Icon(
+                            imageVector = Icons.Filled.Search,
+                            contentDescription = stringResource(R.string.cd_search)
+                        )
+                    }
+                },
+                backgroundColor = MaterialTheme.colors.onPrimary,
+                elevation = 0.dp
```
JolandaVerhoef, line 212: Same comment about elevation and background color

JetNews/app/src/main/java/com/example/jetnews/ui/theme/Color.kt
```diff
@@ -23,3 +23,6 @@ val Red300 = Color(0xffea6d7e)
 val Red700 = Color(0xffdd0d3c)
 val Red800 = Color(0xffd00036)
 val Red900 = Color(0xffc20029)
+
+val ArticleIconBackground = Color(0xff073042)
+val ArticleIconColor = Color(0xff32dd84)
```
JolandaVerhoef, line 28: You're not using these variables anymore right?

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
angeles-bilbao6: The behavior is the same when scrolling with 0.dp elevation, the reason why I remove the elevation is because it sets a shadow under the top bar that doesn't match the design


**App with default 4.dp elevation**
image

**Figma design**
image

JetNews/app/src/main/java/com/example/jetnews/ui/theme/Color.kt
angeles-bilbao6: Just removed them

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
JolandaVerhoef: I can help with creating this xml if you want!

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
JolandaVerhoef: What I mean: If there's a shadow showing, the text that scrolls out of view is not cut off.

With shadow vs without shadow
 

Ideally, we'd give the top app bar an elevation when the user scrolls through the content, but hide the elevation when the offset is at 0 dp. But I can imagine that's too much, so just adding the elevation at all times would be an option. Maybe something to discuss with a designer?

angeles-bilbao6: Got it, I just added the elevation only when offset is bigger than 0

image
image

JolandaVerhoef:
Looking through the comments it seems not all have been addressed yet!

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
angeles-bilbao6: I created icon_article_background, let me know if you feel it is the correct implementation

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
angeles-bilbao6: Done

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
angeles-bilbao6: I replaced it with  MaterialTheme.colors.surface

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
angeles-bilbao6: Done

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
angeles-bilbao6: Fixed it

angeles-bilbao6: Added elevation when scrolling

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
angeles-bilbao6: Added elevation on scroll

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
angeles-bilbao6: Changed the background to MaterialTheme.colors.surface, leave the elevation 0 to avoid a gray line to appear dividing the title form the tab menu, dynamically adding the elevation on scroll event not needed for this screen since the tab menu has its own divider

angeles-bilbao6:
Hi @JolandaVerhoef have you had the chance to review the latest changes for this PR?

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -127,25 +133,45 @@ fun ArticleScreen(
...
                         Icon(
                             imageVector = Icons.Filled.ArrowBack,
-                            contentDescription = stringResource(R.string.cd_navigate_up)
+                            contentDescription = stringResource(R.string.cd_navigate_up),
+                            tint = MaterialTheme.colors.primary
                         )
                     }
-                }
+                },
+                elevation = if (scrollState.firstVisibleItemScrollOffset == 0) 0.dp else 4.dp,
```
nickbutcher, line 173: I think the intention here is to increase the elevation if the list has been scrolled? I don't think this code is sufficcient i.e. if you scroll in a new item but it's at 0 offset the elevation will go back to 0.dp. A neat way to fix this might be an ext fun like:

```
val LazyListState.isScrolled: Boolean
  get() = firstVisibleItemIndex > 0 || firstVisibleItemScrollOffset > 0
```

JetNews/app/src/main/java/com/example/jetnews/ui/article/PostContent.kt
```diff
@@ -78,8 +79,9 @@ import com.example.jetnews.ui.theme.JetnewsTheme
 private val defaultSpacerSize = 16.dp
 
 @Composable
-fun PostContent(post: Post, modifier: Modifier = Modifier) {
+fun PostContent(post: Post, state: LazyListState, modifier: Modifier = Modifier) {
```
nickbutcher, line 82: What about adding a default value for `state` i.e. `state: LazyListState = rememberLazyListState()`?

Would make hoisting the state optional and clean up Preview usage.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
nickbutcher: I still see `Image`?

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -162,33 +168,58 @@ fun HomeScreen(
...
+                        onClick = { /* TODO: Open search */ }
+                    ) {
+                        Icon(
+                            imageVector = Icons.Filled.Search,
+                            contentDescription = stringResource(R.string.cd_search)
+                        )
+                    }
+                },
+                backgroundColor = MaterialTheme.colors.surface,
+                elevation = if (scrollState.firstVisibleItemScrollOffset == 0) 0.dp else 4.dp
```
nickbutcher, line 207: Ditto elevation comment.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -275,6 +307,7 @@ private fun HomeScreenErrorAndContent(
 @Composable
 private fun PostList(
     posts: List<Post>,
+    state: LazyListState,
```
nickbutcher, line 310: DItto comment about overing a default.

JetNews/app/src/main/res/drawable/ic_article_icon.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="24dp" android:tint="#3DDC84"
```
nickbutcher, line 1: Needs copyright header.

JetNews/app/src/main/res/drawable/icon_article_background.xml
```diff
@@ -0,0 +1,14 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
nickbutcher, line 1: Copyright header.

nickbutcher:
Thanks for the changes, a few more comments and then I think we can land this.

JoseAlcerreca:
Some emulators on GHA are acting up
https://github.com/ReactiveCircus/android-emulator-runner/issues/168

so if 2/3 emulator runs pass, consider the PR green please.

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
angeles-bilbao6: Done

JetNews/app/src/main/java/com/example/jetnews/ui/article/PostContent.kt
angeles-bilbao6: Done

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
angeles-bilbao6: Done

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
angeles-bilbao6: Done

JetNews/app/src/main/res/drawable/ic_article_icon.xml
angeles-bilbao6: Removed file, it was never used

JetNews/app/src/main/res/drawable/icon_article_background.xml
angeles-bilbao6: Done

angeles-bilbao6:
Hi @nickbutcher I have already addressed your latest comments, let me know if you are ok with the fixes.

angeles-bilbao6:
Hi @nickbutcherhave you had the chance to review the latest changes?

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
JoseAlcerreca: The Android logo is not centered in the circle, I checked the mocks and it should be smaller.

The mock's logo has weird antennae though.

JoseAlcerreca: image
vs
image

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
JoseAlcerreca: +1, please replace with icon

From IconButton:
> content should typically be an Icon, using an icon from androidx.compose.material.icons.Icons. If using a custom icon, note that the typical size for the internal icon is 24 x 24 dp.

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
JoseAlcerreca: Please rename the file "LazyListUtils.kt" as it's not holding any state.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
JoseAlcerreca: I don't think this was addressed in PostList

JetNews/app/src/main/res/drawable/icon_article_background.xml
JoseAlcerreca: This file is missing `<?xml version="1.0" encoding="utf-8"?>`

JoseAlcerreca:
Went through all the comments resolving some and commented on the ones that were not addressed.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
angeles-bilbao6: You are right, I just fixed it

JetNews/app/src/main/res/drawable/icon_article_background.xml
angeles-bilbao6: Added it

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
angeles-bilbao6: Done

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
angeles-bilbao6: I just made the change

angeles-bilbao6:
Hi @JoseAlcerreca I just addressed your last comments

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -187,10 +185,10 @@ fun HomeScreen(
                 },
                 navigationIcon = {
                     IconButton(onClick = { coroutineScope.launch { openDrawer() } }) {
-                        Image(
+                        Icon(
                             painter = painterResource(R.drawable.ic_jetnews_logo),
                             contentDescription = stringResource(R.string.cd_open_navigation_drawer),
-                            colorFilter = ColorFilter.tint(MaterialTheme.colors.primary)
+                            tint = (MaterialTheme.colors.primary)
```
JoseAlcerreca, line 191:
```suggestion
                            tint = MaterialTheme.colors.primary
```

JoseAlcerreca (approved):

JoseAlcerreca:
Woohoo, thanks @angeles-bilbao6!

PR #470, opened 2021-04-15
Migrate to rememberCoilPainter()

chrisbanes:
New in `Accompanist 0.7.2-SNAPSHOT`

### DIFF
--- a/Crane/app/src/main/java/androidx/compose/samples/crane/base/ExploreSection.kt
+++ b/Crane/app/src/main/java/androidx/compose/samples/crane/base/ExploreSection.kt
@@ -47,7 +47,8 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.unit.dp
-import com.google.accompanist.coil.CoilImage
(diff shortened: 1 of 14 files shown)


Crane/app/src/main/java/androidx/compose/samples/crane/base/ExploreSection.kt
```diff
@@ -93,21 +94,27 @@ private fun ExploreItem(
...
+                    painter = painter,
+                    contentDescription = null,
+                    contentScale = ContentScale.Crop,
+                )
+
+                if (painter.loadState == ImageLoadState.Loading) {
+                    Image(
+                        painter = painterResource(id = R.drawable.ic_crane_logo),
+                        contentDescription = null,
+                        modifier = Modifier
```
manuelvicnt, line 112: Missing the `Modifier.fillMaxSize` that was previously in the `Box` of `loading`. Is it not needed anymore?

chrisbanes: It's makes no difference really, as a `ExploreImageContainer` is set to 60.dp, and the images are all larger than that... but I added an explicit `fillMaxSize()` anyway

manuelvicnt (approved):

Issue #460, opened 2021-04-01
[Jetsurvey] UI polish bugs


c5inco:
Polish bugs identified by the Android Developer UX team

Login/signup screen

- [ ] TextField corners on Login/Signup should be more round (12.dp versus 4.dp)
- [x] Spacing above "Sign in" text should be 64.dp
- [x] Space between "Sign in" and email text field should be 20.dp
- [x] Space between Continue button and "or" should be 28.dp (text baseline aligned)

Base elements question polish

- [x] Close X mark should be on right
- [x] Close X mark should be vertically centered with "<x> of <y>" text
- [x] Close X mark edge should align with end of progress bar
- [x] Spacing between progress bar and question should be 32.dp
- [x] Options content should all be left-aligned with question text
- [x] Next/Previous buttons should be taller to 48.dp
- [x] Missing elevation divider above Next/Previous buttons
- [x] Vertical padding of description should be 24.dp
- [x] Selected option should primary style
- [x] Space between options should be 16.dp
- [x] No motion for change in progress indicator
- [x] Progress bar background missing in dark theme

Super hero / movie questions

- [ ] Images missing for super hero and movie questions
- [x] Space between "Select date" and date button should be 24.dp

Pick a date question

- [x] Date button should be exposed dropdown
- [x] Date picker says "Selected date" on open instead of the selected date
- [x] Selected date should be value of exposed dropdown

Scale selfies question

- [x] Slider design implementation not according to spec
- [x] On page load, mid-point selected by default for slider

Photo question

- [x] Taken photo should be clipped, keep size of photo frame the same
- [ ] Progress indicator for loading taken photo might need polish

PR #477, opened 2021-04-21
[Jetsurvey] Photo was not showing with rememberCoilPainter

angeles-bilbao6:
Preview image stop showing after migrating to rememberCoilPainter

https://github.com/android/compose-samples/issues/460

Taken photo should be clipped, keep size of photo frame the same

Before
image

After
image

### DIFF
--- a/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
+++ b/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
@@ -26,8 +26,10 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.aspectRatio
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.wrapContentSize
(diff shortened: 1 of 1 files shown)


chrisbanes:
So the reason this is happening is because the main UI is in an `LazyColumn`, so each item implicitly has no max height. This means that using `fillMaxHeight()` has no effect (as there is no max size) and results in the item having a height of 0. 

`rememberCoilPainter()` sees this size and skips loading due to the 0 height. Working on a fix now

JoseAlcerreca:
@angeles-bilbao6 please rebase this PR to include the fix in #478

angeles-bilbao6:
@JoseAlcerreca Done, I kept the max height value because else way the photo won't respect the design expected height.

chrisbanes:
Hmmm, where did the value of `302.dp` come from? I don't think this is right, and will only 'work' for specific display sizes. If we want to force an aspect ratio, we can use `Modifier.apsectRatio(...)`.

angeles-bilbao6:
@chrisbanes You are right, I just made the change

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
```diff
@@ -346,8 +347,8 @@ private fun PhotoQuestion(
                     contentDescription = null,
                     modifier = Modifier
                         .fillMaxWidth()
-                        // We set a minimum height, so that Coil has a height to initially use
-                        .heightIn(min = 96.dp),
+                        .heightIn(96.dp)
+                        .aspectRatio(1.22f)
```
chrisbanes, line 351: `1.22f` is an interesting aspect ratio. I'd probably use `aspectRatio(4 / 3f)` which is more typical and common.

angeles-bilbao6: Figma design has an image size of 371 / 302 px that is why I chose that aspect ratio, but I will update to use 4/3f

chrisbanes (approved):

JoseAlcerreca:
Thanks!

PR #479, opened 2021-04-22
[JetNews] Implement navigation library

JolandaVerhoef:
- [x] Implement navigation library and create `NavGraph.kt`
- [x] Pass specific navigation actions down the composable chain, instead of the generic `navigateTo` method
- [x] Move App Drawer to top level composable so it can directly interact with the NavController.
- [x] Update @Preview instances to reflect the moved app drawer


JetNews/app/src/main/java/com/example/jetnews/ui/AppDrawer.kt
```diff
@@ -44,34 +45,43 @@ import androidx.compose.ui.res.painterResource
...
+import androidx.navigation.compose.currentBackStackEntryAsState
+import androidx.navigation.compose.navigate
+import androidx.navigation.compose.rememberNavController
 import com.example.jetnews.R
 
 @Composable
 fun AppDrawer(
-    navigateTo: (Screen) -> Unit,
-    currentScreen: Screen,
+    navController: NavController,
```
JoseAlcerreca, line 57: I think (we need more experience on this) that passing navController down is a smell.

Ideally, everything related to navigation is encapsulated in the NavHost file.

I think this can be reverted and the navController stuff hoisted

JetNews/app/src/main/java/com/example/jetnews/ui/NavGraph.kt
```diff
@@ -0,0 +1,90 @@
...
+    const val ARTICLE_ROUTE = "post"
+    const val ARTICLE_ID_KEY = "postId"
+}
+
+@Composable
+fun NavGraph(
+    appContainer: AppContainer,
+    navController: NavHostController = rememberNavController(),
+    startDestination: String = MainDestinations.HOME_ROUTE,
+    scaffoldState: ScaffoldState
```
JoseAlcerreca, line 48: Nit: Move param to second position

JetNews/app/src/main/java/com/example/jetnews/ui/NavGraph.kt
```diff
@@ -0,0 +1,90 @@
...
+        composable(MainDestinations.INTERESTS_ROUTE) {
+            InterestsScreen(
+                interestsRepository = appContainer.interestsRepository,
+                scaffoldState = scaffoldState
+            )
+        }
+        composable("${MainDestinations.ARTICLE_ROUTE}/{$ARTICLE_ID_KEY}") { backStackEntry ->
+            val arguments = requireNotNull(backStackEntry.arguments)
+            ArticleScreen(
+                postId = arguments.getString(ARTICLE_ID_KEY)!!,
```
JoseAlcerreca, line 72: Instead of crashing here and on line 70, default to null or an empty string and deal with it inside ArticleScreen showing an error message.

JetNews/app/src/main/java/com/example/jetnews/ui/NavGraph.kt
```diff
@@ -0,0 +1,90 @@
...
+                onBack = actions.upPress,
+                postsRepository = appContainer.postsRepository
```
JoseAlcerreca, line 74: Passing a whole repository to a small composable seems like an antipattern. I know this is done for simplicity but I'd either use a simple ArticleViewModel inside ArticleScreen or grab data from a singleton object adding the appropriate "this is done for simplicity" comments.

JoseAlcerreca: Looking at the code it seems like this was already implemented this way. We can deal with it later.

JetNews/app/src/main/java/com/example/jetnews/ui/NavGraph.kt
```diff
@@ -0,0 +1,90 @@
...
+ */
+object MainDestinations {
+    const val HOME_ROUTE = "home"
+    const val INTERESTS_ROUTE = "interests"
+    const val ARTICLE_ROUTE = "post"
+    const val ARTICLE_ID_KEY = "postId"
+}
+
+@Composable
+fun NavGraph(
```
JoseAlcerreca, line 44: Rename to JetNewsNavGraph to clarify this is not toolkit?

JetNews/README.md
```diff
@@ -22,8 +22,8 @@ screen uses a navigation drawer.
 
 Package [`com.example.jetnews.ui`][1]
 
-[`JetnewsApp.kt`][2] arranges the different screens in the `NavDrawerLayout`. It also implements a simple
-navigation pattern.
+[`JetnewsApp.kt`][2] arranges the different screens in the `NavDrawerLayout`. It also implements 
+the navigation library to navigate from one screen to the next.
```
JoseAlcerreca, line 26: This file doesn't implement that, it's in NavGraph.kt. I'd add another line with it and rephrase as it's not "implementing the library".

JoseAlcerreca (commented):
Looking good!

JetNews/app/src/main/java/com/example/jetnews/ui/AppDrawer.kt
JolandaVerhoef: Done

JetNews/app/src/main/java/com/example/jetnews/ui/NavGraph.kt
JolandaVerhoef: Done, and default parameter added as well. Not sure about best practices for adding defaults.

JetNews/app/src/main/java/com/example/jetnews/ui/NavGraph.kt
JolandaVerhoef: Done. Not sure I like it that much, as you end up having to take care of the possible null value in that whole screen. Please let me know what you think about my solution.

JetNews/app/src/main/java/com/example/jetnews/ui/NavGraph.kt
JolandaVerhoef: I'll put it on the list of possible improvements for Jetnews, and will work on it if time allows. Thanks for noticing it!

JetNews/app/src/main/java/com/example/jetnews/ui/NavGraph.kt
JolandaVerhoef: Done

JetNews/README.md
JolandaVerhoef: Done, can you check the rewording if this is more clear?

JolandaVerhoef:
@JoseAlcerreca would you be able to take another look at this? To prevent confusion, I've opted to pass only the `openDrawer()` callback to the screens, instead of the full `scaffoldState`.

JoseAlcerreca (approved):

Issue #487, opened 2021-04-27
[Jetsnack] UI polish bugs


c5inco:
Polish bugs identified by the Android Developer UX team

Home
- [x] Text in "Add to cart" button not centered
- [x] Empty state for Profile should use placeholder graphic
- [x] Second row of cards should use pink gradients (if possible make them alternate between blue and pink)
- [x] Checkout button ripple should be across whole button
- [x] Filter chips "pressed" and "selected" state is not to spec
- [x] Elements should be left aligned with 24.dp margin/guide
- [x] Incorrect gradient borders on chips in dark theme

Cart
- [x] The quantity + and - buttons should have same click state as filter chips

Item details
- [x] Item page background gradient should be Gradient 2-2 (see spec)
- [x] Padding between Details and Ingredients should be 16.dp
- [ ] Modifying quantity to zero should not change padding

Search
- [x] Incorrect background gradients used in dark theme

PR #488, opened 2021-04-28
[Jetsnack] UI Polish

angeles-bilbao6:
https://github.com/android/compose-samples/issues/487

Home

-  Text in "Add to cart" button not centered

Before
image

After
image

-  Empty state for Profile should use placeholder graphic

Before
image

After
image

-  Second row of cards should use pink gradients (if possible make them alternate between blue and pink)
-  Elements should be left aligned with 24.dp margin/guide

Before
image

After
image

-  Checkout button ripple should be across whole button -- FIXED

**Item details**

-  Item page background gradient should be Gradient 2-2 (see spec) -- Spec has a gradient called tornado1
-  Padding between Details and Ingredients should be 16.dp

Before
image

After
image

Search

 Incorrect background gradients used in dark theme

Before
image

After
image


nickbutcher (approved):
Thanks for the fixes!

PR #529, opened 2021-05-15
[Jetsurvey] Fix Email and Password TextFields

ashar-7:
- Add `keyboardType` and `singleLine` to Text Fields (Fix for multiple lines being input through the keyboard)
- Replace keyboard action `onDone` with `onAny` in case a different Ime Action is used (Fixes focus not moving from Email field to Password field on Next action)


Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInSignUp.kt
```diff
@@ -152,12 +153,16 @@ fun Email(
...
         textStyle = MaterialTheme.typography.body2,
         isError = emailState.showErrors(),
-        keyboardOptions = KeyboardOptions.Default.copy(imeAction = imeAction),
+        keyboardOptions = KeyboardOptions.Default.copy(
+            imeAction = imeAction,
+            keyboardType = KeyboardType.Email
+        ),
         keyboardActions = KeyboardActions(
-            onDone = {
+            onAny = {
```
IanGClifton, line 161: I don't think we'd want to use onAny here because that means onImeAction will also be called when next or previous (which should move focus) are triggered. Is there a case where onDone isn't enough?

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInSignUp.kt
```diff
@@ -218,12 +223,16 @@ fun Password(
...
         },
         isError = passwordState.showErrors(),
-        keyboardOptions = KeyboardOptions.Default.copy(imeAction = imeAction),
+        keyboardOptions = KeyboardOptions.Default.copy(
+            imeAction = imeAction,
+            keyboardType = KeyboardType.Password
+        ),
         keyboardActions = KeyboardActions(
-            onDone = {
+            onAny = {
```
IanGClifton, line 231: Same 'onAny' comment here.

IanGClifton (commented):
I just have a question for the onDone -> onAny change but everything else looks good!

PR #539, opened 2021-05-19
Jetnews fewer posts

angeles-bilbao6:
Added show fewer stories popup.

Functionality not yet implemented

image


JetNews/app/src/main/java/com/example/jetnews/ui/home/PostCards.kt
```diff
@@ -128,32 +133,75 @@ fun PostCardSimple(
...
+                    )
+                }
+                PostTitle(post = post)
+                AuthorAndReadTime(
+                    post = post,
+                    modifier = Modifier.padding(top = 4.dp)
+                )
+            }
+        }
+        Row() {
```
JolandaVerhoef, line 163: Using two Rows inside another Row might not be the best way to go. Consider making the Icon 48x48 (with padding) and just put it inside the top level row together with the rest. The inner Clickable should override the outer one, right?

JetNews/app/src/main/java/com/example/jetnews/ui/home/PostCards.kt
```diff
@@ -128,32 +133,75 @@ fun PostCardSimple(
...
-            Icon(
-                imageVector = Icons.Filled.MoreVert,
-                contentDescription = stringResource(R.string.cd_more_actions)
+        if (openDialog) {
+
+            AlertDialog(
+                modifier = Modifier.padding(20.dp),
+                onDismissRequest = {
+                    openDialog = false
+                },
```
JolandaVerhoef, line 179: nit: this can be on one line

JetNews/app/src/main/java/com/example/jetnews/ui/home/PostCards.kt
```diff
@@ -128,32 +133,75 @@ fun PostCardSimple(
...
-                post = post,
-                modifier = Modifier.padding(top = 4.dp)
-            )
         }
-        CompositionLocalProvider(LocalContentAlpha provides ContentAlpha.medium) {
-            Icon(
-                imageVector = Icons.Filled.MoreVert,
-                contentDescription = stringResource(R.string.cd_more_actions)
+        if (openDialog) {
+
```
JolandaVerhoef, line 174: nit: remove empty line

JetNews/app/src/main/java/com/example/jetnews/ui/home/PostCards.kt
```diff
@@ -128,32 +133,75 @@ fun PostCardSimple(
...
+                    Text(
+                        text = stringResource(id = R.string.agree),
+                        style = MaterialTheme.typography.button,
+                        color = MaterialTheme.colors.primary,
+                        modifier = Modifier
+                            .padding(15.dp)
+                            .clickable {
+
+                                openDialog = false
+                            }
```
JolandaVerhoef, line 202: nit: this can be one line

JetNews/app/src/main/res/values/strings.xml
```diff
@@ -25,4 +25,8 @@
     <string name="cd_bookmark">Bookmark</string>
     <string name="unbookmark">unbookmark</string>
     <string name="bookmark">bookmark</string>
+    <string name="fewer_stories">Show fewer stories like this?</string>
+    <string name="fewer_stories_content">This feature is not yet implemented</string>
+    <string name="agree">AGREE</string>
+    <string name="disagree">DISAGREE</string>
```
JolandaVerhoef, line 31: nit: this string is unused

JolandaVerhoef (changes requested):

JetNews/app/src/main/java/com/example/jetnews/ui/home/PostCards.kt
angeles-bilbao6: You are right!

JolandaVerhoef:
You should use an IconButton for the "more menu" icon. This makes sure the tap target is sufficiently large (48x48) and a ripple is shown when the user presses the icon. 

However, if you use an IconButton, the icon should be placed all the way in the top right of the parent container. To do that, you can move the padding down into the children. So the image would get a top, left, and bottom padding of 16dp, and the column with the title in it would get a top + bottom padding of 16dp. That way, you can leave out the padding of the icon button, and place it all the way at the top right. 
The parent container can be clickable as a whole, as the icon button will override this anyway.

When testing, please make sure that the ripple of the parent container fills the whole row (it currently doesn't) and that the more icon also has a fancy ripple.

JolandaVerhoef (approved):
LGTM!

PR #540, opened 2021-05-19
[Jetsnack] Move to Accompanist SystemUiController

chrisbanes:
Removed the version in Jetsnack in favour of the published
library in Accompanist.


nickbutcher (approved):
Thanks!

PR #548, opened 2021-05-26
[Jetsnack] Improve RTL layouts

JolandaVerhoef:


Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Grid.kt
```diff
@@ -57,7 +57,7 @@ fun VerticalGrid(
             val columnY = Array(columns) { 0 }
             placeables.forEachIndexed { index, placeable ->
                 val column = index % columns
-                placeable.place(
+                placeable.placeRelative(
```
nickbutcher, line 60: We should probably remove this and use `LazyGrid` but not needed for this PR.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Snacks.kt
```diff
@@ -100,7 +102,10 @@ fun SnackCollection(
                 modifier = Modifier.align(Alignment.CenterVertically)
             ) {
                 Icon(
-                    imageVector = Icons.Outlined.ArrowForward,
+                    imageVector = mirroringIcon(
```
nickbutcher, line 105: Could this use `mirroringBackIcon`?

nickbutcher (approved):

Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Snacks.kt
JolandaVerhoef: Nope, in this layout the arrow points the other way.

See the forward pointing arrows in the screenshot:

Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Grid.kt
JolandaVerhoef: Good point!

PR #551, opened 2021-05-29
Fix: constrast, constast, etc --> contrast

eneim:
I guess this fix is trivial that we do not need an issue to track. But if I should follow the contribution rules, please advise and I will do it.


nickbutcher (approved):

nickbutcher:
Thanks!

PR #560, opened 2021-06-04
[Crane] Fix Map zoom controls not working bug

manuelvicnt:


Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
```diff
@@ -192,15 +195,14 @@ private fun MapViewContainer(
     latitude: String,
     longitude: String
 ) {
+    val cameraPosition = remember(map) { LatLng(latitude.toDouble(), longitude.toDouble()) }
```
JoseAlcerreca, line 198: This is a good opportunity to showcase `rememberSaveable`?

manuelvicnt: That'd be a micro-optimization IMO, as this is not a state to remember, just a constant I don't want to compute multiple times.

JoseAlcerreca (approved):

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
chrisbanes, line 198: This won't update if `latitude` or `longitude` change 🤔

manuelvicnt: And `map` doesn't make any sense there! Thanks. Let's try to write correct code! Updating...

chrisbanes (approved):

PR #563, opened 2021-06-10
Remove jcenter

yogurtearl:
Jcenter is now read only and no longer needed. 

Added maven central where needed.
Use Volley 1.2.0 which is available on maven.google.com (required as a dep of the Maps SDK)


nickbutcher:
Volley?!? 😳

nickbutcher:
Thanks!

PR #567, opened 2021-06-12
[Owl] Use repeatable preview annotation for dark mode

takahirom:
Thank you for the awesome samples!
I applied the same approach as below.
https://github.com/android/compose-samples/blob/2fbf0c2eb7f8487e469a9be0376b3d6c1c7bfff5/JetNews/app/src/main/java/com/example/jetnews/ui/home/PostCards.kt#L251


takahirom:
Sorry. I should use dev_xx branch for this change. I will close this.

PR #586, opened 2021-06-25
[Jetsnack] Avoid mutating colors parameter

### LINKED ISSUE #584: [JetSnack] ProvideJetsnackColors mutates it's parameter
See: https://android-review.googlesource.com/c/platform/frameworks/support/+/1636460/

Line: https://github.com/android/compose-samples/blob/fd7a5bb94808200df9fb8ccb8e17cb352c0aa5ae/Jetsnack/app/src/main/java/com/example/jetsnack/ui/theme/Theme.kt#L244

This should copy the parameter to keep the same object caching behavior as `MaterialTheme` without modifying the passed parameter.

Raised in kotlinlang: https://kotlinlang.slack.com/archives/CJLTWPH7S/p1624580246331200

alexvanyo:
Adds a `copy` method to `JetsnackColors`, which is used to copy the passed in `colors` when providing `JetsnackColors`.

This avoids `update()` mutating the `colors` passed in as a parameter, which would manifest itself as a bug where switching the system between dark theme and light theme would only work the first time, as `DarkColorPalette` and `LightColorPalette` would overwrite each other. (right now that bug doesn't occur, since `MainActivity` is being recreated upon configuration changes).

This fixes #584, which also has more context.

### DIFF
--- a/Jetsnack/app/src/main/java/com/example/jetsnack/ui/theme/Theme.kt
+++ b/Jetsnack/app/src/main/java/com/example/jetsnack/ui/theme/Theme.kt
@@ -234,6 +234,37 @@ class JetsnackColors(
         notificationBadge = other.notificationBadge
         isDark = other.isDark
     }
+
+    fun copy(): JetsnackColors = JetsnackColors(
+        gradient6_1 = gradient6_1,
+        gradient6_2 = gradient6_2,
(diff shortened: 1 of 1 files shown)


nickbutcher (approved):

nickbutcher:
Thanks for addressing!

PR #590, opened 2021-07-01
[Jetsnack] Add an example ViewModel to SnackDetail

### LINKED ISSUE #565: Jetsnack - SnackDetail viewmodel code is not properly done
I know the samples are meant as samples, but it would still be best not to showcase work with hacks.

It is a good place to showcase how to create viewmodel associated with navBackStackEntry. Instead of letting view calling repo methods, best to create a viewmodel scoped to the SnackDetail's navBackStackEntry, and call repo methods in SnackDetailViewModel

```
@Composable
fun SnackDetail(
    snackId: Long,
    upPress: () -> Unit
) {
    val snack = remember(snackId) { SnackRepo.getSnack(snackId) }
    val related = remember(snackId) { SnackRepo.getRelated(snackId) }
```

bentrengrove:
Fixes #565 

Implemented a contrived simple ViewModel for the SnackDetail screen. 

It is obviously up to you if you actually want this change merged in, I just did it as part of my learning and thought it might be useful for others. I would also be happy to convert it to using LiveData or StateFlow if that would be a better example but thought I would keep it simple in the first pass.

The ViewModelFactory code was taken straight from Jetcaster in order to be consistent.

### DIFF
--- a/Jetsnack/app/src/main/java/com/example/jetsnack/ui/snackdetail/SnackDetail.kt
+++ b/Jetsnack/app/src/main/java/com/example/jetsnack/ui/snackdetail/SnackDetail.kt
@@ -60,10 +60,10 @@ import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.lerp
 import androidx.compose.ui.unit.sp
 import androidx.compose.ui.util.lerp
+import androidx.lifecycle.viewmodel.compose.viewModel
 import com.example.jetsnack.R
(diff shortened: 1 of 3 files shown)


JoseAlcerreca:
Hey Ben, thanks for your contribution. As this adds nothing new (the tricky stuff comes from Jetcaster), and the implementation of the VM is as simple as it gets, I would rather keep Jetsnack simple.

Keeping the PR open for other thoughts.

PR #593, opened 2021-07-03
Update Dependencies.kt

wiryadev:
Bump android gradle plugin to beta05 for Jetsnack


wiryadev:
@googlebot I signed it!

JoseAlcerreca:
Hi, thanks for your contributions but we only bump versions when a new version of Compose is released. We test everything against a snapshot beforehand and then merge when it's out.

PR #608, opened 2021-07-20
[Jetsnack] Events experiments

manuelvicnt:


Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -17,30 +17,155 @@
...
+    init {
+        // Restore snackbar messages state
+        initialSnackbarMessages.forEach { showSnackbar(it) }
+
+        // Process snackbar events only when the lifecycle is at least STARTED
+        coroutineScope.launch {
+            snackbarMessages
+                .receiveAsFlow()
+                .onEach { pendingSnackbarMessages.remove(it) }
+                .flowWithLifecycle(lifecycle)
```
alexvanyo, line 125: I don't think this behavior is quite what is desired.

`flowWithLifecycle` uses `callbackFlow` under the hood, which by default has a buffer size of `Channel.BUFFERED` (which is 64 by default).

So even though the below `collect` is slow (since `snackbarHostState.showSnackbar` suspends until the message is shown and then hidden), up to 64 messages will be buffered and effectively enqueued to be displayed.

Adding a `.buffer(Channel.RENDEZVOUS)` after `flowWithLifecycle` will remove that 64 element buffer, so the desired 3 element buffer of the `snackbarMessages` `Channel` will be used instead.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -17,30 +17,155 @@
...
+    private val snackbarHostState: SnackbarHostState,
+    coroutineScope: CoroutineScope,
+    lifecycle: Lifecycle,
+    initialSnackbarMessages: List<String>
+) {
+    // Tabs for JetsnackBottomBar
+    val tabs = HomeSections.values()
+
+    // Queue a maximum of 3 snackbar messages
+    private val snackbarMessages = Channel<String>(3, BufferOverflow.DROP_OLDEST)
```
alexvanyo, line 112: The `BufferOverflow.DROP_OLDEST` here doesn't interact with the `pendingSnackbarMessages`.

If a message if enqueued, but is later dropped due to the `Channel` overflowing, that dropped message isn't removed from the `pendingSnackbarMessages` since it will never be received.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
alexvanyo: Hmm, upon further investigation, even `.buffer(Channel.RENDEZVOUS)` will allow an unwanted element to get through with `flowWithLifecycle`:

Suppose `showSnackbar` is called with successive numbers `1`, `2`, `3`, `4`, `5`, ..., `9`, `10` extremely quickly (while the snackbar for `1` is showing)

`2` will be displayed next, followed by `8`, `9`, and `10`, when we really just want `8`, `9` and `10`. I think that is because `2` is internally received by `flowWithLifecycle` even though `1` is still being displayed.

Notably, using `repeatOnLifecycle` directly can avoid that, resulting in the desired `1`, `8`, `9`, `10`:

```kotlin
        coroutineScope.launch {
            lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
                snackbarMessages
                    .receiveAsFlow()
                    .collect { message ->
                        snackbarHostState.showSnackbar(message)
                    }
            }
        }
```

The difference is subtle, but it ensures that a message is received from the channel if and only the `snackbarHostState.showSnackbar` is called with it.

manuelvicnt: Good catch! The underlying implementation of `callbackFlow` bits again :) thank you!

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
manuelvicnt: I don't quite get what you mean. The message is only added to `pendingSnackbarMessages` _iff_ it could be added to the `Channel`. See L135.

alexvanyo: Right, the issue I'm seeing is if the `DROP_OLDEST` comes into play if there are already `3` messages in the `Channel`'s buffer.

`trySend` will always succeed (if the channel isn't closed), since it will kick out the oldest message from the `Channel`, but that removed message isn't ever removed from `pendingSnackbarMessages`.

So upon state saving/restoration, there could be really old messages that were dropped from the `Channel`, but weren't removed from `pendingSnackbarMessages`.

manuelvicnt: Changed the code for this. Please take another look and let me know what you think! Thanks!

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -17,30 +17,193 @@
...
+    private val onBufferOverflow: BufferOverflow = BufferOverflow.SUSPEND
+) {
+    // Cache of the Channel buffer used to restore state
+    val pendingElements = mutableStateListOf<T>()
+    private val pendingElementsMutex = Mutex()
+
+    private val _channel = Channel<T>(capacity, onBufferOverflow)
+    val stream = _channel
+        .receiveAsFlow()
+        .onEach { pendingElementsMutex.withLock { pendingElements.remove(it) } }
```
alexvanyo, line 184: How do you feel about replacing this with `removeFirst()`, instead of `remove(it)`?
If `removeFirst()` or `remove(it)` ever do something different, then I think `remove(it)` could have unintended consequences.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -17,30 +17,193 @@
...
+ * Implementation of a Channel whose buffered elements can be accessed synchronously so that they
+ * can be saved/restored on activity and process recreation.
+ *
+ * In this implementation, the exposed `stream` flow is supposed to be collected once at a time
+ * since in case there are multiple collectors, only one will get the event.
+ * If events need to be broadcasted to multiple collectors, expose stream using the `.shareIn`
+ * operator instead.
+ */
+class StateChannel<T>(
+    private val capacity: Int,
```
alexvanyo, line 174: A small API nit, it might be good to document that this capacity should be a non-negative integer, and not one of the `Channel.***` constants.

In particular trying to pass in `Channel.BUFFERED` would lead to some weird effects.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -17,30 +17,193 @@
...
+    private val onBufferOverflow: BufferOverflow = BufferOverflow.SUSPEND
+) {
+    // Cache of the Channel buffer used to restore state
+    val pendingElements = mutableStateListOf<T>()
```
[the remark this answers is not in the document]
alexvanyo: One more small API nit, the mutable-ness of this list probably shouldn't be exposed since it is guarded by the mutex.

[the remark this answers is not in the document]
manuelvicnt: `mutableStateListOf` is a list whose changes are tracked by Compose as opposed to `mutableListOf`. Having `mutableStateOf(mutableListOf())` has some problems with the Saver as it's a mix between Compose and non-Compose trackable state.

Using the latter was causing the Saver to not react to new lists assigned to `mutableStateOf`. `mutableStateListOf` works as expected

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
manuelvicnt: +1, definitely!

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
manuelvicnt: Sure! Done thanks

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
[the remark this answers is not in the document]
alexvanyo: Got it, is that Compose-tracked state used anywhere? Or in other words, would there be a difference with this:

```suggestion
    val pendingElements = mutableListOf<T>()
```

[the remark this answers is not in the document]
manuelvicnt: Yeah, it's used in the `ScaffoldStateHolderSaver` to save/restore the state after activity and process recreation. `mutableListOf` is not tracked by Compose, therefore, the saver doesn't get updated with the new values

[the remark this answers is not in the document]
alexvanyo: Huh, I'm not seeing any difference in behavior between those in my testing, it looks like the `save` block is called on the lifecycle change, and not upon the tracked state change. Do you know of something I'm missing here?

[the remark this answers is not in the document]
manuelvicnt: mmm this might be due to the fact that I'm not assigning a new list anymore. Let me revert to the original impl. Thanks for pointing out

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -17,30 +17,192 @@
...
+    private val pendingElementsMutex = Mutex()
+
+    private val _channel = Channel<T>(capacity, onBufferOverflow)
+    val stream = _channel
+        .receiveAsFlow()
+        .onEach { pendingElementsMutex.withLock { pendingElements.removeFirstOrNull() } }
+
+    suspend fun send(element: T) {
+        _channel.send(element)
+        pendingElementsMutex.withLock {
```
alexvanyo, line 191: I don't know how much of a problem this would actually be in practice, but because there are two suspension points here (sending to the `_channel `, and acquiring the lock), conceivably there's a chance that the job is cancelled after the element is sent to the `_channel`, but before the lock is acquired (and therefore before the `pendingElements` list is synchronized with it).

If that happens, the `pendingElements` could get out of sync with the actual `Channel`. No idea what the likelihood of that would be, but wanted to bring it up.

alexvanyo (commented):
I'm liking how this is shaping up! I had one more fairly hypothetical comment, haven't been able to see that happen in testing.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
JoseAlcerreca: "if a coroutine is working in a computation and does not check for cancellation, then it cannot be cancelled"

https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative

alexvanyo: Right, that will be true once we enter the `withLock` block.

But before that, if we suspend waiting for the lock and then get cancelled, we've already sent to the channel but won't update the list. Since we're also introducing a suspension point, I also don't think we're guaranteeing that the elements will be added to the list in the same order that they were sent to the channel.

I created https://pl.kotl.in/lNRBrz_4H as a playground for something similar. It's non-deterministic, but if you run it multiple times you should see the `MutableStateFlow` get out of sync with the list.

alexvanyo: I explored a slightly different way to do this [here](https://github.com/android/compose-samples/compare/mv/events_experiments...alexvanyo:av/events_experiments?expand=1), which I believe avoids the problem of the state getting out of sync with the exposed `Flow`.

PR #609, opened 2021-07-21
Use ImageLoaderFactory to create the singleton ImageLoader.

colinrtwhite:
Removes `ProvideImageLoader` and implements `ImageLoaderFactory` in the Crane and Owl apps. This has a couple benefits:

- We only create one `ImageLoader` which is shared for every `rememberImagePainter` call. Each `ImageLoader` has its own memory cache, OkHttpClient, network observer, so it's cheaper to only create one.
- `ImageLoaderFactory` creates the `ImageLoader` lazily the first time `rememberImagePainter` is called.
- `LocalImageLoader provides ImageLoader(context)` sets `LocalImageLoader.current` for any child composables, however it doesn't set the singleton `ImageLoader` in the `Coil` object. If you call `LocalContext.current.imageLoader` elsewhere in the app, it'll return the singleton `ImageLoader` and not the value set in `LocalImageLoader`. Maybe there's a better way for Coil to handle this (open to suggestions!)...

### DIFF
--- a/Crane/app/src/main/java/androidx/compose/samples/crane/CraneApplication.kt
+++ b/Crane/app/src/main/java/androidx/compose/samples/crane/CraneApplication.kt
@@ -20,11 +20,16 @@ import android.app.Application
 import androidx.compose.samples.crane.util.UnsplashSizingInterceptor
 import coil.ImageLoader
 import coil.ImageLoaderFactory
+import coil.compose.rememberImagePainter
 import dagger.hilt.android.HiltAndroidApp
 
 @HiltAndroidApp
 class CraneApplication : Application(), ImageLoaderFactory {
 
+    /**
+     * Create the singleton [ImageLoader].
+     * This is used by [rememberImagePainter] to load images in the app.
+     */
     override fun newImageLoader(): ImageLoader {
         return ImageLoader.Builder(this)
             .componentRegistry {
--- a/Crane/app/src/main/java/androidx/compose/samples/crane/CraneApplication.kt
+++ b/Crane/app/src/main/java/androidx/compose/samples/crane/CraneApplication.kt
@@ -17,7 +17,19 @@
 package androidx.compose.samples.crane
 
 import android.app.Application
+import androidx.compose.samples.crane.util.UnsplashSizingInterceptor
+import coil.ImageLoader
+import coil.ImageLoaderFactory
 import dagger.hilt.android.HiltAndroidApp
 
 @HiltAndroidApp
-class CraneApplication : Application()
+class CraneApplication : Application(), ImageLoaderFactory {
+
+    override fun newImageLoader(): ImageLoader {
+        return ImageLoader.Builder(this)
+            .componentRegistry {
+                add(UnsplashSizingInterceptor)
+            }
+            .build()
+    }
+}
--- a/Crane/app/src/main/java/androidx/compose/samples/crane/util/UnsplashSizingInterceptor.kt
+++ b/Crane/app/src/main/java/androidx/compose/samples/crane/util/UnsplashSizingInterceptor.kt
@@ -16,13 +16,7 @@
 
 package androidx.compose.samples.crane.util
 
-import androidx.compose.runtime.Composable
-import androidx.compose.runtime.CompositionLocalProvider
-import androidx.compose.runtime.remember
-import androidx.compose.ui.platform.LocalContext
-import coil.ImageLoader
 import coil.annotation.ExperimentalCoilApi
-import coil.compose.LocalImageLoader
 import coil.intercept.Interceptor
 import coil.request.ImageResult
 import coil.size.PixelSize
--- a/Crane/app/src/main/java/androidx/compose/samples/crane/util/UnsplashSizingInterceptor.kt
+++ b/Crane/app/src/main/java/androidx/compose/samples/crane/util/UnsplashSizingInterceptor.kt
@@ -28,18 +28,6 @@ import coil.request.ImageResult
 import coil.size.PixelSize
 import okhttp3.HttpUrl
 
-@Composable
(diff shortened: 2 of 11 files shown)


Crane/app/src/main/java/androidx/compose/samples/crane/util/UnsplashSizingInterceptor.kt
```diff
@@ -50,11 +38,9 @@ object UnsplashSizingInterceptor : Interceptor {
         val size = chain.size
         if (data is String &&
             data.startsWith("https://images.unsplash.com/photo-") &&
-            size is PixelSize &&
-            size.width > 0 &&
```
colinrtwhite, line 54: `PixelSize` is guaranteed to have a positive width/height.

Crane/buildSrc/src/main/java/com/example/crane/buildsrc/Dependencies.kt
```diff
@@ -21,7 +21,7 @@ object Versions {
 }
 
 object Libs {
-    const val androidGradlePlugin = "com.android.tools.build:gradle:7.0.0-beta05"
+    const val androidGradlePlugin = "com.android.tools.build:gradle:7.0.0-rc01"
```
nickbutcher, line 24: Please can we revert this AGP upgrade unless there's a specific need? We prefer to keep all the samples in sync and upgrade them together.

Owl/app/src/main/java/com/example/owl/OwlApplication.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import android.app.Application
+import coil.ImageLoader
+import coil.ImageLoaderFactory
+import com.example.owl.ui.utils.UnsplashSizingInterceptor
+
+@Suppress("unused")
+class OwlApplication : Application(), ImageLoaderFactory {
+
+    override fun newImageLoader(): ImageLoader {
```
nickbutcher, line 11: Please can we add a comment explaining this.

Crane/app/src/main/java/androidx/compose/samples/crane/CraneApplication.kt
```diff
@@ -17,7 +17,19 @@
...
+import androidx.compose.samples.crane.util.UnsplashSizingInterceptor
+import coil.ImageLoader
+import coil.ImageLoaderFactory
 import dagger.hilt.android.HiltAndroidApp
 
 @HiltAndroidApp
-class CraneApplication : Application()
+class CraneApplication : Application(), ImageLoaderFactory {
+
+    override fun newImageLoader(): ImageLoader {
```
nickbutcher, line 28: Please can we add a comment explaining this.

Owl/buildSrc/src/main/java/com/example/owl/buildsrc/Dependencies.kt
```diff
@@ -21,7 +21,7 @@ object Versions {
 }
 
 object Libs {
-    const val androidGradlePlugin = "com.android.tools.build:gradle:7.0.0-beta05"
+    const val androidGradlePlugin = "com.android.tools.build:gradle:7.0.0-rc01"
```
nickbutcher, line 24: ditto.

Crane/app/src/main/java/androidx/compose/samples/crane/util/UnsplashSizingInterceptor.kt
nickbutcher: ack—thanks!

nickbutcher (approved):
Thanks for the improvment! Please can we revert the AGP change, other than that LGTM.

colinrtwhite:
@nickbutcher Thanks for the quick review! Updated based on the feedback.

nickbutcher (approved):
Thanks for the changes!

nickbutcher:
Ahh, looks like this needs a `gradlew spotlessApply` please.

colinrtwhite:
@nickbutcher Updated

nickbutcher:
Thanks for the contribution!

ColtonIdle:
@colinrtwhite I have a question about the implementation here. The way that I currently have coil working was that I was using it in a specific compose-only module.

The goal of this module is that everything is self contained, and so I don't know where I would put 

```
override fun newImageLoader(): ImageLoader {
        return ImageLoader.Builder(this).componentRegistry { add(MyInterceptor) }.build()
    }
```

as I don't have an Application class. Any tips?

colinrtwhite:
@ColtonIdle You can use `Coil.setImageLoader`; check out the docs [here](https://coil-kt.github.io/coil/getting_started/#singleton).

ColtonIdle:
@colinrtwhite gotta be honest. Not sure where I would end up calling that setImageLoader method since my module is just filled with composables.

I have a ImageWithUrl that just wraps a Coil image. i.e.

```
@Composable
fun ImageWithUrl(
    imageUrl: String,
) {
```

I will try to play around with this a bit more as I'm probably going a _bit_ off topic at this point.

ColtonIdle:
For anyone following along... I ended up going this route with the guidance of @colinrtwhite 

```
@Composable
fun ImageWithUrl(
    imageUrl: String,
) {
    Box(modifier) {
        Image(
            painter =
                rememberImagePainter(
                    data = imageUrl,
                    builder = { crossfade(true) },
                    imageLoader = getImageLoader(LocalContext.current),
                ),
            contentDescription = null)
    }
}
private var imageLoader: ImageLoader? = null
@Synchronized
private fun getImageLoader(context: Context): ImageLoader {
    imageLoader?.let {
        return it
    }
    return ImageLoader.Builder(context)
        .componentRegistry { add(MyInterceptor) }
        .build()
        .also { imageLoader = it }
}
```

PR #618, opened 2021-08-01
Update to Java 11 for artic fox and gradle 7.0.0 better compatibility

Canato:
## Description:
Update Java version to `11` for better compatibility with **Artic Fox** and `gradle 7.0.0`


Jetcaster/build.gradle
```diff
@@ -72,7 +72,7 @@ subprojects {
             freeCompilerArgs += '-Xopt-in=kotlin.Experimental'
 
             // Set JVM target to 1.8
```
crjacinro, line 74: can we update this comment also?

Canato: Thanks, missed

PR #629, opened 2021-08-25
[JetNews] List-detail support for main page

alexvanyo:
Adds an initial list-detail support for the main page.

The overall approach relies on distinguishing between what the user sees as different screens, and what we implement as different navigation routes.

Navigation-wise, the "home" route is responsible for drawing the list detail screen, _as well as_ the list screen and detail screen on their own (for small devices). Distinguishing between which one to show is now simply a matter of keeping track of whether the user is looking at the list, or a selected article.

Doing so allows meeting a couple of nice behaviors:
- We keep track of the user's state (scroll state on the list/detail, as well as which one they were looking at), even while changing which side we are looking at
- We avoid navigating as a side-effect of a size change
- We can make size changes reversible (mostly as a consequence of the previous point)

|List:|Detail:|List and Detail:|
|-|-|-|
|device-2021-09-02-164232|device-2021-09-02-164126|device-2021-09-02-164044|


JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeViewModel.kt
```diff
@@ -37,6 +37,8 @@ import java.util.UUID
  */
 data class HomeUiState(
     val posts: List<Post> = emptyList(),
+    val selectedPostId: String? = null, // TODO back selectedPostId in a SavedStateHandle
```
manuelvicnt, line 40: I'd expect this to expose the actual `Post` and not the id, it's better if the UI doesn't need to figure this out

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -121,103 +139,160 @@ fun HomeScreen(
...
+        }
+    }
+
+    BoxWithConstraints {
+        val useListDetail = when (getWindowSize(maxWidth)) {
+            WindowSize.Compact, WindowSize.Medium -> false
+            WindowSize.Expanded -> true
+        }
+
+        if (useListDetail || uiState.isListOpen) {
```
manuelvicnt, line 167: What's this exactly checking? The `else` branch also looks weird just having the `ArticleScreen` when this is supposed to be `HomeScreen`. Is there a better way to handle this? 

Have some problems with this if statement: 1) it's not clear when this should be true or false, and 2) need to figure out if there's a way to remove the `ArticleScreen` call from the else branch

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -121,103 +139,160 @@ fun HomeScreen(
...
+                    )
+                    if (snackbarResult == SnackbarResult.ActionPerformed) {
+                        onRefreshPostsState()
+                    }
+                    // Once the message is displayed and dismissed, notify the ViewModel
+                    onErrorDismissState(errorMessage.id)
+                }
+            }
+        } else {
+            val selectedPost = uiState.posts.find { it.id == uiState.selectedPostId }
```
manuelvicnt, line 269: Better if this logic goes to the VM and we expose the actual post and not just the ID

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -121,103 +139,160 @@ fun HomeScreen(
...
+
+            if (selectedPost != null) {
+                val lazyListState = detailLazyListStates.getValue(selectedPost.id)
+
+                ArticleScreen(
```
manuelvicnt, line 274: Weird to have the `ArticleScreen` call on its own here. Is this full screen and for Compat Windows? Cannot follow the logic in here. More comments are needed.

But if this is for mobile phones, let's see if it's possible to make a navigation call

manuelvicnt: Oh! It might be the case we don't call `navigate(ARTICLE_SCREEN)` anymore, is that the case? Then we wouldn't be able to use the `ArticleViewModel`. This looks ... complicated

manuelvicnt: Oh! It might be the case we don't call `navigate(ARTICLE_SCREEN)` anymore, is that the case? Then we wouldn't be able to use the `ArticleViewModel`. This looks ... complicated

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -121,103 +139,160 @@ fun HomeScreen(
...
+
+                // If we are just showing the detail, have a back press switch to the list.
+                if (!useListDetail) {
+                    BackHandler {
+                        onInteractWithList()
+                    }
+                }
+            } else {
+                // TODO: Improve UX
+                LaunchedEffect(Unit) {
```
manuelvicnt, line 292: Confusing, not sure what's happening here

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -121,103 +139,160 @@ fun HomeScreen(
...
+                    onToggleFavorite = {
+                        onToggleFavorite(selectedPost.id)
+                    },
+                    lazyListState = lazyListState,
+                )
+
+                // If we are just showing the detail, have a back press switch to the list.
+                if (!useListDetail) {
+                    BackHandler {
+                        onInteractWithList()
```
manuelvicnt, line 287: Can we make Navigation do this for us?

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -257,23 +332,98 @@ private fun LoadingContent(
...
+                        .then(
+                            if (useListDetail) {
+                                Modifier.width(334.dp)
+                            } else {
+                                Modifier.fillMaxSize()
+                            }
+                        )
+                        .pointerInput(Unit) {
+                            while (currentCoroutineContext().isActive) {
+                                awaitPointerEventScope {
```
manuelvicnt, line 387: How often does this happen? Can we optimize this in any way?

manuelvicnt (commented):
Left some initial comments, some of them are due to the lack of familiarity with the navigation flow you have in mind. We should try to work on simplifying it as the implementation detail looks tedious. Ideally, someone without the flow chart should be able to follow along easily.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
alexvanyo: Right, I'm avoiding doing any sort of `navigate` as a side-effect of a resize, which implies that we need to show the `ArticleScreen` (in mostly its entirety, with the bottom bar, etc.) at this navigation route.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
alexvanyo: This is the same issue that the `ArticleScreen` currently has on its own: If after process death we get back to this screen and can't load the article, it'd be better to have an affordance to retry or go back rather than just jumping back to the list.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
alexvanyo: Technically yes, _but_ that would require performing a navigate upon switching from the list/detail to the detail screen, which we want to avoid for a couple of reasons:
- Performing navigation in a side-effect results in some jank, since it happens after composition
- We'd also need to do another navigate upon switching back from detail to the list/detail screen to preserve reversability

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
alexvanyo: I'm not sure... it should be pretty lightweight I'd hope?
If we really needed to optimize it, we could only install this modifier on the side that isn't currently active.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
alexvanyo: This changed a bit with the latest version, let me know what you think!

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
alexvanyo: I changed this with the latest commit, which more cleanly decides between the 3 possible "screens" from the `HomeRoute` `@Composable`. This also makes it clearer that it's fine to call into the `ArticleScreen`, since the `HomeRoute` is a higher level component.

alexvanyo:
@manuelvicnt I just pushed up another commit, which makes quite a few changes.

Most notably, I've removed the article route entirely (for now), so articles are now only displayed from the `home` route.

I also created a `HomeRoute` `@Composable`, which is the primary component responsible for determining which `HomeScreen` to display (based on the size of the screen, and the available state).

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
```diff
@@ -0,0 +1,182 @@
...
+        is HomeUiState.HasPosts -> uiState.postsFeed.allPosts
+        is HomeUiState.NoPosts -> emptyList()
+    }.associate { post ->
+        key(post.id) {
+            post.id to rememberLazyListState()
+        }
+    }
+
+    BoxWithConstraints {
+        val useListDetail = when (getWindowSize(maxWidth)) {
```
manuelvicnt, line 111: Not everyone is familiar with the list/detail terminology. Could we use something more explicit?

```suggestion
        val showArticleDetails = when (getWindowSize(maxWidth)) {
```

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
```diff
@@ -0,0 +1,182 @@
...
+            WindowSize.Compact, WindowSize.Medium -> false
+            WindowSize.Expanded -> true
+        }
+
+        // Determine which type of the home screen to display
+        val homeScreenType = if (useListDetail) {
+            HomeScreenType.ListArticle
+        } else {
+            when (uiState) {
+                is HomeUiState.HasPosts -> if (uiState.isArticleOpen) {
```
manuelvicnt, line 121: Formatting here is weird and makes reading the logic a bit harder. Can we move the if statement to a new line?

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
```diff
@@ -0,0 +1,182 @@
...
+                        onToggleFavorite(uiState.selectedPost.id)
+                    },
+                    lazyListState = articleDetailLazyListStates.getValue(
+                        uiState.selectedPost.id
+                    ),
+                )
+
+                // If we are just showing the detail, have a back press switch to the list.
+                BackHandler {
+                    onInteractWithList()
```
manuelvicnt, line 182: Is this enough to provoke back navigation to the list? If so, can you explain a bit more about how this happens in the comment above? or above this line? thank you

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
```diff
@@ -0,0 +1,182 @@
...
+                }
+            }
+        }
+    }
+}
+
+private enum class HomeScreenType {
+    ListArticle,
+    List,
```
manuelvicnt, line 191: `List` is confusing... can we have `Feed` instead? As mentioned in the other comment, most people don't know about the list/detail terminology, so we could be more explicit in these names.

e.g. `FeedWithArticleDetails`, `Feed`, `ArticleDetails`

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
```diff
@@ -60,76 +66,176 @@ import com.example.jetnews.R
...
+                            .fillMaxSize()
+                            .pointerInput(Unit) {
+                                while (currentCoroutineContext().isActive) {
+                                    awaitPointerEventScope {
+                                        awaitPointerEvent(PointerEventPass.Initial)
+                                        onInteractWithDetail(detailPost.id)
+                                    }
+                                }
+                            },
+                        contentPadding = rememberInsetsPaddingValues(
```
manuelvicnt, line 159: This content padding is the same as above, should we reuse it? we can extract it out to a `contentPadding` variable

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
```diff
@@ -60,76 +66,176 @@ import com.example.jetnews.R
...
+                    articleDetailLazyListStates.getValue(detailPost.id)
+                }
+
+                // Key against the post id to avoid sharing any state between different posts
+                key(detailPost.id) {
+                    PostContent(
+                        post = detailPost,
+                        modifier = modifier
+                            .fillMaxSize()
+                            .pointerInput(Unit) {
```
manuelvicnt, line 151: The logic for this `pointerInput` and the above is the same except the action to take. We could create a modifier out of this with a more meaningful name and pass the lambda in

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeViewModel.kt
```diff
@@ -23,29 +23,74 @@ import com.example.jetnews.R
...
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import java.util.UUID
 
 /**
  * UI state for the Home screen
  */
-data class HomeUiState(
-    val posts: List<Post> = emptyList(),
+sealed interface HomeUiState {
```
manuelvicnt, line 40: I assume this is a pattern for preventing inconsistent UI states. If that's the case, document this thoroughly as this might be confusing for a lot of people if they aren't aware of the context behind it.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
manuelvicnt: We need to be conscious about this as every user input triggers a `uiState` update in the VM

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeViewModel.kt
```diff
@@ -55,17 +100,24 @@ class HomeViewModel(
     private val postsRepository: PostsRepository
 ) : ViewModel() {
 
+    private val viewModelState = MutableStateFlow(HomeViewModelState(isLoading = true))
+
     // UI state exposed to the UI
-    private val _uiState = MutableStateFlow(HomeUiState(loading = true))
-    val uiState: StateFlow<HomeUiState> = _uiState.asStateFlow()
+    val uiState = viewModelState
+        .map { it.toUiState() }
```
manuelvicnt, line 107: What about using `distinctUntilChanged()`? Is this needed, or is the `MutableStateFlow` always going to cover that case?

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
alexvanyo: `showArticleDetails` is a little confusing, since `showArticleDetails` could be `false` and we still show the article, I'll rename this to `showFeedAndArticleDetails` I think.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
alexvanyo: These are subtly different, due to the `applyStart` vs `applyEnd`. Although I need to update this logic anyway now that the nav rail is involved as well, so thanks for the call out!

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeViewModel.kt
alexvanyo: Yup, `MutableStateFlow` has a `distinctUntilChanged` built in. So no need to add one separately in this mapping.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
alexvanyo: Even better, I can inline this since it's only used directly below to create the `homeScreenType`.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
alexvanyo: Updated with the latest commits.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
alexvanyo: I created a modifier here, and removed one place where it was unnecessary.
I feel like we shouldn't need to be too concerned about this firing extremely often, the only expense should be contesting the CAS loop?

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
alexvanyo: Updated, and added more documentation here!

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
alexvanyo: Added!

alexvanyo:
@manuelvicnt This is ready for another look! I've fully rebased on to the navrail PR, and have fixed up a couple things related to that rebase.

I also got to all of the comments you left, and added quite a bit of extra inline documentation around the approach.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
```diff
@@ -108,66 +116,64 @@ fun HomeRoute(
...
         // Determine which type of the home screen to display
-        val homeScreenType = if (useListDetail) {
-            HomeScreenType.ListArticle
-        } else {
-            when (uiState) {
-                is HomeUiState.HasPosts -> if (uiState.isArticleOpen) {
-                    HomeScreenType.Article
-                } else {
-                    HomeScreenType.List
+        val homeScreenType = when (getWindowSize(maxWidth)) {
```
manuelvicnt, line 120: We should only calculate the `windowSize` whenever the `maxWidth` changes. What about including this in a `remember`?

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
```diff
@@ -170,17 +160,29 @@ fun HomeListDetailScreen(
 }
 
 /**
- * The home screen displaying a list.
+ * A [Modifier] that tracks all input, and calls [block] every time input is received.
+ */
+private fun Modifier.notifyInput(block: () -> Unit): Modifier =
+    pointerInput(block) {
```
manuelvicnt, line 166: You might not want to reset this `pointerInput` if the lambda changes. What about passing `Unit` as `key` and the protecting the block for updates?

```suggestion
    val blockState by rememberUpdatedState(block)
    pointerInput(Unit) {
```

manuelvicnt (approved):
Two nits but LGTM overall. Thank you!

manuelvicnt:
Also @alexvanyo, can you add screenshots to the description of the PR for reviewers to get a sense of what's been implemented? At some point, we'll also have those mocks in the README. Thanks!

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
```diff
@@ -0,0 +1,209 @@
...
+        is HomeUiState.NoPosts -> emptyList()
+    }.associate { post ->
+        key(post.id) {
+            post.id to rememberLazyListState()
+        }
+    }
+
+    BoxWithConstraints {
+        // Determine which type of the home screen to display
+        val homeScreenType = when (getWindowSize(maxWidth)) {
```
JolandaVerhoef, line 120: Where's `maxWidth` coming from?

JolandaVerhoef:
Is there a reason that you included the migration from `List<Post>` to `PostFeed` in this PR? It's making it difficult to follow the changes related to that migration vs the actual changes made for the large screen support.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
manuelvicnt: From `BoxWithConstraints`

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
alexvanyo: Updated! Thanks for the explanation offline as well for why we want to do this!

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
alexvanyo: Updated! This seems like it shouldn't make too much of a difference, since `getWindowSize` should be pretty cheap.

alexvanyo:
@JolandaVerhoef Ah I apologize, that ended up being a refactor that happened partway through while cleaning everything up. Ideally that should have been pulled out into a separate PR that didn't touch the UI at all.

That ended up being changed to more nicely support having the topmost article selected by default. Previously, the post list was expecting a hardcoded list that was pulled apart into different sections, so the 4th post was arbitrarily chosen to be the "highlighted item." With the refactor to `PostsFeed`, those sections are now specified by the data object, so we can nicely choose the "highlighted item" to be selected by default.

JolandaVerhoef (approved):

alexvanyo:
Thanks @manuelvicnt and @JolandaVerhoef !

PR #632, opened 2021-08-30
[Jetnews] Add WindowSize breakpoints

manuelvicnt:


JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
```diff
@@ -0,0 +1,35 @@
...
+ *     - Compact: Most phones in portrait mode
+ *     - Medium: Most foldables and tablets in portrait mode
+ *     - Expanded: Most tablets in landscape mode
+ *
+ * More info: https://material.io/archive/guidelines/layout/responsive-ui.html
+ */
+enum class WindowSize { Compact, Medium, Expanded }
+
+fun getWindowSize(width: Dp): WindowSize = when (width.value) {
+    in 0f..599f -> WindowSize.Compact
```
alexvanyo, line 32: Since these are float ranges, anything greater than `599` but smaller than `600` (like `599.5`) will fall through to the `Expanded` case.

It might be clearer to just do a cascading when where the edges are clear:

```kotlin
fun getWindowSize(width: Dp): WindowSize = when {
    width < 600.dp -> WindowSize.Compact
    width <= 800.dp -> WindowSize.Medium
    else -> WindowSize.Expanded
}
```

alexvanyo (changes requested):
Just a small issue with the current implementation. It would be good to have a few tests for the function, just to encode the desired behavior especially around the breakpoints.

JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
manuelvicnt: Yeah, sounds good. improved this and added tests

JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
```diff
@@ -0,0 +1,40 @@
...
+ *     - Compact: Most phones in portrait mode
+ *     - Medium: Most foldables and tablets in portrait mode
+ *     - Expanded: Most tablets in landscape mode
+ *
+ * More info: https://material.io/archive/guidelines/layout/responsive-ui.html
+ */
+enum class WindowSize { Compact, Medium, Expanded }
+
+fun getWindowSize(width: Dp): WindowSize =
+    if (width.value >= 0f) {
```
alexvanyo, line 32: Nit: you could put this into the `when` as well: `width.value < 0f -> throw IllegalArgumentException(...)`

alexvanyo (approved):
Looks good, just one optional nit.

JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
manuelvicnt: Done! Thanks

Issue #634, opened 2021-08-30
JetCaster crash when click on "follow" with talkback


FannyDemey:
Hi 

I've tested Jetcaster app with Talkback to make accessibility improvement. I came across this bug that makes the app crash systematically : 
Way to reproduce : 
- Start the app using Talkback
- When focus is on a button to follow a new podcast
- Then double tap.
- RESULT : => app crash with this message and we are not able to restart the app (we have to remove data to be able to do so)
`2021-08-30 11:10:10.025 19023-19023/com.example.jetcaster E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.example.jetcaster, PID: 19023
    java.lang.NullPointerException
        at com.google.accompanist.pager.Pager$Pager$semanticsAxisRange$1$1.invoke(Pager.kt:421)
        at com.google.accompanist.pager.Pager$Pager$semanticsAxisRange$1$1.invoke(Pager.kt:234)
        at androidx.compose.ui.platform.AndroidComposeViewAccessibilityDelegateCompat.populateAccessibilityNodeInfoProperties(AndroidComposeViewAccessibilityDelegateCompat.android.kt:647)
        at androidx.compose.ui.platform.AndroidComposeViewAccessibilityDelegateCompat.createNodeInfo(AndroidComposeViewAccessibilityDelegateCompat.android.kt:297)
        at androidx.compose.ui.platform.AndroidComposeViewAccessibilityDelegateCompat.access$createNodeInfo(AndroidComposeViewAccessibilityDelegateCompat.android.kt:101)
        at androidx.compose.ui.platform.AndroidComposeViewAccessibilityDelegateCompat$MyNodeProvider.createAccessibilityNodeInfo(AndroidComposeViewAccessibilityDelegateCompat.android.kt:2273)
        at android.view.AccessibilityInteractionController$AccessibilityNodePrefetcher.prefetchAccessibilityNodeInfos(AccessibilityInteractionController.java:1034)
        at android.view.AccessibilityInteractionController.findAccessibilityNodeInfoByAccessibilityIdUiThread(AccessibilityInteractionController.java:341)
        at android.view.AccessibilityInteractionController.access$400(AccessibilityInteractionController.java:75)
        at android.view.AccessibilityInteractionController$PrivateHandler.handleMessage(AccessibilityInteractionController.java:1393)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
[...]

PR #635, opened 2021-08-30
[JetCaster] fix: update pager dependency

FannyDemey:
to fix crash with HorizontalPager used with Talkback (Issue #634)


JolandaVerhoef:
Update has been done as part of updating all samples to newest dependencies.

PR #636, opened 2021-08-30
[Jetnews] Add NavRail for large screens

manuelvicnt:
Add `NavRail` to the different screens in Jetnews. When the `NavRail` appears, the `Drawer` and possible `BottomBar` layouts disappear. This happens for foldable and tablet layouts.

Screenshot_1630336661
Screenshot_1630336666
Screenshot_1630511690


JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
```diff
@@ -53,22 +58,40 @@ fun JetnewsApp(
...
+                Scaffold(
+                    scaffoldState = scaffoldState,
+                    // If the window size is Compact, show the AppDrawer.
+                    // Otherwise, a NavRail will be shown in individual screens
+                    drawerContent = if (windowSize == WindowSize.Compact) {
+                        {
+                            AppDrawer(
+                                currentRoute = currentRoute,
+                                navigateToHome = {
+                                    navController.navigate(MainDestinations.HOME_ROUTE)
```
alexvanyo, line 73: Navigating here without popping up keeps adding the home and interests to the backstack, which I don't think is the behavior we want.

alexvanyo: Ah, looking at the old behavior this actually was the same behavior that was there before.

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
```diff
@@ -53,22 +58,40 @@ fun JetnewsApp(
...
+                                }
+                            )
+                        }
+                    } else {
+                        null
+                    }
+                ) { innerPaddingModifier ->
+                    JetnewsNavGraph(
+                        appContainer = appContainer,
+                        showNavRail = windowSize != WindowSize.Compact,
```
alexvanyo, line 89: This is just an approach thing, but what made you decide to pass down the resolved `showNavRail` here, versus the `windowSize`?

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -94,80 +96,114 @@ class TabContent(val section: Sections, val content: @Composable () -> Unit)
...
+    tabContent: List<TabContent>,
+    currentSection: Sections,
+    showNavRail: Boolean,
+    onTabChange: (Sections) -> Unit,
+    navigateToHome: () -> Unit,
+    openDrawer: () -> Unit,
+    scaffoldState: ScaffoldState
+) {
+    if (showNavRail) {
+        Row(Modifier.fillMaxSize()) {
```
alexvanyo, line 152: I don't know the implications of this, but is there a difference between putting this `if` as close as possible to the content that is removed? For example, I think we could always have a `Row`, and then only have one place where `InterestsScreen` is defined (probably with an `if` statement in the arguments for the `navigationIconContent`).

Is there an advantage one way or the other? Or just situational based on other requirements (animations, etc.)

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
```diff
@@ -53,22 +58,40 @@ fun JetnewsApp(
...
-                drawerContent = {
-                    AppDrawer(
-                        currentRoute = currentRoute,
-                        navigateToHome = { navController.navigate(MainDestinations.HOME_ROUTE) },
-                        navigateToInterests = { navController.navigate(MainDestinations.INTERESTS_ROUTE) },
-                        closeDrawer = { coroutineScope.launch { scaffoldState.drawerState.close() } }
+
+            BoxWithConstraints {
+                val windowSize = getWindowSize(maxWidth)
+                Scaffold(
```
alexvanyo, line 64: The approach looks good to me overall, although it made me notice this `Scaffold` here, meaning there's two scaffolds in play. If we put this `DrawerContent` in each screen, I think we would lose the nice animation of destinations switching out while the drawer closes.

I also noticed that you can still open the drawer on the article detail screen, which seems a little strange.

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
JoseAlcerreca: +1 to keeping things primitive

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
manuelvicnt: Yeah, let's fix that in a separate PR. Thanks!

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
manuelvicnt: `showNavRail` is a better API in this case as it's clear what's the intend behind it vs `windowSize` which is not clear what the composable should behave. It's also easier for testing and previewing this composable.

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
manuelvicnt: The point of having the if statement above `Row` is to avoid the `Row` composable when it's not needed (i.e. in Compact mode).

This might be a micro-optimization though, but I think this reads better than using always the `Row` and having to call the `InterestsScreen` like:

```
InterestsScreen(
    tabContent = tabContent,
    tab = currentSection,
    onTabChange = onTabChange,
    scaffoldState = scaffoldState,
    navigationIconContent = if (!showNavRail) { {
        IconButton(onClick = openDrawer) {
            Icon(
                painter = painterResource(R.drawable.ic_jetnews_logo),
                contentDescription = stringResource(
                    R.string.cd_open_navigation_drawer
                ),
                tint = MaterialTheme.colors.primary
            )
        }
    } } else {
        null
    }
)
```

WDYT?

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
manuelvicnt: Nothing has changed regarding the Drawer behavior, it should work as it used to be. We just hide it when the `NavRail` needs to be shown.

manuelvicnt: Being able to open the drawer from the article screen used to happen as well before, let's try to tackle that in a separate CL

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
manuelvicnt: What do you mean with "primitive" @JoseAlcerreca?

parishsu:
screenshot LGTM - one nit not related to nav rail - the "bookmark" icon seem to be a different color (on the nav rail they should all be "high emphasis")

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
alexvanyo: Makes sense right now, although this is an interesting quirk of Compose I don't think I was aware of until now:

Let's say that `InterestsScreen` had some internal state (animations, scroll state, etc.)

Having it defined twice in separate locations means that that state would be reset if the one that is displayed switches. Having them in the same place (defined once, with the `if` statements within the declaration) I believe means that internal state will be preserved.

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
alexvanyo: I might be reading ahead to the list/detail case, but keeping it as the `windowSize` allows any route to decide individually what to do based on the `windowSize`, instead of being told what it has to do. I don't feel super strongly though either way.

JetNews/app/src/main/java/com/example/jetnews/ui/components/AppNavRail.kt
```diff
@@ -0,0 +1,141 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.example.jetnews.R
+import com.example.jetnews.ui.MainDestinations
+import com.example.jetnews.ui.theme.JetnewsTheme
+import com.google.accompanist.insets.systemBarsPadding
+
+@Composable
+fun JetnewsNavRail(
```
alexvanyo, line 48: Just the comment from offline, that we should be able to use the prebuilt `NavigationRail` component here.

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -64,30 +65,33 @@ import com.example.jetnews.data.posts.impl.BlockingFakePostsRepository
...
     articleViewModel: ArticleViewModel,
+    showNavRail: Boolean,
     onBack: () -> Unit
 ) {
     // UiState of the ArticleScreen
     val uiState by articleViewModel.uiState.collectAsState()
 
     if (uiState.post != null) {
-        ArticleScreen(
+        ArticleAdaptiveScreen(
```
alexvanyo, line 92: I agree that it probably makes sense to keep this as just `ArticleScreen`, `Adaptive` should be the rule, not the exception! The `Route`/`Screen` naming difference should help distinguish between these overloads though.

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
manuelvicnt: Let's leave it like this for this PR, and if there are reasons for change it in the future, let's do so. I prefer to keep it like this for this particular PR because of the reasons mentioned above

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
manuelvicnt: Very good point, I need to change this! Thank you

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
manuelvicnt: Makes sense! Thank you

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -64,22 +65,25 @@ import com.example.jetnews.data.posts.impl.BlockingFakePostsRepository
...
- * Displays the Article screen.
+ * Stateful composable that displays the Navigation route for the Article screen.
  *
  * @param articleViewModel ViewModel that handles the business logic of this screen
+ * @param showNavRail (state) whether the Drawer or NavigationRail needs to be shown
  * @param onBack (event) request back navigation
  */
 @Composable
-fun ArticleScreen(
+fun ArticleRoute(
```
alexvanyo, line 84: 👍

JetNews/app/src/main/java/com/example/jetnews/ui/components/AppNavRail.kt
manuelvicnt: Added thanks! PTAL

alexvanyo (changes requested):
device-2021-09-01-114645

Two small issues here:
- In dark mode, the elevation of the toolbar doesn't quite match the elevation of the nav rail, which looks a little strange. I wonder if we want to have those match?
- When there's a vertical navigation bar, the padding isn't being handled properly with the nav bar (it looks like this is also an existing issue for the drawer)

manuelvicnt:
Thanks @alexvanyo ! Fixed the issue with the elevation in dark mode. For the other one, we should probably create issues in the repo in case someone wants to fix them before we can spend time on it. The other navigation issue might be worth adding there as well.

alexvanyo (approved):
@manuelvicnt Sounds good! I'm going to have to take the insets into account for the list/detail view, so I will probably be able to handle it there when adding in the `NavRail` as well.

This looks good to me to merge, to keep progress going!

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
JoseAlcerreca: I mean I prefer to pass a well-named boolean instead of a windowSize, whose reason to be there is not clear.

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -181,26 +229,76 @@ fun ArticleScreen(
...
  * Bottom bar for Article screen
  *
- * @param post (state) used in share sheet to share the post
  * @param onUnimplementedAction (event) called when the user performs an unimplemented action
  * @param isFavorite (state) if this post is currently a favorite
  * @param onToggleFavorite (event) request this post toggle it's favorite status
+ * @param onSharePost (event) request this post to be shared
+ */
+@Composable
+private fun InterestsNavRail(
```
JoseAlcerreca, line 245: Isn't this the article nav rail?

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -104,25 +108,84 @@ fun ArticleScreen(
...
+                onToggleFavorite = onToggleFavorite,
+                onSharePost = { sharePost(post, context) }
+            )
+        }
+        ArticleScreenContent(
+            post = post,
+            // Allow opening the Drawer if the NavRail is not on the screen
+            navigationIconContent = if (!showNavRail) {
+                {
+                    IconButton(onClick = onBack) {
```
JoseAlcerreca, line 149: For some reason this onClick is only working for the IconButton, but not the Icon inside. You have to click outside of the arrow for it to work!

It's strange, Icon is setting semantics but not clickable anywhere.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
```diff
@@ -444,24 +504,50 @@ private fun PostListDivider() {
...
+fun PreviewHomeScreenWithNavRail() {
+    val posts = runBlocking {
+        (BlockingFakePostsRepository().getPosts() as Result.Success).data
+    }
+    JetnewsTheme {
+        Surface {
+            HomeScreen(
+                uiState = HomeUiState(posts = posts),
+                showNavRail = true,
+                onToggleFavorite = { /*TODO*/ },
```
JoseAlcerreca: nit: This should be NO-OP instead of TODO (it's a keyword AS uses for actual to-dos)

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -427,12 +449,50 @@ private fun TopicDivider() {
     )
 }
 
+@Composable
+private fun rememberTapContent(
```
JoseAlcerreca, line 453: typo

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
manuelvicnt: Yeah! Agreed, so no changes required here

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -91,83 +93,113 @@ enum class Sections(@StringRes val titleResId: Int) {
...
-
-    val publicationSection = TabContent(Sections.Publications) {
-        val selectedPublications by interestsViewModel.selectedPublications.collectAsState()
-        PublicationList(
-            publications = uiState.publications,
-            selectedPublications = selectedPublications,
-            onPublicationSelect = { interestsViewModel.togglePublicationSelected(it) }
-        )
+    val tabContent = rememberTapContent(interestsViewModel)
+    val (currentSection, updateSection) = rememberSaveable {
```
JoseAlcerreca, line 113: What is the purpose of this? If I go to the people tab, press home and back to interests, I see topics tab.

JoseAlcerreca (changes requested):
I'm seeing rendering issues on the Freeform emulator but I think it's because of the emulator. Can you confirm nothing weird happens when you click on one of the interests?

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
manuelvicnt: Good catch! Thank you

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreen.kt
manuelvicnt: 👍  changed all of them to `NO-OP`

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
manuelvicnt: Good catch! Fixed thanks

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
manuelvicnt: It's working locally™ for me. Does it happen with other emulators?

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
manuelvicnt: This works currently for configuration changes on this screen. But it's true that it's not preserved when going to a different screen. Probably we can tackle this in a separate PR. As discussed offline, this could be hoisted in `JetnewsNavGraph`, for example, or in the backStackEntry of the destination

JoseAlcerreca (approved):

PR #639, opened 2021-09-02
[Jetnews] Interests screen LS support

manuelvicnt:
Adapting the interests screen to large UIs according to mocks.

Screenshot_1630573010
Screenshot_1630573015
Screenshot_1630573024
Screenshot_1630573029
Screenshot_1630573042


JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -349,27 +333,51 @@ private fun TabWithTopics(
...
+    BoxWithConstraints {
+        val windowSize = remember(maxWidth) { getWindowSize(maxWidth) }
+        val columns = remember(windowSize) { if (windowSize == WindowSize.Compact) 1 else 2 }
+
+        val itemMaxWidth = rememberItemMaxWidth(windowMaxWidth = maxWidth, columns = columns)
+        val groupedSections: Array<MutableList<InterestSection>> = remember(columns) {
+            val sectionsInColumns = Array<MutableList<InterestSection>>(columns) { mutableListOf() }
+            sections.forEachIndexed { index, section ->
+                val column = index % columns
+                sectionsInColumns[column] += section
```
JolandaVerhoef, line 349: This bit could do with a comment

JolandaVerhoef (approved):
Two minor comments about ripples, other than that it looks good.

One suggestion; add another preview for a large-screen version with only 1 column.

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -420,13 +434,70 @@ private fun TopicItem(itemTitle: String, selected: Boolean, onToggle: () -> Unit
...
+                MaterialTheme.colors.onSurface.copy(alpha = 0.8f)
+            }
+            Tab(
+                selected = selectedTabIndex == index,
+                onClick = {
+                    updateSection(tabContent.section)
+                },
+                modifier = Modifier
+                    .heightIn(min = 48.dp)
+                    .padding(horizontal = 16.dp, vertical = 12.dp)
```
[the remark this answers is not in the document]
manuelvicnt: Removed this padding because it's not doing anything and just messes up with the ripple

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
manuelvicnt: Done, thanks!

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -371,27 +359,57 @@ private fun TabWithTopics(
...
-                    style = MaterialTheme.typography.subtitle1
-                )
+    BoxWithConstraints {
+        val windowSize = remember(maxWidth) { getWindowSize(maxWidth) }
+        val columns = remember(windowSize) { if (windowSize == WindowSize.Compact) 1 else 2 }
+
+        val itemMaxWidth = rememberItemMaxWidth(windowMaxWidth = maxWidth, columns = columns)
+        // Group interest sections given the number of columns that fill the screen.
+        // Sections are distributed evenly on the screen using the modulus operator.
+        val groupedSections: Array<MutableList<InterestSection>> = remember(columns) {
```
alexvanyo, line 373: I think this can be simplified as is pretty easily:

```
        val groupedSections = sections
            .withIndex()
            .groupBy({ it.index % columns }) { it.value }
            .values
```

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -371,27 +359,57 @@ private fun TabWithTopics(
...
+            horizontalArrangement = Arrangement.SpaceEvenly,
+            modifier = Modifier.navigationBarsPadding().fillMaxSize()
+        ) {
+            groupedSections.forEach { sectionsInGroup ->
+                LazyColumn(
+                    Modifier
+                        .widthIn(max = itemMaxWidth)
+                        .padding(horizontal = if (columns > 1) 8.dp else 0.dp)
+                ) {
+                    sectionsInGroup.forEach { (section, topics) ->
```
alexvanyo, line 392: One thing that's a little strange with the current logic: the number of items in each column isn't really balanced. If all of the even sections have 20 items and all of the odd sections have 2 items, then they could be really unbalanced, which is a bit strange.

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -442,13 +466,67 @@ private fun TopicItem(itemTitle: String, selected: Boolean, onToggle: () -> Unit
...
+        }
+    }
+}
+
+/**
+ * Returns the max width for a Topic Item given maxWidth and number of columns constraints.
+ */
+@Composable
+private fun rememberItemMaxWidth(windowMaxWidth: Dp, columns: Int) =
+    remember(windowMaxWidth, columns) {
```
alexvanyo, line 518: Are these `remember` a bit of a premature optimization? It seems like this logic should be pretty simple to compute, and it avoids forgetting a necessary key here.

alexvanyo (commented):
Overall looks good, just a couple small comments.

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
manuelvicnt: Sure, fixed this. As the algorithm is more expensive if we also try to find the shortest column, I moved the execution to the background thread 👍

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
manuelvicnt: I don't think they are to be honest. Sure, the logic is simple to compute but even like that, I feel like that logic shouldn't be re-executed on every recomposition (as that could happen quite often). As the result of this computation is not likely to change often, I prefer to `remember` it

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -510,6 +507,46 @@ private fun InterestsTabRow(
...
+/**
+ * Return state for interest sections grouped given the number of columns that fill the screen.
+ * Sections are distributed to balance the number of items on each column.
+ */
+@Composable
+private fun sectionsGroupedInColumnsAsState(
+    sections: List<InterestSection>,
+    columns: Int,
+    defaultDispatcher: CoroutineDispatcher = Dispatchers.Default
+): State<Array<MutableList<InterestSection>>> = produceState(
```
alexvanyo, line 486: We should definitely avoid putting a `Mutable***` object into a `State` (even though we won't modify it elsewhere).

If that list were to change, Compose wouldn't be able to track it.

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -510,6 +507,46 @@ private fun InterestsTabRow(
...
+    initialValue = arrayOf(),
+    keys = arrayOf(sections, columns)
+) {
+    withContext(defaultDispatcher) { // Move the execution of this logic to the background thread
+        val sectionsInColumns = Array<MutableList<InterestSection>>(columns) { mutableListOf() }
+        // For each section, find the column with less elements and add it there
+        sections.forEach { section ->
+            val shortestColumnIndex = sectionsInColumns.getIndexShortestElement()
```
alexvanyo, line 494: I think we can avoid having to search through all of the existing `sectionsInColumns` if we keep a running tally of the number of elements in each column. I think that should make it a bit more efficient as well (on the order of `O(sections.size * columns)` instead of `O(sections.size^2 * columns)`)

If we do that, we probably don't need to put this in the background thread either? It shouldn't be substantially slower that the older grouping method.

manuelvicnt: Done thanks!

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
manuelvicnt: Done thanks!

alexvanyo (approved):
Looks good to me!

PR #641, opened 2021-09-03
[Jetnews] Large screen support for the Interests tab

manuelvicnt:
Screenshot_1630671753
Screenshot_1630671770


JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -479,34 +480,65 @@ private fun InterestsTabRow(
...
+    selectedTabIndex: Int,
+    updateSection: (Sections) -> Unit,
+    tabContent: List<TabContent>,
+    modifier: Modifier = Modifier
+) {
+    tabContent.forEachIndexed { index, content ->
+        val colorText = if (selectedTabIndex == index) {
+            MaterialTheme.colors.primary
+        } else {
+            MaterialTheme.colors.onSurface.copy(alpha = 0.8f)
```
JolandaVerhoef, line 525: Should this be an emphasis instead? What's the material guide for tabs?

JolandaVerhoef (approved):
One minor comment, looks good otherwise

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
JolandaVerhoef: As discussed, this code just moved, this alpha wasn't specifically added. We can address this in a different PR.

manuelvicnt: As discussed offline, let's leave it as is for now. Can revisit in the future, thank you!

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -479,34 +480,65 @@ private fun InterestsTabRow(
...
-        backgroundColor = MaterialTheme.colors.onPrimary,
-        contentColor = MaterialTheme.colors.primary
-    ) {
-        tabContent.forEachIndexed { index, tabContent ->
-            val colorText = if (selectedTabIndex == index) {
-                MaterialTheme.colors.primary
-            } else {
-                MaterialTheme.colors.onSurface.copy(alpha = 0.8f)
+    BoxWithConstraints {
+        val windowSize = remember(maxWidth) { getWindowSize(maxWidth) }
```
alexvanyo, line 484: Just one comment here: since this `maxWidth` is in a nested component, the size classes will be slightly off from the size classes at the root level since the entire screen isn't available at this point.

For making "root-level" type decisions, having a composition local seems like it might be the way to go...

manuelvicnt: Yeah, since the recommendation we want to give here is not 100% clear, I wanted to go for this implementation for now with the hope of revising all usages of `BoxWithConstraints` when we have a clearer guidance

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -479,34 +480,65 @@ private fun InterestsTabRow(
...
-                onClick = { updateSection(tabContent.section) },
-                modifier = Modifier.heightIn(min = 48.dp)
-            ) {
-                Text(
-                    text = stringResource(id = tabContent.section.titleResId),
-                    color = colorText,
-                    style = MaterialTheme.typography.subtitle2,
-                    modifier = Modifier.paddingFromBaseline(top = 20.dp)
-                )
+            else -> {
```
alexvanyo, line 495: Minor nit: I personally prefer to always add in all cases directly, rather than use an `else`. Once the exhaustive change lands in Kotlin, that will cause a compiler exception initially if a new size class is added or one is removed, which allows re-evaluating all of the decisions like this without having to search for them.

manuelvicnt: Done, thanks!

alexvanyo (approved):

PR #645, opened 2021-09-06
[Jetsnack] Snackbar support

manuelvicnt:
Display an error message when increasing an item quantity in the shopping cart every 5 times.

Screenshot_1630936030

### DIFF
--- a/Jetsnack/app/src/main/java/com/example/jetsnack/model/SnackbarManager.kt
+++ b/Jetsnack/app/src/main/java/com/example/jetsnack/model/SnackbarManager.kt
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.jetsnack.model
+
+import androidx.annotation.StringRes
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.update
+import java.util.UUID
+
+data class Message(val id: Long, @StringRes val messageId: Int)
+
+/**
+ * Class responsible for managing Snackbar messages to show on the screen
+ */
+object SnackbarManager {
+
+    private val _messages: MutableStateFlow<List<Message>> = MutableStateFlow(emptyList())
+    val messages: StateFlow<List<Message>> get() = _messages.asStateFlow()
+
+    fun showMessage(@StringRes messageTextId: Int) {
+        _messages.update { currentMessages ->
+            currentMessages + Message(
+                id = UUID.randomUUID().mostSignificantBits,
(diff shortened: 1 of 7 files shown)


Jetsnack/app/src/main/java/com/example/jetsnack/model/SnackbarManager.kt
```diff
@@ -0,0 +1,50 @@
...
+package com.example.jetsnack.model
+
+import androidx.annotation.StringRes
+import java.util.UUID
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.update
+
+data class Message(val id: Long, @StringRes val messageId: Int)
```
JolandaVerhoef, line 26: This will not allow the consumer to pass dynamic strings, or buttons with callbacks. That's fine for this sample, but we might want to think about a more generic approach.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -17,30 +17,65 @@
...
+            val currentMessages by SnackbarManager.messages.collectAsState()
+            if (currentMessages.isNotEmpty()) {
+                val message = currentMessages[0]
+                val messageText: String = stringResource(message.messageId)
+
+                // Effect running in a coroutine that displays the Snackbar on the screen
+                // If there's a change to messageText, SnackbarManager, or scaffoldState, the
+                // previous effect will be cancelled and a new one will start with the new values
+                LaunchedEffect(messageText, SnackbarManager, scaffoldState) {
+                    scaffoldState.snackbarHostState.showSnackbar(messageText)
```
JolandaVerhoef, line 74: Can you add a comment to explain that this call blocks (within the coroutine) until the Snackbar gets dismissed (either by the user or by the timeout). This is not really clear when reading this code, so it looks as though the ViewModel is directly notified when the snackbar starts showing.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/cart/CartViewModel.kt
```diff
@@ -27,18 +30,26 @@ import kotlinx.coroutines.flow.StateFlow
...
-        val currentCount = _orderLines.value.first { it.snack.id == snackId }.count
-        updateSnackCount(snackId, currentCount + 1)
+        if (increaseCountErrorCounter < 4) {
+            val currentCount = _orderLines.value.first { it.snack.id == snackId }.count
+            updateSnackCount(snackId, currentCount + 1)
+            increaseCountErrorCounter++
+        } else {
+            increaseCountErrorCounter = 0
+            snackbarManager.showMessage(R.string.cart_increase_error)
+        }
```
JolandaVerhoef, line 52: Can you use something similar to Jetnews failure counter? 
```
private fun shouldRandomlyFail(): Boolean = ++requestCount % 5 == 0
```

Jetsnack/app/src/main/java/com/example/jetsnack/model/SnackbarManager.kt
manuelvicnt: I think it's fine to leave it as is for now as we don't have more complex messages to show.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
manuelvicnt: Added more comments, thank you

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/cart/CartViewModel.kt
manuelvicnt: Improved logic, thanks!

JolandaVerhoef (approved):

PR #646, opened 2021-09-07
[Jetsnack] Add state holder to JetsnackApp

manuelvicnt:


Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
```diff
@@ -108,21 +109,57 @@ enum class HomeSections(
     PROFILE(R.string.home_profile, Icons.Outlined.AccountCircle, "home/profile")
 }
 
+/**
+ * Remembers and creates an instance of [BottomBarStateHolder]
+ */
+@Composable
+fun rememberBottomBarStateHolder(navController: NavController, tabs: Array<HomeSections>) =
+    remember(navController, tabs) {
```
JolandaVerhoef, line 117: Should this be `rememberSaveable`? If not, why?

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
```diff
@@ -108,21 +109,57 @@ enum class HomeSections(
...
+    val tabs: Array<HomeSections>
+) {
+    private val sections = HomeSections.values()
+    private val currentRoute: String?
+        get() = navController.currentDestination?.route
+
+    val routes = sections.map { it.route }
+
+    // Reading this attribute will cause recompositions when the route becomes valid or invalid
+    val isValidRoute: Boolean
```
JolandaVerhoef, line 136: What makes a route "valid"? Maybe add a comment to explain that the bottom bar should only be shown on specific pages, and this boolean checks if the current destination is in that set? Or even consider renaming to "shouldBeDisplayed" or something like that?

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
```diff
@@ -108,21 +109,57 @@ enum class HomeSections(
...
+        @Composable get() =
+            navController.currentBackStackEntryAsState().value?.destination?.route in routes
+
+    val currentSection: HomeSections
+        get() = sections.first { it.route == currentRoute }
```
JolandaVerhoef, line 141: Can this getter fail if the user is on a different route?

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
```diff
@@ -108,21 +109,57 @@ enum class HomeSections(
...
+
+    fun navigateToRoute(route: String) {
+        if (route != currentRoute) {
+            navController.navigate(route) {
+                launchSingleTop = true
+                restoreState = true
+                popUpTo(findStartDestination(navController.graph).id) {
+                    saveState = true
+                }
```
JolandaVerhoef, line 150: This part doesn't really make sense to me, maybe add a comment?

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
manuelvicnt: The only information worth saving is the `navController` which is already saved by the Navigation library. This information here is not that important or derived state

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
manuelvicnt: If this is called when shouldn't be, it'll throw an exception. I think it's fine

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
manuelvicnt: Not sure what this does, maybe @nickbutcher does

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
manuelvicnt: Done, thanks!

JolandaVerhoef (approved):

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
nickbutcher: I think that @JoseAlcerreca added this? Should we be passing the `navCotroller` around—wouldn't it be better to hoist params of whether to show bottom bar and where to navigate when selected?

manuelvicnt: Yeah, good point. I think it's better to create a `JetsnackAppStateHolder` and remove the `JetsnackNavGraph` composable as that's what takes the `navController` as a dependency and complicates things.

manuelvicnt: Made the changes and created the state holder for the App which makes much more sense. PTAL! Thank you

Also added more comments

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
```diff
@@ -110,19 +105,17 @@ enum class HomeSections(
 
 @Composable
 fun JetsnackBottomBar(
-    navController: NavController,
+    shouldBeShown: Boolean,
```
alexvanyo, line 108: Just trying to get my bearings for idiomatic Compose code: what you think about avoiding this parameter, and instead just wrapping the `JetsnackBottomBar` in an `if`? To me that feels a little bit closer to the "if you don't want to show a composable, just don't call it" idea.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackAppStateHolder.kt
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/**
+ * Remembers and creates an instance of [AppStateHolder]
+ */
+@Composable
+fun rememberAppStateHolder(
+    scaffoldState: ScaffoldState = rememberScaffoldState(),
+    navController: NavHostController = rememberNavController(),
+    snackbarManager: SnackbarManager = SnackbarManager,
```
alexvanyo, line 58: I didn't review the `SnackbarManager` PR, but having a global `object SnackbarManager` feels really strange, compared to just instantiating one here (in a `remember` I'd assume).

alexvanyo (approved):
Looks good, just two minor comments.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackAppStateHolder.kt
manuelvicnt: I don't want it to be scoped to the Composition. It's simulating having a "singleton" object that should be injected somehow, as if you had a repository

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/Home.kt
manuelvicnt: Yeah, good point! Fixed, thank you

JolandaVerhoef (approved):
I think this looks very promising. The only thing that scares me a bit is how scalable this would be. If everything happening for the top-level screen (snackbars, bottom bar state, navigation) is in one file, how quickly will that become too large? What would be our strategy then to split it up and keep with the Single Responsibility principle?

This does not need to be addressed in this PR, but I think it's a nice conversation to keep going.

manuelvicnt:
The strategy would be to split up if it makes sense and there are no dependencies between state holders 👍

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackAppStateHolder.kt
alexvanyo: Got it, I think what tripped me up is that it's a little strange that it's a parameter that has a default when there's exactly one object of that type (in other words, what else could you ever pass to the method). Maybe just having a singular place where the class is constructed would make sense? Don't feel very strongly about it though.

PR #649, opened 2021-09-09
Compose v1.0.2

JolandaVerhoef:


Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Button.kt
```diff
@@ -123,7 +123,7 @@ private fun ButtonPreview() {
 private fun RectangleButtonPreview() {
     JetsnackTheme {
         JetsnackButton(onClick = {}, shape = RectangleShape) {
-            Text(text = "Demo")
-        }
+        Text(text = "Demo")
+    }
```
manuelvicnt, line 127: Revert back?

JolandaVerhoef: This is actually done by SpotlessApply. Without it the build fails. I'll find or raise a bug for ktlint or spotless.

JolandaVerhoef: Found it! https://github.com/pinterest/ktlint/issues/1221

manuelvicnt: Cool! Should we revert the version update back until it's fixed?

JolandaVerhoef: Done

manuelvicnt (approved):

PR #650, opened 2021-09-10
[Jetnews] Better TopAppBar in large screens

manuelvicnt:
Screenshot 2021-09-10 at 10 36 42

### DIFF
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/components/AppNavRail.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/components/AppNavRail.kt
@@ -44,7 +44,6 @@ import androidx.compose.ui.unit.dp
 import com.example.jetnews.R
 import com.example.jetnews.ui.JetnewsDestinations
 import com.example.jetnews.ui.theme.JetnewsTheme
-import com.google.accompanist.insets.statusBarsPadding
 
 @OptIn(ExperimentalMaterialApi::class)
 @Composable
@@ -77,7 +76,7 @@ fun AppNavRail(
 ) {
     JetnewsNavRail(
         header = {
-            JetnewsIcon(Modifier.statusBarsPadding().padding(top = 8.dp))
+            JetnewsIcon()
         },
         modifier = modifier
     ) {
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
@@ -103,9 +103,8 @@ fun HomeRoute(
     scaffoldState: ScaffoldState
 ) {
     // Construct the lazy list states for the list and the details outside of deciding which one to
-    // show.
-    // This allows the associated state to survive beyond that decision, and therefore we get to
-    // preserve the scroll throughout any changes to the content.
+    // show. This allows the associated state to survive beyond that decision, and therefore
+    // we get to preserve the scroll throughout any changes to the content.
     val homeListLazyListState = rememberLazyListState()
     val articleDetailLazyListStates = when (uiState) {
         is HomeUiState.HasPosts -> uiState.postsFeed.allPosts
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
@@ -205,7 +205,6 @@ private enum class HomeScreenType {
     ArticleDetails
 }
 
-@Composable
 private fun getHomeScreenType(
     windowSize: WindowSize,
     uiState: HomeUiState
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
@@ -16,7 +16,6 @@
 
 package com.example.jetnews.ui.home
 
-import android.annotation.SuppressLint
 import androidx.activity.compose.BackHandler
 import androidx.compose.foundation.layout.BoxWithConstraints
 import androidx.compose.foundation.layout.padding
@@ -31,6 +30,7 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.key
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.composed
 import com.example.jetnews.ui.article.ArticleScreen
 import com.example.jetnews.utils.WindowSize
 import com.example.jetnews.utils.getWindowSize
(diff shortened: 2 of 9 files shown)


alexvanyo:
There's a few issues with regards to insets with the new app bar:

- List:

When the list and nav bar is visible, it looks like there is no system bar padding on the top, so the search box is underneath the status bar. The swipe-to-refresh is also not padded against the status bar.



- List + Detail:

When the list and detail are both shown, the swipe-to-refresh is not padded against the status bar.
Also, the navigation bar padding is applied twice to the post content, and for both the list and post content we lose the nice edge-to-edge behavior where the content is visible underneath the navigation bar.



Also, with a vertical navigation bar (on either side), the navigation bar padding is applied twice, which leaves an awkward gap.

alexvanyo (changes requested):
Marking as request changes for the insets.

manuelvicnt:
Thanks @alexvanyo! Insets are killing me! Updated and tested all the different possibilities. Thank you!

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
```diff
@@ -210,3 +204,40 @@ private enum class HomeScreenType {
     Feed,
     ArticleDetails
 }
+
+@Composable
```
alexvanyo, line 208: Nit: This doesn't need to be `@Composable`, right?

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
```diff
@@ -210,3 +204,40 @@ private enum class HomeScreenType {
...
+        .statusBarsPadding()
+        .padding(
+            rememberInsetsPaddingValues(
+                insets = LocalWindowInsets.current.navigationBars,
+                applyStart = false,
+                applyBottom = false
+            )
+        )
+} else {
+    Modifier
```
alexvanyo, line 242: When we are not showing the navrail, we still need to apply the insets to the home screens:

We want to apply the bottom system bars to both the list and the detail.
When we are only showing one, we want to apply both the start and end system bars
When we are showing both, we want to apply the start to the list (but not the end), and we want to apply the end to the detail (but not the start).

JetNews/app/src/main/java/com/example/jetnews/ui/components/AppNavRail.kt
```diff
@@ -77,7 +76,7 @@ fun AppNavRail(
 ) {
     JetnewsNavRail(
         header = {
-            JetnewsIcon(Modifier.statusBarsPadding().padding(top = 8.dp))
```
alexvanyo, line 80: This causes the `JetNews` icon to be obscured by the status bar when on the interests screen.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
manuelvicnt: Good catch! It used to be but not after I refactored the code thanks

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
```diff
@@ -112,21 +126,17 @@ fun HomeFeedWithArticleDetailsScreen(
         openDrawer = openDrawer,
         navigateToInterests = navigateToInterests,
         homeListLazyListState = homeListLazyListState,
-        scaffoldState = scaffoldState
-    ) { hasPostsUiState, modifier ->
-        Row {
+        scaffoldState = scaffoldState,
+        modifier = modifier,
+    ) { hasPostsUiState, contentModifier ->
+        Row(Modifier.padding(top = 8.dp)) {
```
alexvanyo, line 132: This top padding looks a little weird when scroll, since the content doesn't scroll out to edge of the unobscured content.
Should we push this padding into an inner component?

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
manuelvicnt: I prefer to cover these insets work on #655 since it changes a lot things

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
```diff
@@ -558,6 +543,108 @@ private fun PostListDivider() {
...
+        elevation = 4.dp,
+        modifier = modifier.fillMaxWidth()
+    ) {
+        Row(
+            verticalAlignment = Alignment.CenterVertically,
+            modifier = Modifier.padding(horizontal = 8.dp)
+        ) {
+            CompositionLocalProvider(LocalContentAlpha provides ContentAlpha.medium) {
+                IconButton(onClick = { /* Functionality not supported yet */ },) {
+                    Icon(imageVector = Icons.Filled.Search, contentDescription = null)
```
alexvanyo, line 563: Nit: We should probably avoid a `null` content description here

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
```diff
@@ -210,3 +204,39 @@ private enum class HomeScreenType {
...
+                    HomeScreenType.Feed
+                }
+            }
+            is HomeUiState.NoPosts -> HomeScreenType.Feed
+        }
+    }
+    WindowSize.Expanded -> HomeScreenType.FeedWithArticleDetails
+}
+
+@SuppressLint("ComposableModifierFactory", "ModifierFactoryExtensionFunction")
```
alexvanyo, line 228: I'm not sure if this will survive with the insets changes, but I feel like we shouldn't be suppressing these lint checks if it does.

alexvanyo (approved):
The new app bar itself looks good with a few nits, as most of the insets complications will be handled in the next PR.

JetNews/app/src/main/java/com/example/jetnews/ui/components/AppNavRail.kt
manuelvicnt: True! Good catch, this is one of the things that are better handled in the other PR. Fixing it here, just in case

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
manuelvicnt: Good call, fixed

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
manuelvicnt: Yup! Added, had to do a bit amount of refactoring for the Article screen in the list/detail but it's looking good! Thanks

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
manuelvicnt: Added thanks!

PR #653, opened 2021-09-13
[JetNews] Fix state saving with drawer

alexvanyo:
Fixes the state saving (via `rememberSaveable`) throughout the app.

Per the design, we conditionally want to allow the drawer to be shown based on the size of the screen. With the current approach we conditionally passed `null` or the drawer content to a top-level `Scaffold`, which would then allow the drawer to be hidden or shown. If we don't want the drawer to be shown, we instead show the nav rail.

This approach has a subtle problem due to the [implementation of `Scaffold`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material/material/src/commonMain/kotlin/androidx/compose/material/Scaffold.kt;l=194;drc=0868ad90686faea3d86e395fb3758e8a1bbbeaac): when we switch the drawer content between `null` and non-null, we are changing the location of where the content lambda (containing everything else in the app) is called.

This changes all of the keys that `rememberSaveable` automatically uses, which is seen as scroll positions not being saved when switching between orientations that allow or disallow showing the drawer.

To workaround this, I replaced the top-level `Scaffold` with a `ModalDrawer` (since that was the only functionality we were using via the top-level `Scaffold` directly anyway). Crucially, this `ModalDrawer` is always called, even if we don't want to allow showing the drawer. This allows calling the content for the rest of the app in exactly one source location. To avoid showing the drawer when we want to show the nav rail instead, we switch between a real, remembered `DrawerState` and a fixed-to-closed drawer state.

### DIFF
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
@@ -17,16 +17,16 @@
 package com.example.jetnews.ui
 
 import androidx.compose.foundation.layout.BoxWithConstraints
-import androidx.compose.foundation.layout.padding
+import androidx.compose.material.DrawerState
+import androidx.compose.material.DrawerValue
 import androidx.compose.material.MaterialTheme
-import androidx.compose.material.Scaffold
-import androidx.compose.material.rememberScaffoldState
+import androidx.compose.material.ModalDrawer
+import androidx.compose.material.rememberDrawerState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.SideEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
-import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.navigation.compose.currentBackStackEntryAsState
 import androidx.navigation.compose.rememberNavController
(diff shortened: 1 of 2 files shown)


JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
```diff
@@ -54,42 +54,51 @@ fun JetnewsApp(
...
-                                    coroutineScope.launch { scaffoldState.drawerState.close() }
-                                }
-                            )
-                        }
-                    } else {
-                        null
-                    }
-                ) { innerPaddingModifier ->
+
+                val allowDrawerToBeShown = windowSize == WindowSize.Compact
```
manuelvicnt, line 68:
```suggestion
                val allowDrawerToBeShown = remember(windowSize) {
                    windowSize == WindowSize.Compact
                }
```

I'd probably remember this as this composable can be recomposed a lot of times due to `BoxWithConstraints`

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
```diff
@@ -54,42 +54,51 @@ fun JetnewsApp(
             val navigationActions = remember(navController) { JetnewsNavigationActions(navController) }
 
             val coroutineScope = rememberCoroutineScope()
-            // This top level scaffold contains the app drawer, which needs to be accessible
+            // This top level composable contains the app drawer, which needs to be accessible
             // from multiple screens. An event to open the drawer is passed down to each
             // screen that needs it.
-            val scaffoldState = rememberScaffoldState()
+            val drawerState = rememberDrawerState(DrawerValue.Closed)
```
manuelvicnt, line 60: We should move these variables closer where they're used. Shouldn't be above BWC

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
```diff
@@ -54,42 +54,51 @@ fun JetnewsApp(
...
-                        }
-                    } else {
-                        null
-                    }
-                ) { innerPaddingModifier ->
+
+                val allowDrawerToBeShown = windowSize == WindowSize.Compact
+
+                // Determine the drawer state to pass to the modal drawer.
+                val sizeAwareDrawerState = if (allowDrawerToBeShown) {
```
manuelvicnt, line 71: This looks like an implementation detail I might not want to see when looking at `JetnewsApp`. Should we move this logic to a separate private composable function that decides which drawer state to return?

```
@Composable
private fun rememberSizeAwareDrawerState(allowDrawerToBeShown: Boolean) {
    val drawerState = rememberDrawerState(DrawerValue.Closed)
    return if (allowDrawerToBeShown) ...
}
```

alexvanyo: Updated in the latest commit!

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
alexvanyo: Updated as part of the extraction into `rememberSizeAwareDrawerState`

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
alexvanyo: I'd hope this simple equality check isn't costly enough that it need to be remembered...
If we feel like we need to remember that, wouldn't we basically be trying to remember all logic that's in the main composable scope?

manuelvicnt: Sure, that's ok

manuelvicnt (approved):

alexvanyo:
Thanks @manuelvicnt !

PR #654, opened 2021-09-14
[JetNews] Convert navrail/drawer decision to use currentWindowMetrics

alexvanyo:
Converts the decision on whether to use the navrail or the drawer to use the current window metrics, as opposed to a top-level `BoxWithConstraints`.

There are three other places where we are currently calling `getWindowSize`:
- Deciding between which home screens to use
- Deciding what interests tab layout to use
- Deciding what interests to use

This PR doesn't change any of those places, but we may want to adjust some (or all) of them to use the current window metrics instead of nested `BoxWithConstraints`.

We also could pass down the window size class manually, or have each of those places individually call `rememberWindowSizeState()`.

### DIFF
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
@@ -16,7 +16,9 @@
 
 package com.example.jetnews.ui
 
-import androidx.compose.foundation.layout.BoxWithConstraints
+import android.app.Activity
+import android.content.Context
+import android.content.ContextWrapper
 import androidx.compose.material.DrawerState
 import androidx.compose.material.DrawerValue
 import androidx.compose.material.MaterialTheme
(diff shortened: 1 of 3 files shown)


JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
```diff
@@ -51,45 +61,88 @@ fun JetnewsApp(
...
+    return currentWindowMetricsFlow.collectAsState(
+        WindowMetricsCalculator.getOrCreate().computeCurrentWindowMetrics(activity)
+    )
+}
+
+/**
+ * Remembers the [WindowSize] corresponding to the current window metrics.
+ */
+@Composable
+private fun rememberWindowSizeState(): WindowSize {
```
manuelvicnt, line 123: Nit: As this is the API used above, move it above `rememberCurrentWindowMetrics(): State<WindowMetrics>`?

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
```diff
@@ -51,45 +61,88 @@ fun JetnewsApp(
...
         }
     }
 }
 
+/**
+ * Returns the [WindowMetrics] corresponding to the current activities
+ * [WindowInfoRepository.currentWindowMetrics] as [State].
+ */
+@Composable
+private fun rememberCurrentWindowMetrics(): State<WindowMetrics> {
```
manuelvicnt, line 109: Nit: What about moving these helper functions to a separate file? These might come at some point from a Compose library

manuelvicnt (approved):

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
alexvanyo: Moved to `WindowSize.kt` in the latest commit!

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
alexvanyo: I updated the order in `WindowSize.kt`, where these were moved to.

PR #657, opened 2021-09-22
[JetNews] Adjust medium behavior

alexvanyo:
This PR refines the window size class handling, primarily by resolving that size class in one place, in the top-level `JetnewsApp`.

I also fixed the issue where the list and detail layout was being used for the `Medium` size class, whereas we want to show either the list or detail at the `Medium` breakpoint.

A couple notes:
- I am passing down both a `WindowSize` and a resolved `Boolean` flag in some places. My rationale there is making a certain decision in exactly once place, and passing down that resolved value. So, therefore we create `isDrawerActive` in exactly one place, and then pass it down, while also passing down the raw `WindowSize`, because some screen may make an unrelated decisions using the size class (1 or 2 columns somewhere).

- The `Medium` size class design is not complete yet, mostly due to the complications created by the nav bar with the article screen. We need to change the contents of the nav bar to match the designs, but we only can do that based on the state contained within the home route. It might, therefore, make sense to revert back to having a different nav bar in each destination.  I'm not planning to address that in this PR, but a later one.

### DIFF
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
@@ -76,7 +76,7 @@ import kotlinx.coroutines.runBlocking
  * Stateless Article Screen that displays a single post adapting the UI to different screen sizes.
  *
  * @param post (state) item to display
- * @param showNavRail (state) whether the Drawer or NavigationRail needs to be shown
+ * @param isDrawerActive (state) if the drawer is active
  * @param onBack (event) request navigate back
  * @param isFavorite (state) is this item currently a favorite
  * @param onToggleFavorite (event) request that this post toggle it's favorite state
@@ -101,7 +101,7 @@ fun ArticleScreen(
         val context = LocalContext.current
         ArticleScreenContent(
             post = post,
-            // Allow opening the Drawer if the NavRail is not on the screen
+            // Allow opening the Drawer if the drawer is active
             navigationIconContent = if (isDrawerActive) {
                 {
                     IconButton(onClick = onBack) {
@@ -115,7 +115,7 @@ fun ArticleScreen(
             } else {
                 null
             },
-            // Show the bottom bar if the NavRail is not on the screen
+            // Show the bottom bar if the drawer is active
             bottomBarContent = if (isDrawerActive) {
                 {
                     BottomBar(
@@ -188,40 +188,6 @@ private fun ArticleScreenContent(
                 .padding(innerPadding),
             state = lazyListState,
         )
-    }
-}
-
-/**
- * Bottom bar for Article screen
- *
- * @param onUnimplementedAction (event) called when the user performs an unimplemented action
- * @param isFavorite (state) if this post is currently a favorite
- * @param onToggleFavorite (event) request this post toggle it's favorite status
- * @param onSharePost (event) request this post to be shared
- */
-@Composable
-private fun ArticleNavRail(
-    onBack: () -> Unit,
-    onUnimplementedAction: () -> Unit,
-    isFavorite: Boolean,
-    onToggleFavorite: () -> Unit,
-    onSharePost: () -> Unit
-) {
-    JetnewsNavRail(
-        header = {
-            IconButton(onClick = onBack) {
-                Icon(
-                    imageVector = Icons.Filled.ArrowBack,
-                    contentDescription = stringResource(R.string.cd_navigate_up),
-                    tint = MaterialTheme.colors.primary
-                )
-            }
-        }
-    ) {
-        FavoriteButton(onClick = onUnimplementedAction)
-        BookmarkButton(isBookmarked = isFavorite, onClick = onToggleFavorite)
-        ShareButton(onClick = onSharePost)
-        TextSettingsButton(onClick = onUnimplementedAction)
     }
 }
 
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
@@ -85,7 +85,7 @@ import kotlinx.coroutines.runBlocking
 @Composable
 fun ArticleScreen(
     post: Post,
-    showNavRail: Boolean,
+    isDrawerActive: Boolean,
     onBack: () -> Unit,
     isFavorite: Boolean,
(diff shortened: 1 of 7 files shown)


JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -99,19 +99,10 @@ fun ArticleScreen(
 
     Row(modifier.fillMaxSize()) {
         val context = LocalContext.current
-        if (showNavRail) {
-            ArticleNavRail(
```
alexvanyo, line 103: This is the main TODO: We want to replace this NavRail with a blank one, other than a back button.

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
```diff
@@ -106,7 +112,13 @@ fun HomeRoute(
         }
     }
 
-    val homeScreenType = getHomeScreenType(showTopAppBar, uiState)
+    val homeScreenType = getHomeScreenType(windowSize, uiState)
+
+    val showTopAppBar = when (windowSize) {
```
alexvanyo, line 117: I moved this decision to here: Only this route cares about whether to show the top app bar, or switch it out.

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -287,13 +290,14 @@ private fun TabWithTopics(
  */
 @Composable
 private fun TabWithSections(
+    windowSize: WindowSize,
     sections: List<InterestSection>,
     selectedTopics: Set<TopicSelection>,
     onTopicSelect: (TopicSelection) -> Unit
 ) {
+    val columns = remember(windowSize) { if (windowSize == WindowSize.Compact) 1 else 2 }
```
alexvanyo, line 298: Just wanted to highlight this case:
We are keeping the `BoxWithConstraints`, because we want to base some decision around the local space available.
However, we are also making a decision based on the window size class, which we pass in.

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
manuelvicnt: What do you mean with the blank one? This is needed for the article immersive screen but we'd want to have something custom and hide the top-level one. It's probably ok to remove this code now as it's not getting called anyway

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -99,19 +99,10 @@ fun ArticleScreen(
...
-                isFavorite = isFavorite,
-                onToggleFavorite = onToggleFavorite,
-                onSharePost = { sharePost(post, context) }
-            )
-        }
         ArticleScreenContent(
             post = post,
             // Allow opening the Drawer if the NavRail is not on the screen
-            navigationIconContent = if (!showNavRail) {
+            navigationIconContent = if (isDrawerActive) {
```
manuelvicnt, line 105: Nit: update comment above

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -125,7 +116,7 @@ fun ArticleScreen(
                 null
             },
             // Show the bottom bar if the NavRail is not on the screen
-            bottomBarContent = if (!showNavRail) {
+            bottomBarContent = if (isDrawerActive) {
```
manuelvicnt, line 119: Nit: update comment above

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -175,14 +176,15 @@ fun InterestsScreen(
  */
 @Composable
 private fun TabContent(
+    windowSize: WindowSize,
```
manuelvicnt, line 179: Nit: include this param in the Kdoc

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -202,11 +204,12 @@ private fun TabContent(
  */
 @Composable
 private fun TopicList(
+    windowSize: WindowSize,
```
manuelvicnt, line 207: Nit: add this param to the kdoc

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
manuelvicnt: SG. In the two cases where we're using BWC at the moment, should we add a TODO saying: "We could optimize this code by removing BoxWithConstraints and use a custom Layout that places elements on the screen given the available width"?

JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
```diff
@@ -42,6 +43,7 @@ import androidx.window.layout.WindowMetricsCalculator
  */
 enum class WindowSize { Compact, Medium, Expanded }
 
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```
manuelvicnt, line 46: Nit: the default is already `PRIVATE`

manuelvicnt (commented):
Some nits but LGTM overall. Thanks!

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
alexvanyo: Added!

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
alexvanyo: Discussed offline for the strategy for medium screens. If you don't mind, I'll do all of that in a follow up PR after this one.

manuelvicnt (approved):

manuelvicnt:
Fine to merge and fix the MEDIUM layouts as the discussions we're having offline. It sounds like we might want to treat MEDIUM as COMPACT

PR #659, opened 2021-09-24
[Jetnews] Consolidate compact and medium layouts

manuelvicnt:
Screenshot_1632480644
Screenshot_1632480638
Screenshot_1632480627
Screenshot_1632480631

### DIFF
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
@@ -34,14 +34,14 @@ import com.example.jetnews.ui.article.ArticleScreen
  * Note: AAC ViewModels don't work with Compose Previews currently.
  *
  * @param homeViewModel ViewModel that handles the business logic of this screen
- * @param isDrawerActive (state) whether the drawer is active
+ * @param isExpandedScreen (state) whether the screen is expanded
  * @param openDrawer (event) request opening the app drawer
  * @param scaffoldState (state) state for the [Scaffold] component on this screen
  */
 @Composable
 fun HomeRoute(
     homeViewModel: HomeViewModel,
-    isDrawerActive: Boolean,
+    isExpandedScreen: Boolean,
     openDrawer: () -> Unit,
     scaffoldState: ScaffoldState = rememberScaffoldState()
 ) {
@@ -50,7 +50,7 @@ fun HomeRoute(
 
     HomeRoute(
         uiState = uiState,
-        isDrawerActive = isDrawerActive,
+        isExpandedScreen = isExpandedScreen,
         onToggleFavorite = { homeViewModel.toggleFavourite(it) },
         onSelectPost = { homeViewModel.selectArticle(it) },
         onRefreshPosts = { homeViewModel.refreshPosts() },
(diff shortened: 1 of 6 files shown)


JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
```diff
@@ -112,18 +106,12 @@ fun HomeRoute(
...
-        WindowSize.Medium, WindowSize.Expanded -> false
-    }
-
+    val homeScreenType = getHomeScreenType(isDrawerActive, uiState)
     when (homeScreenType) {
         HomeScreenType.FeedWithArticleDetails -> {
             HomeFeedWithArticleDetailsScreen(
                 uiState = uiState,
-                showTopAppBar = showTopAppBar,
+                showTopAppBar = isDrawerActive,
```
JolandaVerhoef, line 114: I'm a bit confused that you're passing on `isDrawerActive` instead of something like `isLargeScreen`. What's the logic you want to bring across? That the home screen type and top app bar visibility depend on whether the drawer is active, or on whether the screen is large? I'd say that second explanation feels more reasonable. So I'd consider passing a boolean called `isExpandedScreen`?

manuelvicnt: `isExpandedScreen` might be a better name now that we don't pass `windowSize`. WDYT @alexvanyo ? I'm not a fan of `showTopAppBar` because it doesn't actually represent what's going on

manuelvicnt: Changed the PR to use `isExpandedScreen` for the top level components + routes. Then move to a more descriptive name when made sense. Thanks!

JolandaVerhoef (approved):

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -287,22 +282,19 @@ private fun TabWithTopics(
...
-    windowSize: WindowSize,
     sections: List<InterestSection>,
     selectedTopics: Set<TopicSelection>,
     onTopicSelect: (TopicSelection) -> Unit
 ) {
-    val columns = remember(windowSize) { if (windowSize == WindowSize.Compact) 1 else 2 }
-
     // TODO: Optimize by manually positioning items, rather than using BoxWithConstraints
     BoxWithConstraints {
+        val columns = if (maxWidth < 600.dp) 1 else 2
```
alexvanyo, line 297: Just want to highlight this change: Since we're not quite following the designs with this consolidation, I think it makes sense to make this a local decision again. If/when we re-add the medium layout, we can probably return to the original logic.

alexvanyo (approved):
LGTM!

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
manuelvicnt: I'm working on removing BWC from the interests screen, you'll have a PR coming your way soon :D

manuelvicnt: Created #665 as a draft at the moment, but feel free to take a look

PR #667, opened 2021-09-28
[JetNews] Move window size calculation to activity level

alexvanyo:
This is a slight reworking of the window size class logic, with no functional changes.

The main change is that `rememberWindowSizeClass()` becomes an extension of the `Activity` which encourages its use at the top-level `@Composable` (that is, directly inside `setContent`). This deliberately makes it more difficult to fetch the window size class arbitrarily from nested composables (by looking up the activity), in favor of passing it down explicitly.

### DIFF
--- a/JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
@@ -17,16 +17,15 @@
 package com.example.jetnews.utils
 
 import android.app.Activity
-import android.content.Context
-import android.content.ContextWrapper
 import androidx.annotation.VisibleForTesting
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
+import androidx.compose.ui.geometry.Size
+import androidx.compose.ui.graphics.toComposeRect
 import androidx.compose.ui.platform.LocalConfiguration
-import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalDensity
-import androidx.compose.ui.unit.Dp
-import androidx.window.layout.WindowMetrics
+import androidx.compose.ui.unit.DpSize
+import androidx.compose.ui.unit.dp
 import androidx.window.layout.WindowMetricsCalculator
 
 /**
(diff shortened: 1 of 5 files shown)


JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
```diff
@@ -39,50 +38,44 @@ import androidx.window.layout.WindowMetricsCalculator
...
 
+/**
+ * Partitions a [DpSize] into a enumerated [WindowSize] class.
+ */
 @VisibleForTesting
-fun getWindowSize(width: Dp): WindowSize = when {
-    width.value < 0f -> throw IllegalArgumentException("Dp value cannot be negative")
-    width.value < 600f -> WindowSize.Compact
-    width.value < 840f -> WindowSize.Medium
+fun getWindowSizeClass(windowDpSize: DpSize): WindowSize = when {
```
manuelvicnt, line 45: This function shouldn't be annotated with `@VisibleForTesting`, right? It's used JetsnackApp

JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
```diff
@@ -39,50 +38,44 @@ import androidx.window.layout.WindowMetricsCalculator
...
-    val density = LocalDensity.current
-
-    return remember(windowMetrics, density) {
-        with(density) {
-            getWindowSize(windowMetrics.bounds.width().toDp())
-        }
+private fun Activity.rememberWindowSize(): Size {
+    val configuration = LocalConfiguration.current
+    // WindowMetricsCalculator implicitly depends on the configuration through the activity,
+    // so recalculate it upon changes.
```
manuelvicnt, line 59: Nit: "re-calculate"?

JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
```diff
@@ -39,50 +38,44 @@ import androidx.window.layout.WindowMetricsCalculator
...
  */
 @Composable
-private fun rememberCurrentWindowMetrics(): WindowMetrics {
-    val activity = LocalContext.current.findActivity()
-    val configuration = LocalConfiguration.current
-    return remember(activity, configuration) {
-        WindowMetricsCalculator.getOrCreate().computeCurrentWindowMetrics(activity)
-    }
-}
+fun Activity.rememberWindowSizeClass(): WindowSize {
```
manuelvicnt, line 70: Nit: Move this public function above the private one

manuelvicnt (commented):
Overall LGTM, added some nits. Thank you

JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
alexvanyo: I assume you mean `JetNewsApp`?

The only place it is used is in `WindowSize`, so `@VisibleForTesting` is correct.

JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
alexvanyo: Updated!

JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
alexvanyo: Done!

JetNews/app/src/main/java/com/example/jetnews/utils/WindowSize.kt
manuelvicnt: Ah ok ok! Makes sense, Thanks for moving it down

manuelvicnt (approved):

PR #673, opened 2021-10-04
[Jetcaster] Add navigation

manuelvicnt:


Jetcaster/app/src/main/java/com/example/jetcaster/ui/JetcasterAppState.kt
```diff
@@ -0,0 +1,73 @@
...
+    var isOnline by mutableStateOf(checkIfOnline())
+        private set
+
+    fun refreshOnline() {
+        isOnline = checkIfOnline()
+    }
+
+    @Suppress("DEPRECATION")
+    private fun checkIfOnline(): Boolean {
+        val cm = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
```
alexvanyo, line 63: Small nit: I like the `androidx.core` extension here, which avoids having to know the right `Context` constant:
```suggestion
        val cm = context.getSystemService<ConnectivityManager>()!!
```

alexvanyo (approved):
LGTM, just a small nit for code that was moved.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/JetcasterAppState.kt
JoseAlcerreca: But let's keep it nullable.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/JetcasterApp.kt
```diff
@@ -16,47 +16,34 @@
...
-private fun checkIfOnline(context: Context): Boolean {
-    val cm = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
-
-    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
-        val capabilities = cm.getNetworkCapabilities(cm.activeNetwork) ?: return false
-
-        capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) &&
-            capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
+fun JetcasterApp(
+    appState: JetcasterAppState = rememberJetcasterAppState()
```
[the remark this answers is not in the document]
manuelvicnt: We could, but as `JetcasterAppState` is also needed to handle the Navigation logic + state, I think it's fine to leave it here to keep all the logic centralised in one place and not need to create another state holder for the Activity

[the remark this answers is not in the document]
manuelvicnt: In fact, I'd rather have it here as this is not coupled to an Activity or Fragment. This logic could be also put in a screen level state holder, in that case you wouldn't pass down that information from the Activity level

Jetcaster/app/src/main/java/com/example/jetcaster/ui/JetcasterAppState.kt
manuelvicnt: Done! Thanks

PR #685, opened 2021-10-20
Remove PreviewParameterProvider from src/debug

### LINKED ISSUE #682: [JetNews] release build compile error
When switching to the release build there are compile errors around the PreviewParameters as the PostPreviewParameterProvider only exists in src/debug, as recommended in the tooling documentation.
What is the best practice for these ParameterProvider classes to only exist in the debug builds but still have functional release builds?

alexvanyo:
Moves the `PostPreviewParameterProvider` from `src/debug` to `src/main`. This allows release builds to successfully compile, even though it should be stripped by R8 since it will only be used for `@Preview` functions.

This fixes #682


JolandaVerhoef:
Fix looks good. But I'm worried that our CI didn't catch this failing. Should we consider building a release artifact during the CI run to make sure there aren't any issues doing that?

PR #686, opened 2021-10-21
Add assembleRelease to CI steps

alexvanyo:
Adds `assembleRelease` to the build steps for all projects.

This will ensure that changes can't be made that prevent `release` variants of apps from failing to compile, although it doesn't verify any runtime behavior (i.e. `release` variants could crash on startup).

### DIFF
--- a/.github/workflows/Crane.yaml
+++ b/.github/workflows/Crane.yaml
@@ -43,7 +43,7 @@ jobs:
 
       - name: Build project
         working-directory: ${{ env.SAMPLE_PATH }}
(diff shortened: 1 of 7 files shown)


alexvanyo:
@JoseAlcerreca The `paths` for all projects were configured to only run on changes to the subdirectories containing each project. This PR was just modifying the workflows, so nothing was changed in the previously defined `paths`.

In the last commit, I add a self-referential item so that changing the workflow for a project will also trigger the build for that project, which did the trick to actually kick off builds.

PR #695, opened 2021-11-03
Update R8 to avoid minified crash with navigation-compose

### LINKED ISSUE #692: [JetNews, JetSnack, Owl] The release version will crash if navigation-compose is used with minifyEnabled
When launching the release builds of Jetnews, Jetsnack, and Owl
the following crash occurs:

```
11-02 14:43:30.994 28787 28787 E AndroidRuntime: FATAL EXCEPTION: main
11-02 14:43:30.994 28787 28787 E AndroidRuntime: Process: com.example.owl, PID: 28787
11-02 14:43:30.994 28787 28787 E AndroidRuntime: java.lang.IllegalAccessError: Illegal class access: 'uu0$a' attempting to access 'androidx.navigation.compose.NavHostKt$PopulateVisibleList$1$1$observer$1' (declaration of 'uu0$a' appears in /data/app/~~NgrI5-Z5F9_KRcBP9LmPFA==/com.example.owl-bQbZtxrcUChvaK343C1SXw==/base.apk)
```

All these apps use navigation-compose and set minifyEnabled to true for release build.
My app also has the same issue.
But it works well when I use navigation-compose older than 2.4.0-alpha08.

alexvanyo:
As reported in #692 , using version `2.4.0-alpha09` or `2.4.0-alpha10` of `navigation-compose` with a minified build will result in a crash upon startup.

I created an upstream issue [here](https://issuetracker.google.com/issues/204910734) on the navigation side, but it looks like an R8 issue since updating to a newer version than default resolves the issue.

This PR adds that workaround, updating R8 for the 3 affected projects: `JetNews`, `Jetsnack`, and `Owl` (`Jetcaster` uses `navigation-compose`, but doesn't minify the release build). If `navigation-compose` changes its implementation to avoid this bug, or the default version of R8 fixes the issue, we can revert this commit.

This fixes #692

### DIFF
--- a/JetNews/build.gradle
+++ b/JetNews/build.gradle
@@ -23,9 +23,13 @@ buildscript {
     repositories {
         google()
         mavenCentral()
+        maven {
+            url 'https://storage.googleapis.com/r8-releases/raw'
+        }
     }
 
     dependencies {
+        classpath 'com.android.tools:r8:3.1.34' // Needed for release builds with navigation-compose 2.4.0-alpha10 (https://issuetracker.google.com/issues/204910734)
         classpath 'com.android.tools.build:gradle:7.0.3'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
--- a/Jetsnack/build.gradle
+++ b/Jetsnack/build.gradle
@@ -21,8 +21,12 @@ buildscript {
     repositories {
         google()
         mavenCentral()
+        maven {
+            url 'https://storage.googleapis.com/r8-releases/raw'
+        }
     }
     dependencies {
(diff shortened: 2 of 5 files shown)


JoseAlcerreca:
Isn't this fixed in 2.4.0 beta?

alexvanyo:
Just checked, and yep, updating to `2.4.0-beta02` seems to fix this issue (`2.4.0-beta01` still crashes)

I'll update all of the samples to the latest navigation version instead.

JoseAlcerreca:
I believe they were updated already!

JoseAlcerreca:
Correction: they weren't, and accompanist is using beta01. Is it urgent or can it wait for the next 1.1 release?

JoseAlcerreca (changes requested):
Abandon and update to nav beta02 or wait for next release

PR #703, opened 2021-11-11
[Jetsnack] Prevent recompose by passing the provider instead of the value

takahirom:
After watching the AndroidDevSummit, I learned that for values like scrolling, it helps to pass a Provider to prevent Recompose! It was a great session!
After watching the session, I went to look at the code and it didn't look like this practice had been adapted, so I adapted it.

https://youtu.be/zMKMwh9gZuI?t=1481

I checked to see if the Image and Title composers were recomposed with println().
Before the change, there was quite a lot of recompose, but after the change, there was no recompose on scroll.
Also, the animation seems to be fine, probably because the State is seen from the Layout via the lambda.

### DIFF
--- a/Jetsnack/app/src/main/java/com/example/jetsnack/ui/snackdetail/SnackDetail.kt
+++ b/Jetsnack/app/src/main/java/com/example/jetsnack/ui/snackdetail/SnackDetail.kt
@@ -102,8 +102,8 @@ fun SnackDetail(
         val scroll = rememberScrollState(0)
         Header()
         Body(related, scroll)
-        Title(snack, scroll.value)
-        Image(snack.imageUrl, scroll.value)
+        Title(snack = snack, scrollProvider = { scroll.value })
+        Image(imageUrl = snack.imageUrl, scrollProvider = { scroll.value })
         Up(upPress)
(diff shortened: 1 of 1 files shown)


CliffordHDiscovery:
GREAT find!

Please just smoketest this on TV as well... leanback has done some strangeeeeee things to composables before... 
I am particularly worried about the onFocus animation for when a card is selected. 
First version is in develop currently with its improvement in https://github.com/DNI-Mercury/dplus-android-client/pull/3918

nickbutcher:
Thank you for the PR. This fix currently sits in the [js-layout-perf](https://github.com/android/compose-samples/tree/nb/js-layout-perf) branch which also includes a utility for logging recompositions that I think it would be useful to include. Do you want to put up a PR merging that branch?

takahirom:
Thank you for telling me! I didn't know that branch 👀 
I was hoping there'd be somewhere where we could see the implementation of performance improvements, so if it's already there, then this PR doesn't need to be!
I didn't know how to use the inline Composable function to check Composition, so I learned that.
https://github.com/android/compose-samples/commit/23d9b04bf14ce8f055d4a78da1c690930fc29c83#diff-e83b192c5826252df92c898fa5492cb47fbb1f9af6440551ca4dd1e405038e36R33

Issue #698, opened 2021-11-05
[Crane] Google Maps Beta deprecated


iNoles:
By according to https://developers.google.com/maps/documentation/android-sdk/v310-beta, how we can proceed from this?

PR #709, opened 2021-11-29
Migrated Google Maps to 18.0.0.

jasonsaruulo:
#698 
Followed https://github.com/googlemaps/android-v3-migration.


bentrengrove:
Thank you for doing the work on this and upgrading maps for us, it is greatly appreciated. I have actually just replaced the maps component with the new Google Maps Compose library and so this is no longer needed though.

Check out #735 to see.

PR #723, opened 2022-01-07
When I scroll down the explore section the h6 black color disappear

rasoulian:
Hi guys,

I'm new in android development, when I debugged Crane sample , I saw the color of h6 is disappear sometimes.
I did some changes to keep the colors of h6, but I don't fix it the right way in my opinion.
I wish this pr can help you to fix this bug in the right ways.

Best wishes.


Dorian-MSF (commented):
good thanks

PR #729, opened 2022-01-26
[Jetnews] Replace custom WindowSizeClasses with material3-window-size-class

alexvanyo:
An update to Jetnews to replace the custom WindowSizeClass implementation with the library-provided ones from material3-window-size-class.


JetNews/app/build.gradle
```diff
@@ -97,6 +97,7 @@ dependencies {
     implementation "androidx.compose.foundation:foundation-layout:$compose_version"
     implementation "androidx.compose.material:material:$compose_version"
     implementation "androidx.compose.material:material-icons-extended:$compose_version"
+    implementation "androidx.compose.material3:material3-window-size-class:1.0.0-alpha10"
```
riggaroo, line 100: [nit] Crane has a dependency mechanism so I followed that pattern to extract the dependency, should we consider doing that too here? or at least extracting the version somewhere?
```
  implementation Libs.AndroidX.Compose.materialWindow
```

riggaroo (approved):

JetNews/app/build.gradle
alexvanyo: Yeah, I think in the long term having a standardized method throughout the Compose samples would be great for managing versions (maybe version catalogs?)

For right now though I think it makes sense to keep it as is since the other versions are inline below.

JolandaVerhoef: JetNews is the odd one out that doesn't use extracted dependencies. This has the benefit that we currently actually use the lint warnings for seeing which libraries have updates available. It also makes it easier for developers to copy-paste dependencies.

JolandaVerhoef (approved):

PR #731, opened 2022-02-02
[Jetcaster] Using isSeparating to check tableTop mode instead of HALF_OPENED

joyl1216:
The idea of the change is to use [isSeparating](window/window/src/main/java/androidx/window/layout/HardwareFoldingFeature.kt) to check which UI should be shown instead of HALF_OPENED.

Currently, in Jetcaster, the player screen will be cut off on the Microsoft Surface Duo device when the device is open (more than 180-degree hinge) until the device is folded to HALF_OPENED (less than 180). 
Screenshot_1643317559

With the change, the player screen will behave well on the Surface Duo device, no matter it is half-open, or fully open. 
Screenshot_1643317445

The behavior of Samsung Fold device would stay the same, showing the regular UI when the device is fully open and the tableTop UI at half-open.

### DIFF
--- a/Jetcaster/app/src/main/java/com/example/jetcaster/util/WindowInfoUtil.kt
+++ b/Jetcaster/app/src/main/java/com/example/jetcaster/util/WindowInfoUtil.kt
@@ -37,10 +37,10 @@ sealed interface DevicePosture {
 @OptIn(ExperimentalContracts::class)
 fun isTableTopPosture(foldFeature: FoldingFeature?): Boolean {
     contract { returns(true) implies (foldFeature != null) }
-    return foldFeature?.state == FoldingFeature.State.HALF_OPENED &&
+    return foldFeature?.isSeparating == true &&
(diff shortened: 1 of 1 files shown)


alexvanyo (approved):
Looks good to me, thanks for the contribution!

pfmaggi:
Thanks @joyl1216 for the contribution,
May be better to keep `isTableTopPosture` and `isBookPosture` as they are and add an extra check with the `isSeparating` method to account for dual screen devices.

The change you suggest break the connection between the methods name and what they actually do: on a duo they return always true independently from the actual posture, they just account for the hinge orientation.

joyl1216:
@pfmaggi Hi Pietro, thanks for the comment. It totally makes sense to keep `isTableTopPosture` and `isBookPosture`. I am happy to create a new pull request to introduce a new function called `isPostureSeparating` based on `isSeparating` and update the player screen with the new function. How does the new function name sound to you?

PR #735, opened 2022-02-07
[Crane] Update Crane to use new Google Maps library

### LINKED ISSUE #698: [Crane] Google Maps Beta deprecated

bentrengrove:
- Updated Crane to use the new Google Maps Compose library https://github.com/googlemaps/android-maps-compose
- Removed map lifecycle code that is no longer needed

Fixes #698

Screenshot_20220207_111918


Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
```diff
@@ -177,58 +173,69 @@ fun DetailsContent(
...
-    val cameraPosition = remember(latitude, longitude) {
+    val cityLocation = remember(latitude, longitude) {
         LatLng(latitude.toDouble(), longitude.toDouble())
     }
 
-    LaunchedEffect(map) {
-        val googleMap = map.awaitMap()
-        googleMap.addMarker { position(cameraPosition) }
-        googleMap.moveCamera(CameraUpdateFactory.newLatLng(cameraPosition))
+    cameraState.position = CameraPosition.fromLatLngZoom(
```
manuelvicnt, line 186: This will be executed on every recomposition. Is it really needed or can go in an effect? Regardless, maybe add a comment explaining it? Thanks

IIUC, this will reset a zoom that the user could've set to `InitialZoom`, and we might not want that, right?

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
```diff
@@ -177,58 +173,69 @@ fun DetailsContent(
...
-            googleMap.setZoom(mapZoom)
-            // Move camera to the same place to trigger the zoom update
-            googleMap.moveCamera(CameraUpdateFactory.newLatLng(cameraPosition))
-        }
+    val mapUiSettings = remember {
+        // We are providing our own zoom controls so disable the built-in ones.
+        MapUiSettings(zoomControlsEnabled = false)
     }
+
+    ZoomControls(
```
manuelvicnt, line 217: Should `ZoomControls` and `GoogleMap` be wrapped in a `Column`? That'd make it more reusable as right now, it works because higher up in the hierarchy, there's a `Column`

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
```diff
@@ -177,58 +173,69 @@ fun DetailsContent(
...
-}
-
-@Composable
-private fun MapViewContainer(
-    map: MapView,
+fun CityMapView(
     latitude: String,
-    longitude: String
+    longitude: String,
+    cameraState: CameraPositionState = rememberSaveable(saver = CameraPositionState.Saver) { CameraPositionState() },
```
manuelvicnt, line 179:
```suggestion
    cameraPositionState: CameraPositionState = rememberSaveable(saver = CameraPositionState.Saver) { CameraPositionState() },
```

Are there any other camera states? might be better to be explicit in the name to avoid confusion

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
```diff
@@ -177,58 +173,69 @@ fun DetailsContent(
...
-    // composable. In this way, when an update to the MapView happens, this composable won't
-    // recompose and the MapView won't need to be recreated.
-    val mapView = rememberMapViewWithLifecycle()
-    MapViewContainer(mapView, latitude, longitude)
-}
-
-@Composable
-private fun MapViewContainer(
-    map: MapView,
+fun CityMapView(
```
manuelvicnt, line 176: It's not clear what the responsibilities of `CityMapView` and `MapViewContainer` are. Maybe add comments in the function's KDoc or merge those composables together?

Crane/app/src/androidTest/java/androidx/compose/samples/crane/details/DetailsActivityTest.kt
```diff
@@ -59,61 +51,67 @@ class DetailsActivityTest {
...
-            )
-        )
-    }
+        mapState = CameraPositionState()
+
+        val countDownLatch = CountDownLatch(1)
+
+        composeTestRule.setContent {
+            CraneTheme {
+                Column {
```
manuelvicnt, line 69: This `Column` shouldn't be needed if added to the `MapViewContainer`, see my comment below

Crane/app/src/androidTest/java/androidx/compose/samples/crane/details/DetailsActivityTest.kt
```diff
@@ -59,61 +51,67 @@ class DetailsActivityTest {
...
 
-    @After
-    fun tearDown() {
-        scenario.close()
+        countDownLatch.await()
     }
 
     @Test
     fun mapView_cameraPositioned() {
-        composeTestRule.onNodeWithText(cityDetails.city.nameToDisplay).assertIsDisplayed()
```
manuelvicnt, line 83: Should we keep testing that the name and description of the map is on the screen?

bentrengrove: Added a new test file for it

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
bentrengrove: So it only recomposes this latitude or longitude changes but I agree it looks a bit fragile. 

Ideally we would just not expose the cameraPositionState and use the built in initialiser for position, the problem is then we can't test that the map is being positioned correctly. I really wanted this Composable to kind be as described "show a map at this position with a marker" but in order to achieve the "at this position" part and be testable it needs to initialise the camera somewhere.

LaunchedEffect doesn't feel right because then the setting of the position is delayed
DisposableEffect doesn't feel right because there is nothing to dispose
SideEffect runs too late so has similar results to LaunchedEffect
I have done a trick with remember here to make sure it only reinitialises when cityLocation changes but it feels like a hack.

The other thing that could be done is we just pass in an initialised camerastate in both tests and code but that feels like it won't be testing much.

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
```diff
@@ -177,58 +173,69 @@ fun DetailsContent(
...
-            googleMap.moveCamera(CameraUpdateFactory.newLatLng(cameraPosition))
-        }
+    val mapUiSettings = remember {
+        // We are providing our own zoom controls so disable the built-in ones.
+        MapUiSettings(zoomControlsEnabled = false)
     }
+
+    ZoomControls(
+        onZoomIn = { cameraPositionState.move(CameraUpdateFactory.zoomIn()) },
+        onZoomOut = { cameraPositionState.move(CameraUpdateFactory.zoomOut()) }
```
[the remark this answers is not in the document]
bentrengrove: Good point, thank you

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
bentrengrove: Added comments, I like them separated because they do have different responsibilities.

Crane/app/src/androidTest/java/androidx/compose/samples/crane/details/DetailsContentTests.kt
```diff
@@ -0,0 +1,53 @@
...
+import androidx.compose.ui.test.onNodeWithText
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+
+class DetailsContentTests {
+    private val city = MADRID
+    private val testExploreModel = ExploreModel(city, "description", "imageUrl")
+
+    @get:Rule(order = 1)
```
manuelvicnt, line 34: Nit: if Hilt is not used in the test, the rule doesn't need to contain an `order`

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
```diff
@@ -193,12 +189,14 @@ fun CityMapView(
...
     }
 
-    // Set our initial map position when the city location changes
-    LaunchedEffect(cityLocation) {
-        cameraPositionState.position = CameraPosition.fromLatLngZoom(
-            cityLocation,
-            InitialZoom
-        )
+    // Set our initial camera position whenever the city location changes.
+    remember(cityLocation) {
```
manuelvicnt, line 193: I don't think this is a good usage of `remember` TBH. Maybe ask eng? If you don't want it to be in a `LaunchedEffect` because setting the position is delayed one frame, it'd probably be best to have it as it was before: being called on every recomposition.

bentrengrove: Yeah I don't like it either, I think you are right. The single frame for `LaunchedEffect` doesn't really matter at the end of the day so I have switched it to that.

manuelvicnt (approved):

Crane/app/build.gradle
```diff
@@ -29,6 +29,11 @@ if (rootProject.file("local.properties").exists()) {
     properties.load(rootProject.file("local.properties").newDataInputStream())
 }
 
+// For testing on CI, a key can also be provided via an environment variable.
+// The local.properties key takes precedence
+var systemGoogleMapsKey = System.getenv("GOOGLE_MAPS_KEY") ?: ""
```
manuelvicnt, line 34:
```suggestion
String systemGoogleMapsKey = System.getenv("GOOGLE_MAPS_KEY") ?: ""
```

Is this the right way to define these variables? Looking at the lines above, the type can be indicated? Also, I've seen `def` in other places

Crane/app/src/main/java/androidx/compose/samples/crane/details/DetailsActivity.kt
```diff
@@ -176,59 +175,84 @@ fun DetailsContent(
...
+    val cityLocation = remember(latitude, longitude) {
         LatLng(latitude.toDouble(), longitude.toDouble())
     }
 
-    LaunchedEffect(map) {
-        val googleMap = map.awaitMap()
-        googleMap.addMarker { position(cameraPosition) }
-        googleMap.moveCamera(CameraUpdateFactory.newLatLng(cameraPosition))
+    // Set our initial camera position whenever the city location changes.
+    LaunchedEffect(cityLocation) {
```
manuelvicnt, line 194:
```suggestion
    LaunchedEffect(cityLocation, cameraPositionState) {
```

As `cameraPositionState` is also used in the effect block, we should "key" it as well

bentrengrove: That would break it, it's only meant to happen on first load. I think I am just going to fully remove it and pass in only a cameraPositionState. There is definitely something not right about this section

PR #749, opened 2022-03-04
Inital Reply M3 theme setup.

Gurupreet:
1:) Focus on user creating theme using theme builder or directly using from design
2:) Complete color schemes for light and dark with clear differentiation between color usage.
3:) Using ReplyM3 name for theme to highlight M3 of colorScheme and theming.


Reply/app/src/main/java/com/example/reply/ui/theme/Type.kt
```diff
@@ -6,13 +6,78 @@ import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.sp
 
-// Set of Material typography styles to start with
-val Typography = Typography(
+//Material 3 typography
```
ricknout, line 9: nit: Space after //

Reply/app/src/main/java/com/example/reply/ui/theme/Theme.kt
```diff
@@ -15,43 +15,90 @@ import androidx.compose.ui.platform.LocalContext
...
 
 @Composable
-fun ReplyTheme(
+fun ReplyM3Theme(
     darkTheme: Boolean = isSystemInDarkTheme(),
     dynamicColor: Boolean = true,
     content: @Composable () -> Unit
 ) {
-    val colorScheme = when {
+    val selectedColorScheme = when {
```
ricknout, line 83: I think we should change this to `replyColorScheme` to align with the `replyTypography` naming

Reply/app/src/main/java/com/example/reply/ui/theme/Theme.kt
```diff
@@ -15,43 +15,90 @@ import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalView
 import androidx.core.view.ViewCompat
 
-private val DarkColorScheme = darkColorScheme(
-    primary = Purple80,
-    secondary = PurpleGrey80,
-    tertiary = Pink80
+//Material 3 color schemes
```
ricknout, line 18: Hmm I actually think the color schemes should go in `Color.kt`

Reply/app/src/main/java/com/example/reply/MainActivity.kt
```diff
@@ -10,13 +10,13 @@ import androidx.compose.material3.Text
...
 import androidx.compose.ui.tooling.preview.Preview
-import com.example.reply.ui.theme.ReplyTheme
+import com.example.reply.ui.theme.ReplyM3Theme
 
 class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContent {
-            ReplyTheme {
+            ReplyM3Theme {
```
ricknout, line 19: I think we can keep this as `ReplyTheme` (otherwise the naming for other things should probably also change eg. `replyM3DarkColorScheme` which would be a bit messy IMO)

Reply/app/src/main/java/com/example/reply/ui/theme/Color.kt
```diff
@@ -2,10 +2,60 @@ package com.example.reply.ui.theme
...
-val PurpleGrey80 = Color(0xFFCCC2DC)
-val Pink80 = Color(0xFFEFB8C8)
+// Generate them via theme builder
+// https://material-foundation.github.io/material-theme-builder/#/custom
 
-val Purple40 = Color(0xFF6650a4)
-val PurpleGrey40 = Color(0xFF625b71)
-val Pink40 = Color(0xFF7D5260)
\ No newline at end of file
+val reply_light_primary= Color(0xFF825500)
```
ricknout, line 8: I filed a bug for the theme builder about this, but... We should change these from snake_case to camelCase

Reply/app/src/main/java/com/example/reply/ui/theme/Theme.kt
Gurupreet: I think it's fine to be in theme because default project also generate colorSchemes in Theme.kt and Color.kt just contains color codes for reference. Many of our samples follows this as well.

Reply/app/src/main/java/com/example/reply/MainActivity.kt
Gurupreet: Agreed 👍🏽

Reply/app/src/main/java/com/example/reply/ui/theme/Color.kt
Gurupreet: Yes I noticed I will raise it to material team as well 👍🏽 Please share link of buganizer if you have.

Reply/app/src/main/java/com/example/reply/ui/theme/Theme.kt
ricknout: Okay, sounds good then

ricknout (approved):

PR #750, opened 2022-03-07
Added dummy data layer and Flow setup for Reply M3. 

Gurupreet:
1:) Added dummy data for Accounts and Emails aligned with **[Reply Android view version](https://github.com/material-components/material-components-android-examples/tree/reply-material-3/Reply)**
2:) Expose flows with delay to simulate network response


Reply/app/src/main/java/com/example/reply/data/EmailsRepositoryImpl.kt
```diff
@@ -0,0 +1,29 @@
+package com.example.reply.data
+
+import com.example.reply.data.local.LocalEmailsDataProvider
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
+
+class EmailsRepositoryImpl : EmailsRepository {
```
ricknout, line 8: Add a comment to describe this

Reply/app/src/main/java/com/example/reply/data/AccountsRepository.kt
```diff
@@ -0,0 +1,9 @@
+package com.example.reply.data
+
+import kotlinx.coroutines.flow.Flow
+
+interface AccountsRepository {
```
ricknout, line 5: Add a comment describing this interface

ricknout (commented):
If there's a link to the Reply M3 Views implementation, please add it to the description

ricknout (approved):

PR #752, opened 2022-03-10
[All] Update to Compose 1.2.0-alpha05

JolandaVerhoef:


Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
```diff
@@ -88,54 +88,27 @@ fun Question(
     onAnswer: (Answer<*>) -> Unit,
     onAction: (Int, SurveyActionType) -> Unit,
     onDoNotAskForPermissions: () -> Unit,
-    openSettings: () -> Unit,
```
florina-muntenescu, line 91: Why was this removed? feels like something outside the scope of an alpha update?

florina-muntenescu: new permissions api. nevermind

florina-muntenescu (approved):

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
```diff
@@ -52,63 +55,61 @@ fun JetnewsApp(
...
+            },
+            drawerState = sizeAwareDrawerState,
+            // Only enable opening the drawer via gestures if the screen is not expanded
+            gesturesEnabled = !isExpandedScreen
+        ) {
+            Row(
+                Modifier
+                    .fillMaxSize()
+                    .statusBarsPadding()
+                    .navigationBarsPadding()
```
alexvanyo, line 98: The equivalent here should be `.windowInsetsPadding(WindowInsets.navigationBars.only(WindowInsetsSides.Horizontal + WindowInsetsSides.Top))`

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -238,7 +238,7 @@ private fun InterestScreenContent(
 private val tabContainerModifier = Modifier
     .fillMaxWidth()
     .wrapContentWidth(Alignment.CenterHorizontally)
-    .navigationBarsPadding(start = false, end = false)
+    .navigationBarsPadding()
```
alexvanyo, line 241: This isn't a direct replacement, but due to consumption, this actually will act the same: the start and end won't be applied again, since they are already being applied in `JetnewsApp`

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/ConversationFragment.kt
```diff
@@ -78,9 +64,8 @@ class ConversationFragment : Fragment() {
                         onNavIconPressed = {
                             activityViewModel.openDrawer()
                         },
-                        // Add padding so that we are inset from any left/right navigation bars
-                        // (usually shown when in landscape orientation)
-                        modifier = Modifier.navigationBarsPadding(bottom = false)
+                        // Add padding so that we are inset from any navigation bars
+                        modifier = Modifier.navigationBarsPadding()
```
alexvanyo, line 68: The direct replacement here would be `windowInsetsPadding(WindowInsets.navigationBars.only(WindowInsetsSides.Horizontal + WindowInsetsSides.Top))`

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -34,43 +35,39 @@ import com.example.jetsnack.ui.home.JetsnackBottomBar
...
+                    modifier = Modifier.systemBarsPadding(),
+                    snackbar = { snackbarData -> JetsnackSnackbar(snackbarData) }
+                )
+            },
+            scaffoldState = appState.scaffoldState
+        ) { innerPaddingModifier ->
+            NavHost(
+                navController = appState.navController,
+                startDestination = MainDestinations.HOME_ROUTE,
+                modifier = Modifier.padding(innerPaddingModifier)
```
alexvanyo, line 65: Just noticed while testing:

Jetsnack doesn't handle start and end navigation bar insets correctly, so content tends to be cut off. Don't want to necessarily hold up this conversion PR, but it'd be good to perhaps go through all insets usage across these samples to ensure there aren't other lingering issues.

alexvanyo (changes requested):
Overall looks great, thank you for doing this!

Most of comments are minor tweaks to keep existing behavior, although in going through these I noticed some inconsistent behavior (that was there before this PR too).

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
simona-anomis: Yeah, that's a good idea, we can do a sample wide audit, thanks!

Crane/app/src/main/java/androidx/compose/samples/crane/calendar/Calendar.kt
```diff
@@ -70,14 +72,10 @@ fun Calendar(
     modifier: Modifier = Modifier
 ) {
     LazyColumn(modifier) {
-        item { Spacer(Modifier.height(32.dp)) }
         for (month in calendarYear) {
             itemsCalendarMonth(month = month, onDayClicked = onDayClicked)
-            item {
-                Spacer(Modifier.height(32.dp))
```
alexvanyo, line 77: Nit: I think the new logic is slightly different than the original, where the last set of items no longer has a `32dp` space below it (in addition to the navigation bars height)

alexvanyo (approved):
Looks good to me! Just one more nit for a logic change I noticed.

PR #760, opened 2022-03-23
[All] Update to Compose 1.2.0-alpha06

IanGClifton:


IanGClifton:
Draft PR: Pending updates for the theme adapter and accompanist

Crane/app/src/main/java/androidx/compose/samples/crane/calendar/Calendar.kt
```diff
@@ -150,7 +150,6 @@ private fun Day(
         onClick = { onDayClicked(day) },
         onClickEnabled = enabled,
         backgroundColor = day.status.color(MaterialTheme.colors),
-        onClickLabel = stringResource(id = R.string.click_label_select)
```
bentrengrove, line 153: What replaces this?

IanGClifton: I mistakenly thought this was the part at the beginning of the Talkback announcement, but that's the state description. (which does still work correctly). By deleting this, the announcement now says "Double-click to activate" instead of "Double-click to select," but Surface doesn't directly expose onClickLabel anymore. I need to think this through.

IanGClifton: I updated the PR so that the onClickLabel is still passed all the way through but just not used in the Surface since there isn't a clean way to do it now. I created a bug (https://github.com/android/compose-samples/issues/766) and referenced it in the DayContainer composable so that no one accidentally deletes the property before we figure out how we want to handle it. Hopefully that makes this easy to update if alpha07 allows the onClickLabel to be passed in.

bentrengrove (approved):

PR #761, opened 2022-03-24
[Crane] Improved Tab selector ripple and added animation transition between tabs

riggaroo:
## Description ✏️

The Tab's on Crane were current don't animate the selection changes, and the ripple shape is currently square - extending past the size of the rounded border. 

This PR changes the ripple to be clipped to the `RoundedCornerShape` and animates the tab selection based on the docs [here](https://developer.android.com/reference/kotlin/androidx/compose/material/package-summary#TabRow(kotlin.Int,androidx.compose.ui.Modifier,androidx.compose.ui.graphics.Color,androidx.compose.ui.graphics.Color,kotlin.Function1,kotlin.Function0,kotlin.Function0)).

## Screenshots 📸

| Before | After |
|-------|-------|
|Crane_TabRow_Before| Crane_TabRow_Improvements|


IanGClifton (approved):
Beautiful!

PR #764, opened 2022-03-25
Update BottomBar height within Article Screen

### LINKED ISSUE #763: [Jetnews] Article BottomBar Buttons are too small to use
Within the Article Screen of JetNews, the bottom bar's buttons are too small to be clickable. 

As seen in the below screenshot (from a Pixel 3a XL device)

ed-george:
Fixes #763

| **Before** | **After** |
|------------|-----------|
|        |         |

### DIFF
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
@@ -213,9 +213,10 @@ private fun BottomBar(
         Row(
             verticalAlignment = Alignment.CenterVertically,
             modifier = Modifier
-                .height(102.dp)
-                .fillMaxWidth()
                 .windowInsetsPadding(WindowInsets.navigationBars.only(WindowInsetsSides.Vertical))
+                .height(56.dp)
+                .fillMaxWidth()
+                
         ) {
             FavoriteButton(onClick = onUnimplementedAction)
             BookmarkButton(isBookmarked = isFavorite, onClick = onToggleFavorite)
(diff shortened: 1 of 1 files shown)


JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
```diff
@@ -213,7 +213,7 @@ private fun BottomBar(
         Row(
             verticalAlignment = Alignment.CenterVertically,
             modifier = Modifier
-                .height(56.dp)
+                .height(102.dp)
```
alexvanyo, line 216: This adds an extra fixed height of `46dp`, but we want that height to based on the insets padding applied below.

The more flexible solution here is to swap the order of the `windowInsetsPadding` and the `height(56.dp)`, so that we set the height after the `windowInsetsPadding` is applied, which ensures that the insets padding doesn't take away the space available to the buttons.

alexvanyo (changes requested):
Thank you for the contribution! I proposed an alternate solution inline which should be more flexible with various inset sizes.

JetNews/app/src/main/java/com/example/jetnews/ui/article/ArticleScreen.kt
ed-george: Ah I see, yes that makes a lot more sense. Will update that now, thanks for the explanation

alexvanyo (approved):
Looks good, thanks for the screenshots in the description!

I ran CI, and it looks like just a minor formatting issue that you should be able to fix automatically fix if you run `./gradlew spotlessApply`

alexvanyo:
Thanks again for spotting this and contributing the fix!

PR #765, opened 2022-03-25
[Crane] Extract String resources

### LINKED ISSUE #612: [Crane] Use String resources
Many strings in the app are hard-coded. They should use the Resources instead.

riggaroo:
## Description ✏️

Fixes #612 - extract all strings to resources instead of them being hardcoded.

### DIFF
--- a/Crane/app/src/main/java/androidx/compose/samples/crane/home/SearchUserInput.kt
+++ b/Crane/app/src/main/java/androidx/compose/samples/crane/home/SearchUserInput.kt
@@ -36,6 +36,8 @@ import androidx.compose.samples.crane.home.PeopleUserInputAnimationState.Invalid
 import androidx.compose.samples.crane.home.PeopleUserInputAnimationState.Valid
 import androidx.compose.samples.crane.ui.CraneTheme
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 
 enum class PeopleUserInputAnimationState { Valid, Invalid }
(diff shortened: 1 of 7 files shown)


Crane/app/src/main/java/androidx/compose/samples/crane/home/SearchUserInput.kt
```diff
@@ -72,8 +74,10 @@ fun PeopleUserInput(
         val tint = tintPeopleUserInput(transitionState)
 
         val people = peopleState.people
+        val resources = LocalContext.current.resources
+        val peopleString = resources.getQuantityString(R.plurals.number_adults_selected, people, people, titleSuffix)
```
bentrengrove, line 78: Plural support was added in the latest version of Compose, good place to try it!

bentrengrove:
Looks like it needs a spotless run.
`./gradlew :app:spotlessApply`

Crane/app/src/main/java/androidx/compose/samples/crane/home/SearchUserInput.kt
riggaroo: Cool @bentrengrove, think it'll be a good place for me to also to try update the docs then? https://developer.android.com/jetpack/compose/resources#string-plurals

bentrengrove: Definitely!

riggaroo: Updated this PR with the change. [d5c2732](https://github.com/android/compose-samples/pull/765/commits/d5c2732bec52ef0c0a91969eed8b0aa80119ae8f)

bentrengrove (approved):
Nice!

PR #767, opened 2022-03-28
[Crane] Fix issue with "Choose Destination" text being editable

riggaroo:
## Description

Previously when clicking on the "Choose Destination" text in the C

rane app, it would let you start editing the hint.  This PR introduces a change that uses `decorationBox` to draw the hint as a separate `Text` field, similar to the way in which Material `TextField` works. 

## Screenshots

| Before | After |
|-------|-------|
| https://user-images.githubusercontent.com/9973046/160439435-405447cf-4b1d-4d91-aa4d-13c18b0efcc7.mp4 | https://user-images.githubusercontent.com/9973046/160439498-4b312a1b-c8cc-4991-8fac-681f839a1f54.mp4  |


manuelvicnt (approved):
Thanks for fixing this!

PR #769, opened 2022-03-30
[Crane] Set the onClick label for DayContainer fix #766

### LINKED ISSUE #766: [Crane] Figure out how to pass onClickLabel to Surface
The [1.2.0-alpha06 androidx.compose.material update](https://developer.android.com/jetpack/androidx/releases/compose-material#1.2.0-alpha06) removed onClickLabel from the Surface API. That means the Crane app's DayContainer no longer has a clean way to specify the onClickLabel ([see here](https://github.com/android/compose-samples/pull/760/files#diff-d74c88d419cea90d8812be0e7bae69c611fee235808168084ad6352f776e2980L210)).

We need to figure out the best way to still set the onClickLabel to ensure the "Double-click to $onClickLabel" announcements are accurate. The selectable version of Surface uses a selectable modifier, which doesn't have an onClickLabel. The clickable Surface sets a clickable modifier, which could support an onClickLabel, but Surface doesn't expose it. Using the non-interactive Surface and specifying the full clickable modifier ourselves might work but's duplicating quite a few lines and would have to be evaluated to see if there are any problems with the other modifiers that Surface adds.

IanGClifton:
This specifies the onClick label to use for the individual calendar
days. I also stopped passing the selected boolean to Surface. When it is
passed, that uses the selectable version of Surface which announces the
composable as a Tab and automatically states "selected" at the start of
the announcement when selected==true but does not announce anything at
the start when not selected.

Instead, it's now using the clickable version of Surface, so our
announcements are like:

"{Selected/not selected}, January 1st 2020, Button, Double-tap to
select"

This seems reasonable, since the other Roles are Checkbox, Switch,
RadioButton, and Image, all of which seem less correct than Button.

### DIFF
--- a/Crane/app/src/main/java/androidx/compose/samples/crane/calendar/Calendar.kt
+++ b/Crane/app/src/main/java/androidx/compose/samples/crane/calendar/Calendar.kt
@@ -56,6 +56,7 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.SemanticsPropertyKey
 import androidx.compose.ui.semantics.SemanticsPropertyReceiver
 import androidx.compose.ui.semantics.clearAndSetSemantics
+import androidx.compose.ui.semantics.onClick
 import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.semantics.stateDescription
 import androidx.compose.ui.semantics.text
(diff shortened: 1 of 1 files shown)


IanGClifton:
This fixes https://github.com/android/compose-samples/issues/766

bentrengrove (approved):

PR #773, opened 2022-04-05
[All] Add a script to update all sample Compose versions

bentrengrove:
This PR adds a script that will search all the sample projects for Compose versions and update them all at the same time to either a release like 1.2.0-alpha06 or a snapshot. In order to achieve this I had to add the ability for JetNews to use a snapshot version.

Used with:
`scripts/upgrade_samples.sh`

Screen Shot 2022-04-05 at 3 38 43 pm

### DIFF
--- a/scripts/upgrade_samples.sh
+++ b/scripts/upgrade_samples.sh
@@ -33,36 +33,36 @@ read snapshot_version;
 
 echo "Upgrading to $compose_version snapshot $snapshot_version"
 
-# Upgrade Dependencies.kt
-# for DEPENDENCIES_FILE in `find . -type f -iname "dependencies.kt"` ; do
-#     echo $DEPENDENCIES_FILE;
-#     COMPOSE_BLOCK=false;
-#     while IFS= read -r line; do
-#         if [[ $line == *"val version ="* ]] && $COMPOSE_BLOCK = true; then
-#             echo "$line" | sed -En 's/".*"/"'$compose_version'"/p'
-#         elif [[ $line == *"val snapshot ="* ]] && $COMPOSE_BLOCK = true; then
-#             echo "$line" | sed -En 's/".*"/"'$snapshot_version'"/p'
-#         else
-#             if [[ $line == *"object Compose {"* ]]; then
-#                 COMPOSE_BLOCK=true;
-#             elif [[ $line == *"}"* ]]; then
-#                 COMPOSE_BLOCK=false;
-#             fi
-#             echo "$line";
-#         fi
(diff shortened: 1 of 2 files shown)


riggaroo (approved):
This is great 🙌🏻  What do you think about adding a line to the README.md about how to run the script?

arriolac:
Very nice!

scripts/upgrade_samples.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+# Copyright (C) 2020 The Android Open Source Project
```
IanGClifton, line 3: 2022 :)

scripts/upgrade_samples.sh
```diff
@@ -0,0 +1,72 @@
...
+
+set -e
+
+echo "Version to change Compose to (e.g 1.2.0-alpha06): ";
+read compose_version;
+
+echo "Snapshot ID: (Blank for none)";
+read snapshot_version;
+
+echo "Upgrading to $compose_version snapshot $snapshot_version"
```
IanGClifton, line 34: Nit: Maybe just a quick if statement here to print the snapshot portion only if there is a snapshot specified (so we don't have output like "Upgrading to 1.2.3 snapshot " when snapshot is null/empty).

bentrengrove: Done

scripts/upgrade_samples.sh
```diff
@@ -0,0 +1,72 @@
...
+#
+# Changes all samples to a new Compose version
+#
+# Example: To run 
+#     ./scripts/upgrade_samples.sh
+#
+########################################################################
+
+set -e
+
```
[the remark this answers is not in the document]
bentrengrove: Done

scripts/upgrade_samples.sh
```diff
@@ -0,0 +1,72 @@
...
+            echo "$line" | sed -En "s/\'.*'/\'$compose_version\'/p"
+        elif [[ $line == *"ext.compose_snapshot_version ="* ]]; then
+            echo "$line" | sed -En "s/\'.*'/\'$snapshot_version\'/p"
+        else
+            echo "$line";
+        fi
+    done < $DEPENDENCIES_FILE > "${DEPENDENCIES_FILE}_new"
+    mv "${DEPENDENCIES_FILE}_new" $DEPENDENCIES_FILE
+done
+
```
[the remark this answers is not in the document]
bentrengrove: Done

IanGClifton (approved):
Awesome, great work, Ben!!

PR #775, opened 2022-04-07
[All] Update to Compose 1.2.0-alpha07

bentrengrove:
Fixes new lint error when you leave the `Scaffold` `contentPadding` parameter unused. 

I switched out some implementations of window insets to the recommended version for this.
- With LazyLists I used `contentPadding` so the content still scrolls under the edge of the screen
- With other layouts just `modifier.padding(contentPadding)` is enough

### DIFF
--- a/Crane/app/src/main/java/androidx/compose/samples/crane/calendar/Calendar.kt
+++ b/Crane/app/src/main/java/androidx/compose/samples/crane/calendar/Calendar.kt
@@ -17,16 +17,21 @@
 package androidx.compose.samples.crane.calendar
 
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.ExperimentalLayoutApi
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.WindowInsets
+import androidx.compose.foundation.layout.consumedWindowInsets
 import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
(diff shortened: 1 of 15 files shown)


JolandaVerhoef (approved):

Crane/app/src/main/java/androidx/compose/samples/crane/calendar/Calendar.kt
```diff
@@ -70,13 +68,16 @@ typealias CalendarWeek = List<CalendarDay>
 fun Calendar(
     calendarYear: CalendarYear,
     onDayClicked: (CalendarDay, CalendarMonth) -> Unit,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    contentPadding: PaddingValues = PaddingValues()
 ) {
-    LazyColumn(modifier) {
+    LazyColumn(
+        modifier = modifier,
```
alexvanyo, line 75: I'm not sure I understand the removal of the insets spacer here.

The `contentPadding` from the `Scaffold` could be from a top or bottom bar, and can exist independently from the insets.

However, it is important that insets aren't double applied. To that end, I think we should mark the contentPadding as a consumption of the insets, and then keep the `Spacer` below: 
```suggestion
        modifier = modifier.consumedWindowInsets(contentPadding),
```

alexvanyo (changes requested):

Crane/app/src/main/java/androidx/compose/samples/crane/calendar/Calendar.kt
bentrengrove: I did test it out and it looked like it was working. I will try out this solution as well.

bentrengrove: Fixed, could you review again please? I wasn't sure if I still needed the contentPadding below or not

alexvanyo (approved):
LGTM!

PR #778, opened 2022-04-13
[JetSnack] Move scroll state reads into layout phase via lambdas

### LINKED ISSUE #231: [Jetsnack] Vertical scrolling is janky
Tested on a Pixel 2XL running Android 11.

bentrengrove:
Fixes janky scrolling on the JetSnack details page.
Fixes #231 

It should not be necessary to recompose just to relayout a page. JetSnack has a collapsing toolbar like layout on the details page, it shrinks the image and moves the title text up on scroll. This was previously done using in composition but by moving the read of the scroll value into a lambda, we can read the value only during layout and skip the composition phase entirely.


riggaroo (approved):

juyeop03:
Your idea is so impressive for me.
Thank you 👍

manosbatsis:
Great example for explaining this within my  team. Thank you @bentrengrove !

PR #779, opened 2022-04-19
[Jetsurvey]: Prevent email text field from going off screen.

### LINKED ISSUE #757: [Jetsurvey] When the Email composable is focused, the UI is pushed off screen
Hi, 
First of all, Thank you for this wonderful code sample. 

**Issue**
The email editext in JetSurvey App goes offscreen while clicking in it. 
image

arriolac:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open a GitHub issue as a bug/feature request before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Appropriate docs were updated (if necessary)

Fixes #757 🦕

Screenshot_20220419_154252

### DIFF
--- a/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/WelcomeScreen.kt
+++ b/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/WelcomeScreen.kt
@@ -22,7 +22,6 @@ import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.wrapContentHeight
 import androidx.compose.foundation.rememberScrollState
@@ -46,7 +45,6 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.input.ImeAction
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.tooling.preview.Preview
-import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import com.example.compose.jetsurvey.R
 import com.example.compose.jetsurvey.theme.JetsurveyTheme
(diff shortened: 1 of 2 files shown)


alexvanyo:
Could we fix this more directly without changing the `windowSoftInputMode` by updating the `WelcomeScreen` setup for the animation?

It looks like we should be able to replace the existing `.onGloballyPositioned` with a simpler approach that removes the cyclic data dependencies:

```kotlin
@Composable
fun WelcomeScreen(onEvent: (WelcomeEvent) -> Unit) {
    var showBranding by remember { mutableStateOf(true) }

    Surface(modifier = Modifier.supportWideScreen()) {
        Column(
            modifier = Modifier
                .fillMaxWidth()
                .verticalScroll(rememberScrollState())
        ) {
            Spacer(
                modifier = Modifier
                    .weight(1f, fill = showBranding)
                    .animateContentSize()
            )

            AnimatedVisibility(
                showBranding,
                Modifier.fillMaxWidth()
            ) {
                Branding()
            }

            Spacer(
                modifier = Modifier
                    .weight(1f, fill = showBranding)
                    .animateContentSize()
            )

            SignInCreateAccount(
                onEvent = onEvent,
                onFocusChange = { focused -> showBranding = !focused },
                modifier = Modifier
                    .fillMaxWidth()
                    .padding(horizontal = 20.dp)
            )
        }
    }
}
```

arriolac:
@alexvanyo I considered doing a more direct approach like that but initially went against it since I'm not seeing any other usage of `onGloballyPositioned` or `onSizeChanged` in other samples. However, I do agree that the current code is a bit convoluted and we want to demo best practices here.

alexvanyo:
I think it'd be worth consulting others, but my impression from https://developer.android.com/jetpack/compose/phases#recomp-loop is that `onGloballyPositioned` is generally discouraged if it leads to the layout changing again, so I'm not sure if that's something we really want to highlight here.

arriolac:
Assigning this issue to @astamato to weigh in as well.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/WelcomeScreen.kt
```diff
@@ -60,38 +59,33 @@ sealed class WelcomeEvent {
...
-    currentOffsetHolder.value = if (showBranding) 0f else -brandingBottom
-    val currentOffsetHolderDp =
-        with(LocalDensity.current) { currentOffsetHolder.value.toDp() }
-    val heightDp = with(LocalDensity.current) { heightWithBranding.toDp() }
     Surface(modifier = Modifier.supportWideScreen()) {
-        val offset by animateDpAsState(targetValue = currentOffsetHolderDp)
         Column(
             modifier = Modifier
                 .fillMaxWidth()
-                .brandingPreferredHeight(showBranding, heightDp)
```
alexvanyo, line 77: Nit: the `brandingPreferredHeight` modifier should no longer be needed with this approach

alexvanyo (approved):
LGTM as-is, feel free to wait for other input though!

astamato:
Agree with this new approach :) nice one!

PR #801, opened 2022-05-12
Update Coil to 2.0.0.

colinrtwhite:
Updates Coil to `2.0.0`. `2.0.0` has a [number of changes](https://github.com/coil-kt/coil/blob/main/CHANGELOG.md#200---may-10-2022), but the main change for Jetpack Compose is the new `AsyncImage` composable.

I've updated the code to use `AsyncImage` where appropriate and clean up a few minor issues I noticed.


Crane/app/src/main/java/androidx/compose/samples/crane/base/ExploreSection.kt
```diff
@@ -206,6 +199,13 @@ private fun ExploreImage(item: ExploreModel) {
                     .align(Alignment.Center),
             )
         }
+
+        Image(
```
colinrtwhite, line 203: I've intentionally moved this after the `Image` for the loading state. This is so the "success" image is drawn on top of the placeholder if it's already in the memory cache since `AsyncImagePainter.state` is always `Loading` during the initial composition ([more info](https://coil-kt.github.io/coil/compose/#observing-asyncimagepainterstate)).

Owl/app/src/androidTest/java/com/example/owl/ui/fakes/TestImageLoader.kt
```diff
@@ -0,0 +1,79 @@
...
+import coil.request.Disposable
+import coil.request.ImageRequest
+import coil.request.ImageResult
+import coil.request.SuccessResult
+import kotlinx.coroutines.CompletableDeferred
+
+/**
+ * Replaces all remote images with a simple black drawable to make testing faster and hermetic.
+ */
+fun installTestImageLoader() {
```
colinrtwhite, line 38: Refactored this class slightly as it used `LocalImageLoader`, which is now deprecated.

astamato (approved):
LGTM, thank you for taking the time to update Coil :)

simona-anomis (approved):

PR #802, opened 2022-05-12
Remove CODEOWNERS file.

arriolac:
Removing CODEOWNERS file as the config under `.github/blunderbuss.yml` will auto-assign issues and PRs. Blunderbuss guarantees only one person will be assigned to an issue/PR rather than the whole team.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,0 @@
-* @android/compose-devrel


riggaroo:
I've changed the settings to not notify everyone and remove the team when someone is assigned - still would like to keep the codeowners file as its a bit better for assignment than blunderbuss (ie we can put our status as busy and it wont assign if on leave it). As well as the added security of not allowing pushes without a codeowner approval. What do you think?

arriolac:
Agreed that approach is much better. Closing this

Issue #811, opened 2022-05-17
[FR]: Require/force ktlint/spotless on PRs


IanGClifton:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is this a feature request for one of the samples?

- [X] Yes, this is a specific request related to this samples repo.

### Sample app

Crane

### Describe the problem

We can submit code that fails lint checks 😭

### Describe the solution

We could run spotless as part of the build checks and fail when there are issues.

We could also consider `./scripts/gradlew_recursive.sh spotlessApply` as a precommit hook or some other way to easily automate this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bentrengrove:
I narrowed this down a bit. CI is only running on one sample when a PR is submitted, so PRs that touch multiple samples are not linted correctly.

<img width="848" alt="Screen Shot 2022-05-18 at 9 34 24 am" src="https://user-images.githubusercontent.com/19445/168929189-46a33ca9-f0ac-454d-9caa-49e150dae553.png">

bentrengrove:
Narrowed it down further, they are being cancelled as a side effect of https://github.com/android/compose-samples/pull/796

arriolac:
Nice @bentrengrove. I think a fix would be to create sample-specific grouping so cancellations aren't applied globally.

riggaroo:
Oh snap, I'll get on this. Thanks folks for tagging.

riggaroo:
This should fix it - https://github.com/android/compose-samples/pull/814

PR #813, opened 2022-05-17
Applied spotless fixes

IanGClifton:


arriolac (approved):
LGTM. 

Related comment—our GitHub workflows are set up to run spotless on new pull requests. So in theory the triggered GitHub action that integrated this change should have failed. Was that not the case? For example, see: https://github.com/android/compose-samples/blob/main/.github/workflows/Owl.yaml#L50

bentrengrove:
I've just looked into this as well. It looks like it picks just one sample to run, so PRs that touch multiple samples don't get linted properly. It must be the first one that meets the wildcard in the yaml or something?

bentrengrove:
#811 works out the issue

riggaroo (approved):

PR #821, opened 2022-05-20
[Jetcaster] Fix Readme out-of-date

jisungbin:
Since WindowInsets and Coil are no longer supported by Accompanist, they have been removed from app dependencies, but still remain in Readme.

```kt
object Accompanist {
    const val version = "0.24.9-beta"
    const val pager = "com.google.accompanist:accompanist-pager:$version"
}
```

There is only Pager as Accompanist in the app dependencies.


bentrengrove (approved):
Thank you!

PR #834, opened 2022-06-03
Adding secrets-gradle-plugin.

arriolac:
Adding [secrets-gradle-plugin](https://github.com/google/secrets-gradle-plugin) to handle reading key/values from `local.properties`.


arriolac:
Replaced by #835

Issue #746, opened 2022-02-28
[Crane] Google Maps tests are failing to initialise


bentrengrove:
Possibly due to API key restrictions

PR #836, opened 2022-06-03
Adding secrets-gradle-plugin.

arriolac:
Adding [secrets-gradle-plugin](https://github.com/google/secrets-gradle-plugin) to handle reading key/values from `local.properties`.

### DIFF
--- a/.github/workflows/Crane.yaml
+++ b/.github/workflows/Crane.yaml
@@ -53,10 +53,16 @@ jobs:
             ~/.gradle/caches/build-cache-*
           key: gradle-${{ hashFiles('checksum.txt') }}
 
+    - name: Inject Maps API Key
+      run: |
+        # Injecting the key directly into the manifest as secrets-gradle-plugin
+        # isn't picking up the key when creating a local.properties file here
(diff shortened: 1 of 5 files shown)


.github/workflows/Crane.yaml
```diff
@@ -53,10 +53,16 @@ jobs:
             ~/.gradle/caches/build-cache-*
           key: gradle-${{ hashFiles('checksum.txt') }}
 
+    - name: Inject Maps API Key
+      run: |
+        # Injecting the key directly into the manifest as secrets-gradle-plugin
+        # isn't picking up the key when creating a local.properties file here
+        sed -i -e "s,\${googleMapsKey},$GOOGLE_MAPS_KEY,g" ${{ env.SAMPLE_PATH }}/app/src/main/AndroidManifest.xml
+      env:
+        GOOGLE_MAPS_KEY: ${{ secrets.GOOGLE_MAPS_KEY }}
```
riggaroo, line 62: This is great - Would it also fix #746 ? CityMapViewTests

arriolac: This uses the same API key so if the issue is due to key restrictions this won't fix it. I can take a look at #746  though to see what's going on

riggaroo (approved):

PR #837, opened 2022-06-06
[Jetchat] Add Nested scrolling interop API

simona-anomis:
Add the Nested scrolling interop API to Jetchat Profile screen, to showcase the proper usage of this experimental API in the scenario of having a View parent and a composable child. Steps done:

- Refactor ProfileFragment to use an old school XML layout, with a CoordinatorLayout, CollapsingToolbarLayout that hosts a ComposeView for the `JetchatAppBar` composable and a separate ComposeView that will host the `ProfileScreen` composable
- Set the correct Fragment layout and composables using the guidelines on View/Compose interoperability
- Hook up the nested scrolling interop connection between the View parent and composable childeren via `rememberNestedScrollInteropConnection()`  to enable the collapsing behavior and scrolling connection between all these elements
- Minor clean up

More elaborate documentation on this coming soon on http://developer.android.com/

### DIFF
--- a/Jetchat/app/src/main/java/com/example/compose/jetchat/profile/ProfileFragment.kt
+++ b/Jetchat/app/src/main/java/com/example/compose/jetchat/profile/ProfileFragment.kt
@@ -21,13 +21,31 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.MoreVert
+import androidx.compose.material3.Icon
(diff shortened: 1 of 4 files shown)


Jetchat/app/src/main/java/com/example/compose/jetchat/profile/ProfileFragment.kt
```diff
@@ -42,31 +60,61 @@ class ProfileFragment : Fragment() {
...
                         }
                     )
                 }
             }
         }
+
+        rootView.findViewById<ComposeView>(R.id.profile_compose_view).apply {
+            setContent {
+                val userData by viewModel.userData.observeAsState()
+                val nestedScrollInteropConnection = rememberNestedScrollInteropConnection()
```
astamato, line 104: you can use the default no in ProfileScreen?

Levi-Moreira (approved):

Issue #720, opened 2021-12-20
[Jetchat] Rename Scaffold 


JoseAlcerreca:
https://github.com/android/compose-samples/pull/716#pullrequestreview-836364155

JetchatScaffold had scaffoldState: ScaffoldState = rememberScaffoldState() and now it holds the drawerState: DrawerState = rememberDrawerState(), so should be renamed to JetchatDrawer or a more appropriate naming.

PR #838, opened 2022-06-08
[Jetchat] Rename JetchatScaffold

Min-Jae-Bae:
Renamed `JetchatDrawer` to `JetchatContainer` in JetchatScaffold. 
So I renamed the JetchatDrawer file to JetchatContainer file and In the JetchatConatiner file, all the `Drawer` to the `Container`

Renamed :
main_drawer.xml -> main_container.xml


bentrengrove:
Thanks for the PR. Why do you want to rename these functions though?

Min-Jae-Bae:
Thanks for the answer. Header, ItemHeader and Preview function calling in the JetchatConatiner should use the same name as `Conatiner`

bentrengrove:
Sorry, why do the first change though? `Renamed JetchatDrawer to JetchatContainer in JetchatScaffold.`
Why are you renaming it?

Min-Jae-Bae:
https://github.com/android/compose-samples/issues/720#issue-1084769340

so should be renamed a more appropriate naming.

bentrengrove:
Ah, next time please mention the issue in the PR description so we have context. In that case though it was `JetchatScaffold` they wanted renamed, not `JetchatDrawer` so I think you might have the wrong part?

I will let @astamato decide if she wants this change as this is her project

astamato:
Hi! Thanks for the PR! Precisely as @bentrengrove pointed, the issue linked in https://github.com/android/compose-samples/issues/720#issue-1084769340 is about renaming JetchatScaffold to JetchatDrawer containing the Drawer (it no longer contains the scaffold). Don't think the other functions need renaming as the header hasn't changed.

astamato:
Thank you! Tho frame sounds generic, and right now this composable represents the drawer. You think we could do one more adjustment and rename to JetchatDrawer? Cheers!

Min-Jae-Bae:
So renaming JetchatScaffold to JetchatFrameDrawer containing the Drawer @astamato

PR #839, opened 2022-06-09
Reply Detail screen navigation.

Gurupreet:
Screen Shot 2022-05-10 at 11 37 40


Reply/app/src/main/res/values/strings.xml
```diff
@@ -17,4 +17,8 @@
 
     <string name="empty_screen_title">Screen under construction</string>
     <string name="empty_screen_subtitle">This screen is still under construction. This sample will help you learn about adaptive layouts in Jetpack Compose</string>
+
+    <string name="back_button">Back button</string>
```
ricknout, line 21: nit: Maybe this should just be "back"?

ricknout (approved):
A couple small comments but LGTM, although I would wait for @simona-anomis to approve

alexvanyo (changes requested):
I'm getting an immediate crash upon opening the app with the latest branch:

```
java.lang.IllegalArgumentException: navigation destination Inbox is not a direct child of this NavGraph
```

Gurupreet:
> I'm getting an immediate crash upon opening the app with the latest branch:
> 
> ```
> java.lang.IllegalArgumentException: navigation destination Inbox is not a direct child of this NavGraph
> ```

Yes, that's what I was discussing with Simona, it's fixed but I lost access to push changes. I will push the fix 👍

Reply/app/src/main/java/com/example/reply/ui/ReplyApp.kt
```diff
@@ -227,4 +235,75 @@ fun ReplyNavGraph(
...
+//        inboxGraph(
+//            nestedGraphs = {
+//                emailDetailGraph {
+//                    ReplyEmailDetail(email = replyHomeUIState.selectedEmail ?: replyHomeUIState.emails.first()) {
+//                        navController.popBackStack()
+//                    }
+//                }
+//            }
+//        ) {
+//            ReplyInboxScreen(contentType, replyHomeUIState, navigationType, navigateToDetail)
```
Gurupreet, line 263: @simona-anomis this UI doesn't react to replyHomeUIState if put this way, below code react to homeUIState chages.

Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationActions.kt
```diff
@@ -69,4 +76,7 @@ class ReplyNavigationActions(navController: NavHostController) {
             restoreState = true
         }
     }
+    val navigateToDetail: (emailId: Long) -> Unit = { emailId ->
+        navController.navigate("${ReplyRouts.INBOX_ROUTE}/$emailId") {}
```
alexvanyo, line 80: Because we're explicitly navigating to the detail screen while on a small screen, this ends up acting a bit strange to the user.

If I'm on a small screen (say the outside screen of a foldable) and I'm on the inbox list, let's say I click on a conversation. That will navigate me to that specific conversation.

If I then unfold my device, I will still be seeing the conversation take up the entire screen, even though I would have room for a list/detail view.

If I then navigate back, I will then see the list and detail together (and probably the detail I just closed)

This will also be the case when on a tablet and entering/leaving multi-window mode, or resizing on Chrome OS.

Ideally, we wouldn't be navigating at all to the detail (at least, not calling `navController.navigate`) and then handling "navigation" (as perceived by the user) at the route capable of displaying the list, detail, or both.

alexvanyo:
Screenshot_20220613_160744

This isn't directly related to this PR, but wanted to highlight how much space the expanded navigation drawer takes up on the smaller end of `EXPANDED`

Gurupreet:
> Screenshot_20220613_160744
> 
> This isn't directly related to this PR, but wanted to highlight how much space the expanded navigation drawer takes up on the smaller end of `EXPANDED`

Yes I think there are two issues to be fixed here
1:) Permanent navigation drawer should be able to have adjusted width(or at least to maximum element) right now it's constant 360.dp. I have raised a bug on design and Material side for that.

2:) I think EXPANDED and MEDIUM can have one more state in windowSizeClass if anyone wants more granular control on the UI.

Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationActions.kt
Gurupreet: Yes, I am curious on this case too, I am not sure what should be guidance here. May be @simona-anomis can share insights too. If user is deeply navigated to nested screen, and unfold the device->
1:) We pop navigation and re-render detail-list set
2:) We Let detail screen expand
3:) Don't use navigate for Mobile if we suspect user can expand/unfold screen for this purpose/

alexvanyo: The conclusion we came to here for publishing https://developer.android.com/guide/topics/large-screens/navigation-for-responsive-uis#responsive_content_destinations was always to do 3, and assume the user can always resize the screen.

Navigating as a side-effect of an window size changing makes it a lot more difficult to save user state, and also can be frustrating when reversing a size change. (user rotates the device accidentally, rotates back, and their state or location in the app is gone).

What that roughly would look like here is the email route can display the list, detail, or both, and then it handles "navigation" internally, with a `BackHandler` when just the detail is showing, that updates the state driving which one to show if there isn't enough space available.

Gurupreet: I think I agree with that too, I will then keep it as top level navigation and handle using BackHandler assuming user can resize screen anytime.

Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationActions.kt
```diff
@@ -69,4 +71,7 @@ class ReplyNavigationActions(navController: NavHostController) {
             restoreState = true
         }
     }
+    val navigateToDetail: (emailId: Long) -> Unit = { emailId ->
+     //   navController.navigate("${ReplyRouts.INBOX_ROUTE}/$emailId") {}
```
alexvanyo, line 75: I think this `navigateToDetail` needs to be updated. Right now, clicking on an item when just the list is showing doesn't cause anything to happen.

Gurupreet: ok this was redundant now, removed it.

Reply/app/src/main/java/com/example/reply/ui/MainActivity.kt
```diff
@@ -78,7 +78,16 @@ class MainActivity : ComponentActivity() {
...
+                ReplyApp(
+                    windowSize.widthSizeClass,
+                    devicePosture,
+                    uiState,
+                    closeDetailScreen = {
+                        viewModel.closeDetailScreen()
+                    }
+                ) { emailId ->
+                    viewModel.setSelectedEmail(emailId)
+                }
```
alexvanyo, line 90: Very tiny nit, the trailing lambda here looks a little weird since it's another callback instead of displaying content. Could consider moving it into a named parameter like `closeDetailScreen`?

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
```diff
@@ -114,31 +139,108 @@ fun ReplyListOnlyContent(
...
-    modifier: Modifier = Modifier.padding(horizontal = 16.dp, vertical = 4.dp)) {
-    Card(modifier = modifier) {
+    isSelected: Boolean = false,
+    modifier: Modifier = Modifier.padding(horizontal = 16.dp, vertical = 4.dp),
+    navigateToDetail: (Long) -> Unit
+) {
+    Card(
+        modifier = modifier.clickable { navigateToDetail.invoke(email.id) },
+        colors = CardDefaults.cardColors(
+            containerColor = if (isSelected) MaterialTheme.colorScheme.primaryContainer else MaterialTheme.colorScheme.surfaceVariant
```
alexvanyo, line 241: It'd be great if this selected also hooked into the semantics selection for accessibility.

One caveat there, is we would only want that to be displayed when both the list and detail are showing, since  "selected" would be a bit weird if only the list is showing.

Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationActions.kt
```diff
@@ -24,6 +24,8 @@ object ReplyDestinations {
     const val ARTICLES = "Articles"
     const val DM = "DirectMessages"
     const val GROUPS = "Groups"
+    const val DETAIL = "Detail"
+    const val EMAIL_ID_KEY = "emailId"
```
alexvanyo, line 28: Nit: I think these are both unused now?

Reply/app/src/main/res/values/strings.xml
alexvanyo: +1 to this

alexvanyo (commented):
Logic looks good, just a few minor nits and cleanup.

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
Gurupreet: Yes it's only for detail only actually, you won't see for the single list.

Gurupreet: Added semantics for selected 👍🏽

alexvanyo: Could we only add the semantics if we're showing both the list and detail? Right now, the semantics is added for both, so on the list-only screen, we're getting the `Not selected`/`selected` read out.

alexvanyo (commented):
Looks good, just with the one last semantics question.

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
Gurupreet: done 👍🏽

alexvanyo (approved):
Looks good to me!

Reply/app/src/main/java/com/example/reply/ui/ReplyApp.kt
```diff
@@ -88,15 +94,17 @@ fun ReplyApp(
         }
     }
 
-    ReplyNavigationWrapperUI(navigationType, contentType, replyHomeUIState)
+    ReplyNavigationWrapperUI(navigationType, contentType, replyHomeUIState, closeDetailScreen, setSelectedEmail)
```
simona-anomis, line 97: Nit: I'd use named params here to make it obvious what is what

Reply/app/src/main/java/com/example/reply/ui/ReplyApp.kt
```diff
@@ -119,7 +128,7 @@ private fun ReplyNavigationWrapperUI(
                 navigationActions.navigateToGroups,
             )
         }) {
-            ReplyNavGraph(
+            ReplyAppContent(
```
simona-anomis, line 131: Same comment about named params across the PR

Reply/app/src/main/java/com/example/reply/ui/ReplyApp.kt
```diff
@@ -227,4 +238,35 @@ fun ReplyNavGraph(
             }
         }
     }
-}
\ No newline at end of file
+}
+
+@Composable
+private fun ReplyNavHost(
+    modifier: Modifier = Modifier,
```
simona-anomis, line 245: Nit: modifiers tend to be the first optional param, so in this case you'd push it down to the bottom

Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationActions.kt
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
```
simona-anomis, line 14: Comment for this entire class - why treat the navigation actions as vals instead of functions? Especially when naming them as actions "navigateTo..."

Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationComponents.kt
```diff
@@ -1,4 +1,20 @@
-package com.example.reply.ui
+/*
```
simona-anomis, line 1: Comment for the entire class: consider doing what we chatted about and moving all navigation information into the Destinations class (one object per each Destination with its icons, labels etc.) and creating a list of TopLevelDestinations, then iterating & calling .forEach { NavigationBarItem/ NavigationRailItem}. It would make this entire file 4x shorter and avoid a lot of repetition :). Also you'd have one class for centralizing all navigation related information in one place

simona-anomis (commented):
Looks good, left some comments about navigation, feel free to reject if you feel like it's out of scope!

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
```diff
@@ -232,12 +232,14 @@ fun EmailDetailAppBar(
 @Composable
 fun ReplyEmailListItem(
     email: Email,
+    isListAndDetailLayout: Boolean = false,
```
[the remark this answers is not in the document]
Gurupreet: Thanks for suggestion, Done

Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationComponents.kt
Gurupreet: Thanks, updated it to use TopLevelDestinations.

Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationActions.kt
Gurupreet: Thanks, Adopted this approach as well.

PR #842, opened 2022-06-12
[JetSurvey] Use AnimatedContent to slide in and out changes between questions

riggaroo:
Added two usages of AnimatedContent into JetSurvey.

1. Switching between Question Content

| Before  |  After |   
|---|---|
|Survey_Questions_No_animation_Demo_Mode | End_Questions_optimized |

3. Switching from questions to results

| Before  |  After |   
|---|---|
| End_Congrats_Before_DemoMode | End_Congrats_Final_Demo_Mode |

Animate the changes between different question screens, as well as the end result screen.

### DIFF
--- a/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyFragment.kt
+++ b/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyFragment.kt
@@ -22,11 +22,11 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.activity.result.contract.ActivityResultContracts.TakePicture
 import androidx.compose.animation.AnimatedContent
+import androidx.compose.animation.AnimatedContentScope
 import androidx.compose.animation.ExperimentalAnimationApi
 import androidx.compose.animation.core.tween
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
-import androidx.compose.animation.slideInVertically
 import androidx.compose.animation.with
 import androidx.compose.runtime.livedata.observeAsState
 import androidx.compose.ui.platform.ComposeView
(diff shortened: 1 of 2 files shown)


Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyFragment.kt
```diff
@@ -56,20 +64,34 @@ class SurveyFragment : Fragment() {
...
+                    val state = viewModel.uiState.observeAsState().value ?: return@JetsurveyTheme
+                    AnimatedContent(
+                        targetState = state,
+                        transitionSpec = {
+                            fadeIn() + slideIntoContainer(
+                                towards = AnimatedContentScope
+                                    .SlideDirection.Up,
+                                animationSpec = tween(600)
+                            ) with
+                                fadeOut(animationSpec = tween(200))
```
IanGClifton, line 76: Minor nit that the 600 on line 74 and 200 on line 76 feel like magic numbers. It'd be cool to have them defined in the theme and grab them as composition locals, but maybe it's enough to just make them constants since they're only used here. I'm also wondering if these two numbers have a relationship, so it can be something like fastAnimationDuration = 200, slowAnimationDuration = fastAnimationDuration * 3? Or if they should be independent.

IanGClifton (commented):
This is great! Thank you! I had just one super minor comment

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyFragment.kt
```diff
@@ -56,20 +64,34 @@ class SurveyFragment : Fragment() {
             )
             setContent {
                 JetsurveyTheme {
-                    viewModel.uiState.observeAsState().value?.let { surveyState ->
-                        when (surveyState) {
+                    val state = viewModel.uiState.observeAsState().value ?: return@JetsurveyTheme
```
simona-anomis, line 67: Why the `?: return@JetsurveyTheme` change?

simona-anomis (approved):
LGTM!

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyFragment.kt
riggaroo: I needed a non-nullable state, and I generally prefer early returns instead of wrapping content in lets 😄  https://betterprogramming.pub/are-early-returns-any-good-eed4b4d03866

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyFragment.kt
riggaroo: They should be independent - i've added two constants.

Issue #744, opened 2022-02-25
Drawer not opening after double-clicking on the drawer icon (Project JetChat)


ruben8888:
When we do a quick clicking on the drawer icon, the drawer won't open anymore.
I think the problem is in this part of the code
 val drawerOpen by viewModel.drawerShouldBeOpened.collectAsState()
                    if (drawerOpen) {
                        // Open drawer and reset state in VM.
                        LaunchedEffect(Unit) {
                            drawerState.open()
                            viewModel.resetOpenDrawerAction()
                        }
                    }

PR #844, opened 2022-06-13
[Jetchat] Drawer not opening after double-clicking on the drawer icon #744

alajemba-vik:
**Issue:** The drawer refuses to open.

**Repro:**

1. Click the drawer icon twice, or

2. Click the drawer icon, then quickly click anywhere other than the opening drawer before it completely opens.

Performing either of these two actions closes the drawer and leads to the issue stated above. 

**Reason:** The drawer's opening movement is interrupted before it can complete, and because of this, `MainViewModel.resetOpenDrawerAction()` is never called. 

**Intended Fix:** The update in this PR wraps the ` DrawerState.open()` call in a try-finally block.


astamato (approved):
LGTM

PR #851, opened 2022-06-17
[JetSurvey] Improve sign up process

tsuharesu:
* Handle configuration changes for text fields
* Don't submit until data is valid
* Add keyboard type to fields

This also includes changes similar to https://github.com/android/compose-samples/pull/529, so that PR can be closed when this is merged.

cc @alexvanyo


astamato:
Consider not storing unencrypted passwords as a best practice.
the rest LGTM, provided Ian agrees. Also ideally I'd try to use TextFieldValue but that requires a bigger refactor it seems.

tsuharesu:
Thanks @astamato. I changed the password bit.

astamato (approved):

PR #852, opened 2022-06-17
Add support for workflow_dispatch on all GH workflows.

arriolac:
Adding `workflow_dispatch` trigger to all workflows so that they can be manually run from the Actions tab. A number of workflows already supported this (like Crane). This would allow us to build samples from any branch without having to open a pull request.

Screen Shot 2022-06-17 at 1 00 23 PM


IanGClifton (approved):
LGTM

PR #857, opened 2022-06-27
Fixing disabled stickers tab

### LINKED ISSUE #855: [JetChat] Unreachable code In EmojiSelector Composable In UserInput.kt
For code location :- 

https://github.com/android/compose-samples/blob/e6f9d376d435ae8c94f9caff0eaa7a0befccb3a0/Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/UserInput.kt#L437

```
            ExtendedSelectorInnerButton(
                text = stringResource(id = R.string.stickers_label),
                onClick = { selected = EmojiStickerSelector.STICKER },
                selected = false,
                modifier = Modifier.weight(1f)
            )
        }
        Row(modifier = Modifier.verticalScroll(rememberScrollState())) {
            EmojiTable(onTextAdded, modifier = Modifier.padding(8.dp))
        }
    }
    if (selected == EmojiStickerSelector.STICKER) {
        NotAvailablePopup(onDismissed = { selected = EmojiStickerSelector.EMOJI })
    }
}
```
**Description** :- 
In the above code snippet, we can see that ExtendedSelectorInnerButton(Sticker) onClick listener will not get called because Implementation of ExtendedSelectorInnerButton will have enabled property false (we provided constant false value) so for this reason, last `If case` will never be true and   ` NotAvailablePopup(onDismissed = { selected = EmojiStickerSelector.EMOJI })` will be a unreachable code.

astamato:
Fixes #855

Before:

After:

### DIFF
--- a/Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/UserInput.kt
+++ b/Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/UserInput.kt
@@ -483,9 +483,11 @@ fun ExtendedSelectorInnerButton(
     modifier: Modifier = Modifier
 ) {
     val colors = ButtonDefaults.buttonColors(
-        containerColor = MaterialTheme.colorScheme.onSurface.copy(alpha = 0.08f),
+        containerColor = if (selected) MaterialTheme.colorScheme.onSurface.copy(alpha = 0.08f)
+        else Color.Transparent,
         disabledContainerColor = Color.Transparent,
-        contentColor = MaterialTheme.colorScheme.onSurface,
+        contentColor = if (selected) MaterialTheme.colorScheme.onSurface
+        else MaterialTheme.colorScheme.onSurface,
         disabledContentColor = MaterialTheme.colorScheme.onSurface.copy(alpha = 0.74f)
     )
     TextButton(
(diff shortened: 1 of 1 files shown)


IanGClifton (changes requested):
I think when this was originally written, it was meant to have a background around the selected button only (more like how tabs work) but that was implemented by setting the enabled property, which is why clicking doesn't actually work. I don't think there's any selectable button that we can use, but we should be able to do something like changed ExtendedSelectorInnerButton to directly use a Surface:

```
    val colorSelected = MaterialTheme.colorScheme.onSurface.copy(alpha = 0.08f)
    val colorUnselected = Color.Transparent
    val contentColorSelected = MaterialTheme.colorScheme.onSurface
    val contentColorUnselected = MaterialTheme.colorScheme.onSurface.copy(alpha = 0.74f)

    Surface(
        onClick = onClick,
        modifier = modifier.padding(8.dp).height(36.dp),
        shape = RoundedCornerShape(percent = 50),
        color = if (selected) { colorSelected } else { colorUnselected },
        contentColor = if (selected) { contentColorSelected} else { contentColorUnselected},
        selected = selected,
    ) {
        Row(
            horizontalArrangement = Arrangement.Center,
            verticalAlignment = Alignment.CenterVertically,
        ) {
            Text(
                text = text,
                style = MaterialTheme.typography.titleSmall
            )
        }
    }
```

I might be missing some of the styling, but hopefully that's close enough to show what I mean :)

IanGClifton (approved):
Nice, yeah this is a good simple fix :)

PR #860, opened 2022-06-29
[All] Update Compose Compiler to 1.2.0

arriolac:
Updating samples to use latest Compose Compiler 1.2.0 and Kotlin 1.7.0. This is the start of [independent versioning](https://android-developers.googleblog.com/2022/06/independent-versioning-of-Jetpack-Compose-libraries.html) of the compiler and various Compose libraries.

Also updated Room on Jetchat to 2.5.0-alpha02 which supports Kotlin 1.7.0.

### DIFF
--- a/Jetcaster/app/build.gradle
+++ b/Jetcaster/app/build.gradle
@@ -88,7 +88,7 @@ android {
     }
 
     composeOptions {
-        kotlinCompilerExtensionVersion Libs.AndroidX.Compose.version
+        kotlinCompilerExtensionVersion Libs.AndroidX.Compose.compilerVersion
     }
 }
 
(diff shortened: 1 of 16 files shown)


Jetcaster/app/build.gradle
```diff
@@ -88,7 +88,7 @@ android {
     }
 
     composeOptions {
-        kotlinCompilerExtensionVersion Libs.AndroidX.Compose.version
+        kotlinCompilerExtensionVersion = "1.2.0"
```
florina-muntenescu, line 91: should we create a compose compiler version in Libs.AndroidX.Compose so we don't hardcode the version of the compiler? 

Here and in the other projects as well

florina-muntenescu (changes requested):

Jetcaster/app/build.gradle
arriolac: Addressed in d5ff569a

florina-muntenescu (approved):

PR #881, opened 2022-07-10
[JetNews] Use derived state of for scrolled state to avoid recomposition on scrolling

### LINKED ISSUE #841: [JetNews] HomeTopAppBar continuously recompose during scrolling list
https://github.com/android/compose-samples/blob/5f70d4801b9450efa7858c110890c93b0a8e804b/JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt#L294

because using LazyListState.firstVisibleItemScrollOffset directly, so each time this value changes, the recomposition is triggered

use
`val isScrolled by remember {
    derivedStateOf { homeListLazyListState.isScrolled }
}`
then 
`elevation = if (!isScrolled) 0.dp else 4.dp`
maybe better

oas004:
This should fix  #841  😃

### DIFF
--- a/JetNews/app/src/main/java/com/example/jetnews/utils/LazyListUtils.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/utils/LazyListUtils.kt
@@ -17,6 +17,7 @@
 package com.example.jetnews.utils
 
 import androidx.compose.foundation.lazy.LazyListState
+import androidx.compose.runtime.derivedStateOf
 
 val LazyListState.isScrolled: Boolean
-    get() = firstVisibleItemIndex > 0 || firstVisibleItemScrollOffset > 0
+    get() = derivedStateOf { firstVisibleItemIndex > 0 || firstVisibleItemScrollOffset > 0 }.value


bentrengrove (changes requested):
Thank you for this! This is a great idea.

This fixes it for this screen but `isScrolled` is used in multiple places in the app. A better fix would be moving the derivedStateOf to `LazyListUtils.kt`.

e.g

```
val LazyListState.isScrolled: Boolean
    get() = derivedStateOf { firstVisibleItemIndex > 0 || firstVisibleItemScrollOffset > 0 }.value
```

Could you make that change please?

oas004:
> Thank you for this! This is a great idea.
> 
> This fixes it for this screen but `isScrolled` is used in multiple places in the app. A better fix would be moving the derivedStateOf to `LazyListUtils.kt`.
> 
> e.g
> 
> ```
> val LazyListState.isScrolled: Boolean
>     get() = derivedStateOf { firstVisibleItemIndex > 0 || firstVisibleItemScrollOffset > 0 }.value
> ```
> 
> Could you make that change please?

Thanks for review! Very good point! Changed it to use `derivedStateOf` in the `LazyListUtils` instead and rebased the commit 😄

bentrengrove (approved):
LGTM! Thanks for the contribution!

PR #882, opened 2022-07-11
Reply adaptive sample ready for integration to compose samples.

Gurupreet:
This sample is a Jetpack Compose implementation of **Reply**, a[ Material Design study for adaptive design guidance](https://m3.material.io/foundations/adaptive-design/large-screens/overview).

All the CODE PR's are pre merged to [Reply branch](https://github.com/android/compose-samples/tree/Gurupreet/reply-m3) 

README file and version matching with samples is added for this PR.

**Reply highlights**
- Adaptive UI for mobile tablet and desktops
- Foldable support
- Material 3 theming & components
- Dynamic colors and Light/Dark theme support

**Versions**

- Compose: 1.2.0-rc03
- Kotlin: 1.7.0
- Material3: 1.0.0-alpha14

**Related Content**
Codelab related to Reply-> [Compose Adaptive codelab](https://codelabs.developers.google.com/jetpack-compose-adaptability?hl=en#0)

**Screens**

dynamic_theming

medium_and_large_display

reply_large_screen

mobile_tablet

**Known issues**

_b/232495216_ -> Adjustable width of permanent navigation drawer.

Screen Shot 2022-07-11 at 19 16 22

_b/232495216_ -> Navigation Rail content positioning support (Right now it's manually implemented using Layout)

Screen Shot 2022-06-22 at 17 21 02

 

**Future development**

- Add motions and transitions in state changes
- Support for Material-shapes
- Create Email/Reply thread adaptive screens


Reply/README.md
```diff
@@ -0,0 +1,104 @@
...
+* [`LargeFloatingActionButton`](app/src/main/java/com/example/reply/ui/ReplyListContent.kt#100) is used along with bottom navigation ber.
+* [`FloatingActionButton`](app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationComponents.kt#87) is used with Navigation rail for medium to large tablets.
+* [`ExtendedFloatingActionButton`](app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationComponents.kt#214) is used in Navigation drawer for large devices.
+
+#### [Data](app/src/main/java/com/example/reply/data)
+Reply has static local data providers for [email](app/src/main/java/com/example/reply/data/local/LocalEmailsDataProvider.kt) and [account](app/src/main/java/com/example/reply/data/local/LocalAccountsDataProvider.kt) data. It is also using repository pattern where [EmailRepository](app/src/main/java/com/example/reply/data/EmailsRepository.kt) 
+emits the flow of email from local data that is used in [ReplyHomeViewModel](app/src/main/java/com/example/reply/ui/ReplyHomeViewModel.kt) to observe 
+it in view model scope. ViewModel expose this data to ReplyApp composable via [state flow](app/src/main/java/com/example/reply/ui/ReplyHomeViewModel.kt#34).
```
riggaroo, line 83:
```suggestion
it in view model scope. The `ViewModel` exposes this data to ReplyApp composable via [state flow](app/src/main/java/com/example/reply/ui/ReplyHomeViewModel.kt#34).
```

Reply/app/src/main/java/com/example/reply/data/Account.kt
```diff
@@ -0,0 +1,36 @@
...
+ * multiple accounts.
+ */
+data class Account(
+    val id: Long,
+    val uid: Long,
+    val firstName: String,
+    val lastName: String,
+    val email: String,
+    val altEmail: String,
+    @DrawableRes val avatar: Int,
```
riggaroo, line 32: [suggestion] referencing a DrawableRes for a data class that generally could be stored to disk (or come from a server), is not a great idea as the resources aren't stable values (when you build a new APK it might change the value). Have you considered using a url here instead?

Reply/app/src/main/java/com/example/reply/data/Email.kt
```diff
@@ -0,0 +1,45 @@
...
+    val subject: String = "",
+    val body: String = "",
+    val attachments: List<EmailAttachment> = emptyList(),
+    var isImportant: Boolean = false,
+    var isStarred: Boolean = false,
+    var mailbox: MailboxType = MailboxType.INBOX,
+    var createAt: String,
+    val threads: List<Email> = emptyList()
+) {
+    val senderPreview: String = "${sender.fullName} - 4 hrs ago"
```
riggaroo, line 37: [nit] This should be a formatter taken from strings.xml so the text is not hardcoded.

Reply/app/src/main/java/com/example/reply/data/Email.kt
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * A simple data class to represent an Email.
+ */
+data class Email(
+    val id: Long,
+    val sender: Account,
+    val recipients: List<Account> = emptyList(),
+    val subject: String = "",
+    val body: String = "",
```
riggaroo, line 29: [nit] I generally found if I am making a variable have a default value of an empty string, then it should probably be a nullable type instead, with the default set to null. This helps having to avoid understanding what an empty string means - and avoids needing to do content checks to.

Reply/app/src/main/java/com/example/reply/data/local/LocalEmailsDataProvider.kt
```diff
@@ -0,0 +1,333 @@
...
+            """
+                I'll be in your neighborhood doing errands and was hoping to catch you for a coffee this Saturday. If you don't have anything scheduled, it would be great to see you! It feels like its been forever.
+
+                If we do get a chance to get together, remind me to tell you about Kim. She stopped over at the house to say hey to the kids and told me all about her trip to Mexico.
+
+                Talk to you soon,
+
+                Ali
+            """.trimIndent(),
+            createAt = "40 mins ago",
```
riggaroo, line 97: [suggestion] could be nice to have this as actual dates and then compute these strings instead, but I understand its a sample :)

Reply/app/src/main/java/com/example/reply/ui/ReplyApp.kt
```diff
@@ -0,0 +1,288 @@
...
+        navigationContentPosition = navigationContentPosition,
+        replyHomeUIState = replyHomeUIState,
+        closeDetailScreen = closeDetailScreen,
+        navigateToDetail = navigateToDetail
+    )
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+private fun ReplyNavigationWrapperUI(
```
riggaroo, line 132: [nit] I haven't seen us use a UI suffix on composables - perhaps because all composables are UI so it doesn't need it?

riggaroo (commented):
This is overall looking really great - lovely write up and will be super useful for developers, especially the theming stuff 🎉 🙌🏻 . 

I didn't get to review all the code yet, but a few overall suggestions first: 

`Reply/app/src/main/res/drawable/paris_1-4.jpg` can you reduce these image sizes or load them up from somewhere else? This will bloat the repo size as each is a couple of mbs. 

I don't see a Github action file added to the repo to run the build for this sample, can you add one? As well as i'm not sure if spotless has been run on the code yet.

The Reply/app/src/main/res/mipmap-hdpi/ic_launcher.png mipmap icons seem like they have a shadow on them that looks unintentional, can you double check? 

The avatar pictures are all different sizes, aspect ratios and some are .jpeg vs .jpg can you crop them to make them consistent?

Reply/app/src/main/java/com/example/reply/data/Account.kt
Gurupreet: Hey I am using the local drawable's for this sample and local dummy data is created so it will always have reference. Do you mean url as in upload storage and url ?

Reply/app/src/main/java/com/example/reply/data/Email.kt
Gurupreet: Agreed, let me update it.

Gurupreet: It was unused so removed unused params. Thanks

Reply/app/src/main/java/com/example/reply/ui/MainActivity.kt
```diff
@@ -0,0 +1,163 @@
...
+}
+
+@OptIn(ExperimentalMaterial3WindowSizeClassApi::class)
+@Preview(showBackground = true, widthDp = 700, heightDp = 500)
+@Composable
+fun ReplyAppPreviewTablet() {
+    ReplyTheme {
+        ReplyApp(
+            replyHomeUIState = ReplyHomeUIState(emails = LocalEmailsDataProvider.allEmails),
+            windowSize = WindowSizeClass.calculateFromSize(DpSize(700.dp, 500.dp)),
```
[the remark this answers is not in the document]
Gurupreet: The window size calculation is outside the ReplyApp Composable with activity context, and window size param needs a default value.

Reply/app/src/main/java/com/example/reply/data/Email.kt
Gurupreet: Yes it shouldn't have default string values and non nullable as well because it's email's subject and body. removed empty strings.

Reply/app/src/main/java/com/example/reply/data/Account.kt
riggaroo: As long as its not stored to disk and then the reference read from storage on a new app run it should work, but my suggestion I guess is that this isn't really a good practise since this kind of data normally comes from a server, I guess its fine for this instance though 🤷🏻

Reply/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+This sample showcases:
+
+* Adaptive apps for mobile, tablets and foldables 
+* Material navigation components
+* [Material 3 theming][materialtheming] & dynamic colors.
+
+## Screenshots
+
+<img src="screenshots/reply.gif"/>
```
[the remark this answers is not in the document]
riggaroo: I would just change the title to "Designs" then

.github/workflows/Reply.yaml
```diff
@@ -0,0 +1,77 @@
...
+        working-directory: ${{ env.SAMPLE_PATH }}
+        run: ./gradlew assembleDebug --stacktrace
+
+      - name: Build release
+        working-directory: ${{ env.SAMPLE_PATH }}
+        run: ./gradlew assembleRelease --stacktrace
+
+      - name: Run local tests
+        working-directory: ${{ env.SAMPLE_PATH }}
+        run: ./gradlew testDebug --stacktrace
```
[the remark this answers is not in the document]
Gurupreet: Yes I used the JetCaster one, let me see what's the difference.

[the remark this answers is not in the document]
Gurupreet: ahh I see JetCaster is the old one other flows are updated to latest 5 days ago. what a luck 🥇. Updated.

riggaroo:
One other bit of feedback is that the two icons in the bottom bar look very similar and I didn't notice or visually understand the difference, but this is all design feedback.

Gurupreet:
> Nice stuff so far @Gurupreet 🙌🏻
> 
> I'm wondering what the spec for the size of the FAB at the bottom is? Is it correct, it seems a bit large compared to FABs I've seen before Screenshot_20220714_103747
> 
> Another question around the color choices, where does the green and yellow come from? I don't believe I've set this as a theme on my device - Screenshot_20220714_104354 The above screenshot also shows that in dark mode the color difference between the "search bar" and the background isn't as pronounced as the light mode version?

Thanks @riggaroo first of all for very good feedback

- Yes design is using Large-Fab for mobile. I will check with designer if it's intentional
- Reply is using 2 themes, if you are using android 12 and above it will falls to dynamic theming by default otherwise the yellow theme for older versions. were you using android 12 or above ? I will check it should be dynamic theme.
- Thanks it's a surface color for background of search bar, let me check if some issue in dark mode :)

Gurupreet:
> One other bit of feedback is that the two icons in the bottom bar look very similar and I didn't notice or visually understand the difference, but this is all design feedback.

Thanks I checked with designer and updated it to outlined chat icon 💯

Reply/app/src/main/java/com/example/reply/ui/utils/WindowStateUtils.kt
```diff
@@ -0,0 +1,72 @@
...
+ */
+enum class ReplyNavigationContentPosition {
+    TOP, CENTER
+}
+
+/**
+ * App Content shown depending on device size and state.
+ */
+enum class ReplyContentType {
+    LIST_ONLY, LIST_AND_DETAIL
```
alexvanyo, line 71: One small naming thing after this has all come together:

"list only" is a bit of a misnomer, since it could be showing either just the list or just the detail. Maybe `SINGLE_PANE` and `DUAL_PANE` or something similar would be more appropriate?

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
```diff
@@ -0,0 +1,184 @@
...
+    replyHomeUIState: ReplyHomeUIState,
+    navigationType: ReplyNavigationType,
+    closeDetailScreen: () -> Unit,
+    navigateToDetail: (Long) -> Unit
+) {
+    /**
+     * When moving from LIST page to LIST_AND_DETAIL page on screen expand, set first email selected as default
+     * When moving from LIST_AND_DETAIL page to LIST page clear the selection and user should see LIST screen.
+     */
+    LaunchedEffect(key1 = contentType) {
```
alexvanyo, line 63: I'm confused by this logic a little bit.

Right now, there's this behavior:
- Reading a detail while on a small screen
- Cause an activity recreation (resize, other config change)
- I am kicked out of the detail I was reading

In that case, I'd expect to continue reading the detail I was on.

Reply/app/src/main/java/com/example/reply/ui/utils/WindowStateUtils.kt
Gurupreet: Yes I think Single and Dual pane makes more sense. Let me update.

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
Gurupreet: Ok I updated it to following -> 

- It will preserve the selected detail page from SINGLE to DUAL PANE
- From DUAL to SINGLE pane it will be reset to SINGLE pane(No detail) but maintaining the scroll state.

alexvanyo: I think the same issue is still present for activity recreation, since the `LaunchedEffect` isn't only running when the content type changes, it also runs on the first time (after recreation).

So if I'm reading a detail, then an activity recreation happens, the `LaunchedEffect` will run again and close the detail screen I was reading even if it remains in `SINGLE_PANE` the entire time. I think that should be fixable with some sort of additional `var previousContentType`?

Gurupreet: I used the replyHomeUIState's showSingleScreen to decide and it also survives configuration changes 👍🏽

alexvanyo (approved):
Looks good to me!

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
```diff
@@ -0,0 +1,185 @@
...
+                        modifier = Modifier.size(28.dp)
+                    )
+                }
+            }
+        }
+    }
+}
+
+@Composable
+fun ReplyListOnlyContent(
```
alexvanyo, line 114: Nit: This is still `ListOnly`, even though it might show either the detail or the list.

Reply/app/src/main/java/com/example/reply/ui/ReplyApp.kt
```diff
@@ -0,0 +1,286 @@
...
+    if (navigationType == ReplyNavigationType.PERMANENT_NAVIGATION_DRAWER) {
+        // TODO check on custom width of PermanentNavigationDrawer: b/232495216
+        PermanentNavigationDrawer(drawerContent = {
+            NavigationDrawerContent(
+                selectedDestination = selectedDestination,
+                navigationContentPosition = navigationContentPosition,
+                navigateToTopLevelDestination = navigationActions::navigateTo,
+            )
+        }) {
+            ReplyAppContent(
```
alexvanyo, line 158: This is slightly more advanced, but we should add in a `movableContentOf` usage here so that any and all state within `ReplyAppContent` is preserved even if the navigation type changes.

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
```diff
@@ -0,0 +1,185 @@
...
+     */
+
+    LaunchedEffect(key1 = contentType) {
+        if (contentType == ReplyContentType.DUAL_PANE && replyHomeUIState.selectedEmail == null) {
+            replyHomeUIState.emails.firstOrNull()?.let { firstEmail ->
+                navigateToDetail.invoke(firstEmail.id, ReplyContentType.DUAL_PANE)
+            }
+        } else if (contentType == ReplyContentType.SINGLE_PANE && !replyHomeUIState.showDetailScreenOnly) {
+            closeDetailScreen.invoke()
+        }
```
alexvanyo, line 71: Nit: Looking at this some more, I think this is a composition loop that we can avoid with a bit more careful state handling.

The issue here is that `replyHomeUiState.selectedEmail` is doing two things: it is managing which email to show in the detail panel (if any) and it is also managing whether or not to show the detail.

I think if we split apart those issues and combine them back together, we can avoid the need for a `LaunchedEffect` entirely.

Let's say that we added a `isDetailOpen: Boolean` field to `replyHomeUiState`. This field would be set to `true` upon `navigateToDetail`, and set to `false` on `closeDetailScreen`.

We'd still have `selectedEmail`, but instead of starting at `null`, we could just set it as the first email in the list right out of the gate.

I think that would remove the need for `showDetailScreenOnly` as well, which feels a bit strange to have a `ViewModel` trying to know the state of the UI.

alexvanyo (changes requested):

Reply/app/src/main/java/com/example/reply/ui/ReplyApp.kt
Gurupreet: I have never heard/used moveableContentOf, let me check it out :)

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
Gurupreet: Updated to `ReplySinglePaneContent` 👍🏽

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
Gurupreet: Thanks for suggestion I removed the `selectedEmail` dependency and made first email as default. But we will still need LaunchEffect to close List/Detail combination  when moving from Single pane not selected to Double-pane and back.

https://user-images.githubusercontent.com/4903762/180475948-ea1f3a4b-ac97-4aae-932d-4f445b6d28c3.mov

Reply/app/src/main/java/com/example/reply/data/local/LocalEmailsDataProvider.kt
```diff
@@ -0,0 +1,336 @@
...
+import com.example.reply.data.EmailAttachment
+import com.example.reply.data.MailboxType
+
+/**
+ * A static data store of [Email]s.
+ */
+
+object LocalEmailsDataProvider {
+
+    private val threads = listOf(
```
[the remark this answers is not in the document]
Gurupreet: I don't think listOf can take named params. I will check other places if not using named params where it should be 👍🏽

Reply/app/src/main/java/com/example/reply/data/Email.kt
```diff
@@ -0,0 +1,34 @@
...
+    val id: Long,
+    val sender: Account,
+    val recipients: List<Account> = emptyList(),
+    val subject: String,
+    val body: String,
+    val attachments: List<EmailAttachment> = emptyList(),
+    var isImportant: Boolean = false,
+    var isStarred: Boolean = false,
+    var mailbox: MailboxType = MailboxType.INBOX,
+    var createAt: String,
```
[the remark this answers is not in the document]
Gurupreet: oh thanks that's correct. Fixed

Reply/app/src/main/java/com/example/reply/ui/EmptyComingSoon.kt
```diff
@@ -0,0 +1,57 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.example.reply.R
+
+@Composable
+@Preview
```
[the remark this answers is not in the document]
Gurupreet: Done 💯 Will keep in mind of this 👍🏽

Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationActions.kt
```diff
@@ -0,0 +1,88 @@
...
+}
+
+data class ReplyTopLevelDestination(
+    val route: String,
+    val selectedIcon: ImageVector,
+    val unselectedIcon: ImageVector,
+    val iconTextId: Int
+)
+
+class ReplyNavigationActions(private val navController: NavHostController) {
```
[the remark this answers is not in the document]
Gurupreet: I think it can be extended to add more actions in updates like -> navigateToCreateEmail etc or extension approach is better for future actions as well ?

Reply/app/src/main/java/com/example/reply/ui/components/ReplyProfileImage.kt
```diff
@@ -0,0 +1,39 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun ReplyProfileImage(
+    drawableResource: Int,
+    description: String,
+    modifier: Modifier = Modifier.size(40.dp),
```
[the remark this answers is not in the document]
Gurupreet: Not sure if we reached to an agreement on if it should be discarded? but I think for samples we don't use it so I will remove it 👍🏽

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
```diff
@@ -0,0 +1,179 @@
...
+        ReplyEmailDetail(email = replyHomeUIState.selectedEmail) {
+            closeDetailScreen.invoke()
+        }
+    } else {
+        LazyColumn(modifier = modifier, state = emailLazyListState) {
+            item {
+                ReplySearchBar(modifier = Modifier.fillMaxWidth())
+            }
+            items(replyHomeUIState.emails) { email ->
```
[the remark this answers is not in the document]
Gurupreet: Added to all lists 💯

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
```diff
@@ -0,0 +1,179 @@
...
+    /**
+     * When moving from LIST_AND_DETAIL page to LIST page clear the selection and user should see LIST screen.
+     */
+    LaunchedEffect(key1 = contentType) {
+        if (contentType == ReplyContentType.SINGLE_PANE && !replyHomeUIState.isDetailOnlyOpen) {
+            closeDetailScreen.invoke()
+        }
+    }
+
+    val emailLazyListState = rememberLazyListState()
```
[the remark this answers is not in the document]
Gurupreet: Haha I am using it to remember scroll state positions during size/fold/other state changes so scroll is remembered when UI goes from single to double pane etc.

Reply/app/src/main/java/com/example/reply/ui/MainActivity.kt
```diff
@@ -0,0 +1,163 @@
...
+import com.example.reply.ui.utils.DevicePosture
+import com.example.reply.ui.utils.isBookPosture
+import com.example.reply.ui.utils.isSeparating
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.stateIn
+
+class MainActivity : ComponentActivity() {
+
+    private val viewModel: ReplyHomeViewModel by viewModels()
```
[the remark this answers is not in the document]
Gurupreet: Hmm I use this for Activity viewmodel since it's extension to ComponentActivity and tends to use `androidx.lifecycle.viewmodel.compose.viewModel` when I have Compose screen only viewModel  eg: DetailScreenViewModel. 

@manuelvicnt would love to know what do you think is suggested way to use for MainActivity etc in Compose only app.

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
alexvanyo: Got it, that usage of `LaunchedEffect` makes sense to me for that behavior.

Although taking a step back, is that desired? I feel like as a user I'd expect to stay in the same place upon rotating back if I don't interact with the app at all. Perhaps I rotated accidentally, or maybe I was closing my foldable (causing the app to appear on the front screen), and then upon reopening I want to return to where I was before.

Gurupreet: Tbh it's something I just followed Material large screen guidance, I can check with them what was the reason/research behind it. 
For me I can understand this flow seems correct -> I expand device I see detail without touching anything if fold back I should be same state as before (Original list).

Reply/app/src/main/java/com/example/reply/ui/ReplyApp.kt
Gurupreet: Keeping it here for consideration and sharing.

```
val currentContentType by rememberUpdatedState(contentType)
val currentReplyHomeState by rememberUpdatedState(replyHomeUiState)

val rememberedReplyAppContent = remember {
    movableContentOf {
        ReplyNavHost(
            navController = navController,
            contentType = currentContentType,
            replyHomeUIState = currentReplyHomeState,
            navigationType = navigationType,
            closeDetailScreen = closeDetailScreen,
            navigateToDetail = navigateToDetail,
            modifier = Modifier.fillMaxSize(),
        )
    }
}

```

Reply/app/src/main/java/com/example/reply/data/local/LocalEmailsDataProvider.kt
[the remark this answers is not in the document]
simona-anomis: Placed the comment in the wrong place, sorry! I meant when you're constructing the Emails:

        Email(
            8L,
            LocalAccountsDataProvider.getContactAccountByUid(13L),
            listOf(LocalAccountsDataProvider.getDefaultUserAccount()),
            "Your update on Google Play Store is live!",
            """
              Your update, 0.1.1, is now live on the Play Store and available for your alpha users to start testing.
              
              Your alpha testers will be automatically notified. If you'd rather send them a link directly, go to your Google Play Console and follow the instructions for obtaining an open alpha testing link.
            """.trimIndent(),
            mailbox = MailboxType.TRASH,
            createAt = "3 hours ago",
        ),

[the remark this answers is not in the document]
Gurupreet: Got it fixed.

Reply/app/src/main/java/com/example/reply/ui/ReplyApp.kt
alexvanyo: Agreed to take a look at `movableContentOf` as a follow-up!

alexvanyo (approved):
Looks good from me!

bentrengrove (approved):
Nice work! A bunch of nits to improve in the future.

The only ones I think you should fix before merge is upgrading those alpha dependencies and possibly signing with a debug key.

The rest can be done in subsequent PRs

Reply/app/build.gradle
```diff
@@ -0,0 +1,94 @@
...
+        kotlinCompilerExtensionVersion compose_compiler_version
+    }
+    packagingOptions {
+        resources {
+            excludes += '/META-INF/{AL2.0,LGPL2.1}'
+        }
+    }
+}
+
+dependencies {
```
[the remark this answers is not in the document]
Gurupreet: Yes will do in future PR 👍🏽

Reply/app/build.gradle
```diff
@@ -0,0 +1,94 @@
...
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        vectorDrawables {
+            useSupportLibrary true
+        }
+    }
+
+    buildTypes {
+        release {
+            minifyEnabled false
```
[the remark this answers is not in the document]
Gurupreet: Thanks, added 👍🏽

Reply/app/src/main/java/com/example/reply/ui/components/ReplyAppBars.kt
```diff
@@ -0,0 +1,136 @@
...
+    }
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun EmailDetailAppBar(
+    email: Email,
+    isFullScreen: Boolean,
+    modifier: Modifier = Modifier,
+    onBackPressed: () -> Unit
```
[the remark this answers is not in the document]
Gurupreet: I follow this when we are not passing any content to composable but just lambdas actions, they all goes to last like below: 

`@Composable
fun ReplySinglePaneContent(
    replyHomeUIState: ReplyHomeUIState,
    emailLazyListState: LazyListState,
    closeDetailScreen: () -> Unit,
    navigateToDetail: (Long, ReplyContentType) -> Unit
) `

But I think you are right we should't write the implementation as above but like this:

`   ReplySinglePaneContent(
                replyHomeUIState = replyHomeUIState,
                emailLazyListState = emailLazyListState,
                modifier = Modifier.fillMaxSize(),
                closeDetailScreen = closeDetailScreen,
                navigateToDetail = navigateToDetail
            )`
            
            is it preferred way to write last function ?

Reply/app/src/main/java/com/example/reply/ui/MainActivity.kt
```diff
@@ -0,0 +1,163 @@
...
+                scope = lifecycleScope,
+                started = SharingStarted.Eagerly,
+                initialValue = DevicePosture.NormalPosture
+            )
+
+        setContent {
+            ReplyTheme {
+                val windowSize = calculateWindowSizeClass(this)
+                val devicePosture = devicePostureFlow.collectAsState().value
+                val uiState = viewModel.uiState.collectAsState().value
```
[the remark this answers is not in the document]
Gurupreet: Done 👍🏽

Reply/app/src/main/java/com/example/reply/ui/components/ReplyAppBars.kt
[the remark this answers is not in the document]
bentrengrove: Yes, it is better to use named parameters when it doesn't make sense to use trailing lambda syntax.

But the other thing is `Modifier` should always come as the first optional parameter, it should never have non-optional parameters after it except for content lambdas. https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#elements-accept-and-respect-a-modifier-parameter

PR #890, opened 2022-07-19
[Jetcaster] Remove constraintlayout

alanwgeorge:
ConstraintLayout solves a problem that does not exist in Compose.  It's single use didn't seem justified.

image

Also added a nav bar spacer for the episodes scroll list so the list end is not under the system UI.

image

image

### DIFF
--- a/Jetcaster/app/build.gradle
+++ b/Jetcaster/app/build.gradle
@@ -101,8 +101,6 @@ dependencies {
 
     implementation Libs.AndroidX.Activity.activityCompose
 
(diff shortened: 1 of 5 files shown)


riggaroo:
Hi @alanwgeorge Thanks for the contribution! We are in the process of updating our guidance on ConstraintLayout - the tl;dr is that it is valuable for relative positioning of UI elements, and if you want a bit of better readability over a lot of nested Columns and Rows. For now I think we'd like to leave this example usage of ConstraintLayout in the sample as its a showcase of different Compose APIs. Feel free to choose what works best for your use case though.

PR #901, opened 2022-07-27
[All] Update to Compose 1.2 snapshot

florina-muntenescu:


JetNews/build.gradle
```diff
@@ -34,7 +35,7 @@ buildscript {
 }
 
 plugins {
-    id 'com.diffplug.spotless' version '6.7.0'
+    id 'com.diffplug.spotless' version '6.5.2'
```
JolandaVerhoef, line 38: Is there a reason you're downgrading?

simona-anomis: We're upgrading spotless to 6.7.0 but leaving ktlint as it is right now, because of an issue:

https://github.com/pinterest/ktlint/issues/1195

simona-anomis (approved):

Issue #900, opened 2022-07-27
[Bug][Reply]: Using deprecated getWindowInsetsController method in Theme.kt 


florina-muntenescu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### Is this an issue related to one of the samples?

- [X] Yes, this is a specific issue related to this samples repo.

### Sample app

Other (bug not related to sample app)

### What happened?

Reply is using a deprecated `getWindowInsetsController` method in `Theme.kt`

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #902, opened 2022-07-27
[Reply] Fix: Removed deprecated windowInsetController with WindowCompat.

Gurupreet:
fix: #900 
Removing deprecated `ViewCompat.windowInsetsController` with suggested `WIndowCompat.insetsController`

### DIFF
--- a/Reply/app/src/main/java/com/example/reply/ui/theme/Theme.kt
+++ b/Reply/app/src/main/java/com/example/reply/ui/theme/Theme.kt
@@ -29,7 +29,7 @@ import androidx.compose.runtime.SideEffect
 import androidx.compose.ui.graphics.toArgb
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalView
-import androidx.core.view.ViewCompat
+import androidx.core.view.WindowCompat
 
 // Material 3 color schemes
 private val replyDarkColorScheme = darkColorScheme(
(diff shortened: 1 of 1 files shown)


florina-muntenescu (approved):

Reply/app/src/main/java/com/example/reply/ui/theme/Theme.kt
```diff
@@ -107,8 +107,9 @@ fun ReplyTheme(
     val view = LocalView.current
     if (!view.isInEditMode) {
         SideEffect {
-            (view.context as Activity).window.statusBarColor = replyColorScheme.primary.toArgb()
-            ViewCompat.getWindowInsetsController(view)?.isAppearanceLightStatusBars = darkTheme
```
Gurupreet, line 111: This is actually from Material theme builder exported M3 themes. I will raise the fix there as well.

alexvanyo: Looks good! We also could consider using `accompanist/systemuicontroller` instead, which makes this a bit nicer as well.

Gurupreet: Yes I thought about it, Jet samples already showcasing use of `accompanist/systemuicontroller `  so I kept it this way to showcase other way. But let me think on it if there are more usecase of adding accompanist.

PR #903, opened 2022-07-27
Added Reply to bug and feature template for github

Gurupreet:
Reply is new addition to compose sample, so adding it to bug and feature request template for compose samples github.


florina-muntenescu:
Could we remove Rally from here as well?

florina-muntenescu (approved):

PR #912, opened 2022-07-31
[Jetsnack] - Start using version catalog for dependency management

oas004:
First of all, thank you guys for such a awesome collection of compose samples! I really like it! 🤩 

In this PR I have updated the JetSnack project to start using version catalog for dependency management.

I encountered one issue with regards to using the plugin part in the app gradle file.  From what I understand there is an open issue on [Providing a way to use plugin definition from version catalog without version](https://github.com/gradle/gradle/issues/17968). I managed to get around this by converting the gradle file to Kotlin DSL. Would really like some extra feedback on that if someone has time 😄 

---
name: [Jetsnack] - Start using version catalog for dependency management
about: Updating the JetSnack project to start using version catalog
label: 'triage me'
---
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Ensure the tests and linter pass
- [x] Appropriate docs were updated (if necessary)


bentrengrove:
Thank you very much for your contribution, it looks really good! Unfortunately I won't be able to merge it because it would disturb how we manage updating all of our samples. In the future, I would encourage you to open an issue before working on something like this to ensure we could support it.

oas004:
> Thank you very much for your contribution, it looks really good! Unfortunately I won't be able to merge it because it would disturb how we manage updating all of our samples. In the future, I would encourage you to open an issue before working on something like this to ensure we could support it.

No worries at all! I didn't think about that! I will open an issue next time😊 I learned a lot for doing this anyways, so it was totally worth it 😊

Issue #911, opened 2022-07-29
[Bug]:  Constraint Layout Bug of Jetcaster


EasyVector:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### Is this an issue related to one of the samples?

- [ ] Yes, this is a specific issue related to this samples repo.

### Sample app

Jetcaster

### What happened?

So, here is the thing. I did some accessibility testing on a compose sample, Jetcaster, with the largest font size, which you could adjust inside the display setting panel of the system setting app. Just like this:

<img src="https://user-images.githubusercontent.com/25502419/181695358-018f985b-466c-4475-ad3e-424f73e82d25.png" width="250" >

But I spotted inconsistencies among those items specified with constrats inside a lazycolumn, that is, the play icon overlays the title of podcast, and they should be seperated from each other. When normal font size is applied, this issue won't occur. 

<img src="https://user-images.githubusercontent.com/25502419/181694934-4558e620-7d3f-4353-b016-94f1db102be7.png" width="250" >

So my question is, does this error come from the constraint layout system of compose or the Jetcaster itself? I tried to change the code of Jetcaster and the problem still exists. By the way, I used the latest jetpack compose. The device I tested is the emulator of Pixel 5, with the Android 12.0(API 31) system.

### Relevant logcat output

```shell
2022-07-29 14:15:26.267 5832-5841/com.example.jetcaster W/ample.jetcaste: Reducing the number of considered missed Gc histogram windows from 160 to 100
2022-07-29 14:28:15.755 5832-5859/com.example.jetcaster D/EGL_emulation: app_time_stats: avg=1437.98ms min=19.21ms max=2856.75ms count=2
2022-07-29 14:28:17.824 5832-5859/com.example.jetcaster D/EGL_emulation: app_time_stats: avg=28.45ms min=2.12ms max=1162.49ms count=52
2022-07-29 14:28:18.859 5832-5859/com.example.jetcaster D/EGL_emulation: app_time_stats: avg=26.77ms min=2.47ms max=426.97ms count=28
2022-07-29 14:28:19.867 5832-5859/com.example.jetcaster D/EGL_emulation: app_time_stats: avg=16.13ms min=2.68ms max=250.93ms count=40
2022-07-29 14:28:20.898 5832-5859/com.example.jetcaster D/EGL_emulation: app_time_stats: avg=19.46ms min=2.33ms max=212.41ms count=33
[...]

PR #913, opened 2022-07-31
Fix accessibility bugs of overlay and missing text

EasyVector:
I am the creator of issue #911, and I did some changes to the layout code inside to make this app more accessible. 

Because the title of one episode could be more than 2 lines, I commented out the corresponding statement of maxlines. And since the height of episode title is not wrap content, the play icon below could be placed above them, therefore I think it might be better to specify the width of the episode title and podcast title with `suggested wrap content (preferredWrapContent)`. The string of date could overlap the icon on its right, because the width of the string is not specified as `suggested wrap content (preferredWrapContent)`, so I changed its width as well.

Here is the effects before and after my modification.

I tested the code on the emulator of Pixel 5 with the largest font size and display size. You could see all those composables are seperated now, and the episode title could also be seen fully (2 lines -> 3 lines).

Thank you! :)

### DIFF
--- a/Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/category/PodcastCategory.kt
+++ b/Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/category/PodcastCategory.kt
@@ -186,7 +186,8 @@ fun EpisodeListItem(
 
         Text(
             text = episode.title,
-//            maxLines = 2,
+            maxLines = 2,
+            overflow = TextOverflow.Ellipsis,
             style = MaterialTheme.typography.subtitle1,
             modifier = Modifier.constrainAs(episodeTitle) {
                 linkTo(
@@ -207,7 +208,8 @@ fun EpisodeListItem(
         CompositionLocalProvider(LocalContentAlpha provides ContentAlpha.medium) {
             Text(
                 text = podcast.title,
-//                maxLines = 2,
+                maxLines = 2,
+                overflow = TextOverflow.Ellipsis,
                 style = MaterialTheme.typography.subtitle2,
                 modifier = Modifier.constrainAs(podcastTitle) {
                     linkTo(
--- a/Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/category/PodcastCategory.kt
+++ b/Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/category/PodcastCategory.kt
@@ -187,7 +187,6 @@ fun EpisodeListItem(
         Text(
             text = episode.title,
             maxLines = 2,
-            overflow = TextOverflow.Ellipsis,
             style = MaterialTheme.typography.subtitle1,
             modifier = Modifier.constrainAs(episodeTitle) {
                 linkTo(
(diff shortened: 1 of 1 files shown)


Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/category/PodcastCategory.kt
```diff
@@ -186,7 +186,7 @@ fun EpisodeListItem(
 
         Text(
             text = episode.title,
-            maxLines = 2,
+//            maxLines = 2,
```
JolandaVerhoef, line 189: What about leaving the maxLines, but adding an ellipsis? You can add `overflow = TextOverflow.Ellipsis`. Reason for this is that we don't want an extremely long title to create an extremely long list item, so some max value is needed.

JolandaVerhoef (changes requested):
Thanks for your Pull Request! I've added some comments. In addition, can you approve the CLA?

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/category/PodcastCategory.kt
EasyVector: I agree with your viewpoint, if the title is too long, then the whole lazycolumn will also be very long. At the same time, make the overflown text as consecutive dots(ellipsis notation) is necessary because it could remind users that this part of text is much longer. If we do not add overflow properties, users could be misled and unaware of the overflown text. I will add the overflow properties and fix maxLines as well. :)

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/category/PodcastCategory.kt
```diff
@@ -197,7 +197,7 @@ fun EpisodeListItem(
                     bias = 0f
                 )
                 top.linkTo(parent.top, 16.dp)
-
+                height = preferredWrapContent
```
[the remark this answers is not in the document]
EasyVector: OK! :)

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/category/PodcastCategory.kt
```diff
@@ -207,7 +207,7 @@ fun EpisodeListItem(
         CompositionLocalProvider(LocalContentAlpha provides ContentAlpha.medium) {
             Text(
                 text = podcast.title,
-                maxLines = 2,
+//                maxLines = 2,
```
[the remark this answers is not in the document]
EasyVector: OK, just leave it to me! :)

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/category/PodcastCategory.kt
EasyVector: Well, It is the first time that I opened a pull request for Google, and I have just signed the `Contributor License Agreements` on `https://cla.developers.google.com/` and sent the request to Google Gropu as well. If it is not enough, please let me know. Thank you! :)

PR #917, opened 2022-08-04
first gif breaks the design!

arviinmo:
when i watched this repo, i noticed that [this gif](https://github.com/android/compose-samples/blob/main/readme/samples_montage.gif) breaks the readme section. i already reduce some of the gif width, but could be decreased or increased.

(in below i put the screenshot of readme in repo page)
Screenshot 2022-08-04 140050


riggaroo:
Thanks for pointing this out - we merged #926 for the fix.

PR #918, opened 2022-08-04
Upgrade to Compose 1.3

bentrengrove:


JolandaVerhoef (approved):

Crane/buildSrc/src/main/java/com/example/crane/buildsrc/Dependencies.kt
```diff
@@ -59,7 +59,7 @@ object Libs {
         object Compose {
             const val compilerVersion = "1.2.0"
```
JolandaVerhoef, line 60: Should we also consider upgrading the compiler version to 1.3.0-rc02? That includes a Kotlin upgrade from 1.7.0 to 1.7.10. Might be nice to have that latest version in here as well.

JolandaVerhoef (changes requested):

Crane/buildSrc/src/main/java/com/example/crane/buildsrc/Dependencies.kt
bentrengrove: Good point, done!

JolandaVerhoef (approved):

PR #924, opened 2022-08-10
Update main | Compose 1.2.1 | Compiler 1.3.0 | Kotlin 1.7.10

JolandaVerhoef:


Crane/buildSrc/src/main/java/com/example/crane/buildsrc/Dependencies.kt
```diff
@@ -125,4 +126,6 @@ object Urls {
     const val mavenCentralSnapshotRepo = "https://oss.sonatype.org/content/repositories/snapshots/"
     const val composeSnapshotRepo = "https://androidx.dev/snapshots/builds/" +
         "${Libs.AndroidX.Compose.snapshot}/artifacts/repository/"
+    const val composeCompilerSnapshotRepo = "https://androidx.dev/snapshots/builds/" +
+        "${Libs.AndroidX.Compose.snapshot}/artifacts/repository/"
```
bentrengrove, line 130: Looks like you've referenced the wrong snapshot here

Reply/app/build.gradle
```diff
@@ -16,23 +16,19 @@
...
 
 android {
-    compileSdk 32
-
+    compileSdkVersion 32
     defaultConfig {
-        applicationId "com.example.reply"
-        minSdk 21
-        targetSdk 32
+        applicationId 'com.example.jetnews'
```
bentrengrove, line 25: This doesn't look right? Should be reply

Reply/app/build.gradle
```diff
@@ -46,35 +42,49 @@ android {
     }
 
     buildTypes {
+        debug {
+            signingConfig signingConfigs.debug
+        }
+
         release {
             minifyEnabled true
-            signingConfig signingConfigs.debug
```
bentrengrove, line 51: This should still be here, helps easy testing of performance in release

scripts/upgrade_samples.sh
```diff
@@ -27,24 +27,39 @@ set -e
...
 fi
 
+echo "Find Compose releases at: https://jetpad.corp.google.com/libraryreleases?groupId=androidx.compose.ui&artifactId=ui"
 echo "Version to change Compose to (e.g 1.2.0-alpha06): ";
 read compose_version;
 
 echo "Snapshot ID: (Blank for none)";
 read snapshot_version;
 
+echo "Find Compose Compiler releases at: https://jetpad.corp.google.com/groupreleases?groupId=androidx.compose.compiler"
```
bentrengrove, line 43: I don't think we can expose this url externally

bentrengrove (changes requested):

bentrengrove (approved):

PR #925, opened 2022-08-11
Update compose-latest | Compose 1.3.0-alpha03 | Compiler 1.3.0 | Kotlin 1.7.10

JolandaVerhoef:


scripts/upgrade_samples.sh
```diff
@@ -27,30 +27,39 @@ set -e
...
 # Check for clean git status
 if [[ `git status --porcelain` ]]; then
-    read -r -p "You have uncommited git changes. Are you sure you want to continue? [y/N] " response
+    read -r -p "You have uncommited git changes. Are you sure you want to continue? [Y/n] " response
     if [[ "$response" =~ ^([nN])$ ]]; then
         exit 0;
     fi
 fi
 
+echo "Find Compose releases at: https://jetpad.corp.google.com/libraryreleases?groupId=androidx.compose.ui&artifactId=ui"
```
bentrengrove, line 36: I don't think we can expose this url

bentrengrove (approved):
Just that one comment and then good to go

PR #926, opened 2022-08-11
Fix readme image size

riggaroo:


IanGClifton:
Is there any value in keeping the gif itself larger than the display size?

riggaroo:
Probably not, but this is just a fix to fix rendering on the initial github landing screen as its broken right now...

IanGClifton (approved):

PR #942, opened 2022-08-16
Crypto app UI

DeveloperRahnuma:


IanGClifton:
Hi Rahnuma, thank you for your interest in Compose and our samples. Unfortunately, we can't accept completely new samples. Beyond the initial review time, there's a lot of maintenance work that goes into updating every sample with the latest library updates, so we have to be selective about what APIs and functionality each sample covers.

If there's any specific features that you feel aren't demonstrated in the existing samples, you can create an issue to suggest them. If it's something you want to implement, you can also follow the [contributing guidelines](https://github.com/android/compose-samples/blob/main/CONTRIBUTING.md). We highly recommend creating an issue prior to submitting any pull requests because it gives us context and lets us discuss the right approach before seeing the PR for the first time.

PR #944, opened 2022-08-17
refactor: remove unused annotation

solomaticcs:


JolandaVerhoef:
Thank you for your contribution! It seems like our syntax checker is failing, as there's unused imports. Can you remove the imports that became unused because of the annotations you removed?

PR #947, opened 2022-08-17
[Jetcaster] Add TwoPane from accompanist/adaptive for player UI

alexvanyo:
Using `TwoPane` from accompanist/adaptive to support postures and adaptive layout for the player screen in Jetcaster.


riggaroo:
This can probably be a PR into `main` @alexvanyo as it doesn't seem to be an alpha or beta of compose that its depending on.

alexvanyo:
@riggaroo accompanist/adaptive depends on Compose 1.3 for [`PlacementScope.coordinates`](https://developer.android.com/reference/kotlin/androidx/compose/ui/layout/Placeable.PlacementScope#coordinates()), and I think we don't want `main` to depend on 1.3 yet?

riggaroo:
Ah cool, yeah thats fine, thanks!

Jetcaster/app/build.gradle.kts
```diff
@@ -109,6 +110,7 @@ dependencies {
 
     implementation(libs.androidx.window)
 
+    implementation(libs.accompanist.adaptive)
```
Gurupreet, line 113: @alexvanyo So two pane will be part of accompanist adaptive?

alexvanyo: Yes, at least for the time being to explore its API shape and get feedback.

alexvanyo:
@Gurupreet could you take another look here?

This is now targeted against main since Compose 1.3 is stable.

Owl/gradle/libs.versions.toml
```diff
@@ -50,6 +50,7 @@ secrets = "2.0.1"
 targetSdk = "33"
 
 [libraries]
+accompanist-adaptive = { module = "com.google.accompanist:accompanist-adaptive", version.ref = "accompanist" }
```
Gurupreet, line 53: @alexvanyo is there any reason it's added to all samples?

Gurupreet (approved):
Looks good! Can you add screenshot for future to visualise table top etc

Owl/gradle/libs.versions.toml
riggaroo: Yup - exactly that, it doesn't getting added to the actual import, just the toml file.

alexvanyo:
@Gurupreet added an image and updated the README in the latest commit!

PR #948, opened 2022-08-18
[All] Update all samples to version catalogs and kts

bentrengrove:
- Updates all samples to use version catalogs and kts
- Share one global version file to make updating samples easier
- Remove upgrade script as it's not needed anymore


Jetsurvey/gradle/libs.versions.toml.ignore
```diff
@@ -0,0 +1,66 @@
+[versions]
+androidGradlePlugin = "7.2.1"
```
riggaroo, line 2: Was this file meant to be committed? .ignore?

bentrengrove: Oops, no. Needs to be deleted

PR #954, opened 2022-08-19
[Jetsurvey] Converted to use Material 3

IanGClifton:
This updates Jetsurvey from Material 2 to Material 3. All Compose components now use Material 3 instead of Material 2. The date picker is currently still using the MaterialDatePicker DialogFragment the same as before this change. This also fixes the visibility of the status bars when in light mode and updates the screenshot and animated gif.

Before: https://raw.githubusercontent.com/android/compose-samples/70d05efe9c5de6d3788ac3c6023caa20761d410a/Jetsurvey/screenshots/survey.gif
After: https://raw.githubusercontent.com/android/compose-samples/ede93fc907ad32c61bd7f3d55a33af48e782ea2a/Jetsurvey/screenshots/survey.gif


Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInScreen.kt
```diff
@@ -171,16 +171,16 @@ fun ErrorSnackbar(
...
                     )
                 },
                 action = {
-                    data.actionLabel?.let {
+                    data.visuals.actionLabel?.let {
                         TextButton(onClick = onDismiss) {
                             Text(
                                 text = stringResource(id = R.string.dismiss),
-                                color = MaterialTheme.colors.snackbarAction
+                                color = MaterialTheme.colorScheme.inversePrimary
```
ricknout, line 183: This appears to be the default color so you can probably remove this:
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/tokens/SnackbarTokens.kt;l=26

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInSignUp.kt
```diff
@@ -96,13 +92,13 @@ fun SignInSignUpScreen(
     }
 }
 
+@OptIn(ExperimentalMaterial3Api::class) // SmallTopAppBar is experimental in m3
```
ricknout, line 95: I think this comment should have `CenterAlignedTopAppBar` instead?

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInSignUp.kt
```diff
@@ -191,14 +187,13 @@ fun Password(
...
-            CompositionLocalProvider(LocalContentAlpha provides ContentAlpha.medium) {
-                Text(
-                    text = label,
-                    style = MaterialTheme.typography.body2
-                )
-            }
+            Text(
+                text = label,
+                style = MaterialTheme.typography.bodyMedium,
+                color = MaterialTheme.colorScheme.onSurface.copy(alpha = 0.6f)
```
ricknout, line 195: Ideally this should rather be using `colorScheme.onSurfaceVariant` as an equivalent to `ContentAlpha.medium`. Was this alpha value specified in the design? Same goes for others below. If not, a constant for `0.6f` would be great.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInSignUp.kt
```diff
@@ -261,15 +256,12 @@ fun OrSignInAsGuest(
...
-        Surface {
-            CompositionLocalProvider(LocalContentAlpha provides ContentAlpha.medium) {
-                Text(
-                    text = stringResource(id = R.string.or),
-                    style = MaterialTheme.typography.subtitle2,
-                    modifier = Modifier.paddingFromBaseline(top = 25.dp)
-                )
-            }
-        }
+        Text(
```
ricknout, line 259: Is there a reason there was a `Surface` wrapping this before? Is it no longer needed?

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/WelcomeScreen.kt
```diff
@@ -120,7 +120,7 @@ private fun Branding(modifier: Modifier = Modifier) {
 @Composable
 private fun Logo(
     modifier: Modifier = Modifier,
-    lightTheme: Boolean = MaterialTheme.colors.isLight
+    lightTheme: Boolean = surfaceIsLight()
```
ricknout, line 123: Having this as the default value implies that this composable should basically only ever be displayed on top of a `Surface`, but there isn't anything in place to enforce that. I'd prefer no default value and this value passed through at the call site where a `Surface` is present. Same goes for other usages.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
```diff
@@ -235,22 +233,18 @@ private fun QuestionContent(
...
-                color = backgroundColor,
+                color = MaterialTheme.colorScheme.inverseOnSurface,
                 shape = MaterialTheme.shapes.small
             )
     ) {
         Text(
             text = stringResource(id = title),
-            style = MaterialTheme.typography.subtitle1,
+            style = MaterialTheme.typography.titleMedium,
+            color = MaterialTheme.colorScheme.onSurface.copy(alpha = 0.87f),
```
ricknout, line 247: I'm still a bit unsure about these alpha values, but I see they're coming from the designs. In general it would be great to have them as constants somewhere.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/theme/Theme.kt
```diff
@@ -17,60 +17,104 @@
...
+        DarkColors
+    }
+
+    val systemUiController = rememberSystemUiController()
+    DisposableEffect(systemUiController, useDarkTheme) {
+        systemUiController.setSystemBarsColor(
+            color = colors.surface,
+            darkIcons = !useDarkTheme
+        ) {
+            md_theme_dark_surface
```
ricknout, line 105: I think it might be better to leave this out, having dug into the Accompanist API a bit. WDYT?

ricknout (commented):
Looking great! Please can you include before/after screenshots for M2 and M3 (in both light and dark themes)

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInScreen.kt
IanGClifton: If I remove this line, the action ends up being shown in primary instead of inversePrimary. Is that a bug with m3 or are we doing something wrong in this sample that could affect it?

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInSignUp.kt
IanGClifton: I couldn't see any reason for it since this whole Composable gets used only where the background is already matching surface. Maybe it was necessary in an earlier version and wasn't cleaned up previously.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/theme/Theme.kt
IanGClifton: I was originally calling the system APIs directly, but I thought using Accompanist would give 3p devs more exposure to it. Jetsurvey (before any of these changes) has a bug where the status bar content color is too light in light mode, so it's unreadable. I just fixed it in this PR since I was matching everything to the design anyway.

Do you think there's a better alternative?

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInSignUp.kt
IanGClifton: Yeah, you're right. I was originally using SmallTopAppBar but then changed it to CenterAlignedTopAppBar and missed updating my own comment :)

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInSignUp.kt
IanGClifton: Yeah, this was specified as "Jetsurvey/sys/light/on-surface" and opacity 60% in the design (same with the one below). Good call on the constant; I'll do that.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/WelcomeScreen.kt
IanGClifton: Yeah I didn't love this, but actually your idea gave me another one that the two places where this is used really only care about what the content color should be (Surface just happens to be a place that could change that), so I updated the calling sites to specifically check the LocalContentColor.current instead. WDYT?

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
IanGClifton: Yeah, I wasn't a fan of the alphas but at least having the constants makes this cleaner.

IanGClifton:
Found a few more issues that I'm working through now

IanGClifton:
It's looking much closer now! Please take a look

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/SignInScreen.kt
ricknout: Hmm that sounds like a bug in M3 to me, we can keep this as-is for now.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/WelcomeScreen.kt
ricknout: I like the new approach! Perhaps we could use `LocalContentColor.current.luminance() < 0.5f` as the default value in this cases and keep the `lightTheme` parameter from before?

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/theme/Theme.kt
ricknout: Oh I meant just the color within the lambda — its default value should suffice here (I think)

ricknout:
A few minor comments left (and some things to possibly address based on our email thread) but it's very close! Friendly reminder to please include screenshots in the main description :)

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/signinsignup/WelcomeScreen.kt
IanGClifton: Sounds good!

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/theme/Theme.kt
IanGClifton: Ah, got it; yeah I'll get rid of the lambda

Jetsurvey/app/src/main/res/values/themes.xml
```diff
@@ -17,7 +17,7 @@
 <resources>
     <!-- Base application theme. -->
```
ricknout, line 18: One last thing I noticed from the screenshots: the date picker (which I believe comes from the MDC-Android Views library) is using the M2 purple palette. We most likely need to change this XML theme parent to `Theme.Material3.DayNight.NoActionBar`

IanGClifton: Good catch! Yeah, that looks a lot better.

ricknout (approved):
Awesome! LGTM 🎉

PR #972, opened 2022-09-02
Add AnimatedContent animation for Post details

riggaroo:
* Fixed padding on popular posts
* Added Animated content onto post navigation


JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeRoute.kt
```diff
@@ -110,60 +119,68 @@ fun HomeRoute(
...
-            // This doesn't take anything more than notifying that we "interacted with the list"
-            // since that is what drives the display of the feed
-            BackHandler {
-                onInteractWithFeed()
+                // If we are just showing the detail, have a back press switch to the list.
+                // This doesn't take anything more than notifying that we "interacted with the list"
+                // since that is what drives the display of the feed
+                BackHandler {
+                    onInteractWithFeed()
+                }
```
alexvanyo, line 183: I'm wondering if the `BackHandler` should be outside of the `AnimatedContent`, so that it is introduced and removed immediately without being part of the animation.

Otherwise if the article details is animating away due to going back, I think this `BackHandler` will still be present during that animation.

riggaroo: Good point, I'll move it.

PR #970, opened 2022-09-01
Updating to safe flow collection with collectAsStateWithLifecycle()

astamato:

PR #981, opened 2022-09-26
Fix formatting not running on samples

bentrengrove:
Formatting is now done using the `scripts/format.sh` script

- Removed spotless from each sample
- Removed spotless from versions.toml

### DIFF
--- a/scripts/libs.versions.toml
+++ b/scripts/libs.versions.toml
@@ -1,3 +1,7 @@
+#####
+# This file is duplicated to individual samples from the global scripts/libs.versions.toml
+# Do not add a dependency to an individual sample, edit the global version instead.
+#####
 [versions]
 accompanist = "0.25.1"
 androidGradlePlugin = "7.2.2"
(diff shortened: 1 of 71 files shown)


scripts/libs.versions.toml
```diff
@@ -8,10 +12,11 @@ androidx-benchmark-junit4 = "1.1.0-beta04"
 androidx-constraintlayout = "1.0.1"
 androidx-corektx = "1.8.0"
 androidx-lifecycle-compose = "2.5.1"
+androidx-lifecycle-runtime-compose = "2.6.0-alpha02" # merge with above when not alpha
```
riggaroo, line 15: Is this intentional for this PR?

scripts/libs.versions.toml
```diff
@@ -8,10 +12,11 @@ androidx-benchmark-junit4 = "1.1.0-beta04"
...
 androidx-navigation = "2.5.1"
 androidx-palette = "1.0.0"
-androidx-test = "1.4.0"
-androidx-test-espresso = "3.4.0"
+androidx-test = "1.3.0"
+androidx-test-espresso = "3.3.0"
```
riggaroo, line 19: Intentional downgrade?

scripts/libs.versions.toml
bentrengrove: Yes. This PR added it but didn't add it to the global file. https://github.com/android/compose-samples/pull/970

bentrengrove:
I think this might be a bug in actions :(

It's meant to run the updated workflow but because we use a shared file, it goes and pulls the one from main instead. I'm not sure how we can work around this

scripts/libs.versions.toml
bentrengrove: I was getting annoying dependency clashes in one of the samples, but I will fix this before merging

bentrengrove: OK it seems compose 1.3 beta03 pins to these versions so we will have to go with this for now. I assume this will be upgraded again in the next release.

PR #989, opened 2022-10-04
Reply material updates and dual pane support

Gurupreet:
- Added Dual pane support
- Accompanist adaptive lib
- Updated to latest material compose beta release

Fixing issue #997

### DIFF
--- a/Reply/app/build.gradle.kts
+++ b/Reply/app/build.gradle.kts
@@ -104,7 +104,7 @@ dependencies {
     //Material 3 imports
     //implementation(libs.androidx.compose.material3)
     // downgrading to 14 for now until NavRail and Drawer are fixed
-    implementation("androidx.compose.material3:material3:1.0.0-alpha15")
+    implementation("androidx.compose.material3:material3:1.0.0-beta03")
     implementation(libs.androidx.compose.materialWindow)
     implementation(libs.androidx.compose.material.iconsExtended)
 
--- a/Reply/app/build.gradle.kts
+++ b/Reply/app/build.gradle.kts
@@ -106,6 +106,7 @@ dependencies {
     // downgrading to 14 for now until NavRail and Drawer are fixed
     implementation("androidx.compose.material3:material3:1.0.0-beta03")
     implementation(libs.androidx.compose.materialWindow)
+    implementation(libs.accompanist.adaptive)
     implementation(libs.androidx.compose.material.iconsExtended)
 
     implementation(libs.androidx.lifecycle.runtime)
--- a/Reply/app/build.gradle.kts
+++ b/Reply/app/build.gradle.kts
@@ -102,7 +102,7 @@ dependencies {
     debugImplementation(libs.androidx.compose.ui.tooling)
 
     // custom declaration for latest versions of material 3 and adaptive accompanist
-    implementation("androidx.compose.material3:material3:1.0.0-beta03")
+    implementation("androidx.compose.material3:material3:1.0.0-rc01")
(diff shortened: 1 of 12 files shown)


Reply/gradle/libs.versions.toml
```diff
@@ -1,5 +1,5 @@
 [versions]
-accompanist = "0.25.1"
+accompanist = "0.26.2-beta"
```
bentrengrove, line 2: Don't edit this file directly. If you rebase you will see a comment in there for that. It will be overriden during the next update. This file is generated from a central file that manages the versions for all samples.

You might have to manually override the version in your build.gradle similar to how you are doing it with m3, or do this on the compose-latest branch.

bentrengrove (changes requested):

Reply/gradle/libs.versions.toml
Gurupreet: Got it 👌

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
```diff
@@ -70,11 +74,22 @@ fun ReplyInboxScreen(
...
+                )
+            },
+            second = {
+                ReplyEmailDetail(
+                    email = replyHomeUIState.selectedEmail ?: replyHomeUIState.emails.first(),
+                    isFullScreen = false
+                )
+             },
+            strategy = HorizontalTwoPaneStrategy(splitFraction = 0.5f, gapWidth = 16.dp),
+            displayFeatures = calculateDisplayFeatures(activity = LocalContext.current as Activity)
```
Gurupreet, line 92: @alexvanyo is this a way to access display features inside compose or it has to be pass down from main activity ?

Reply/app/build.gradle.kts
```diff
@@ -101,10 +101,10 @@ dependencies {
     implementation(libs.androidx.compose.ui.tooling.preview)
     debugImplementation(libs.androidx.compose.ui.tooling)
 
-    //Material 3 imports
-    //implementation(libs.androidx.compose.material3)
-    // downgrading to 14 for now until NavRail and Drawer are fixed
-    implementation("androidx.compose.material3:material3:1.0.0-alpha15")
+    // custom declaration for latest versions of material 3 and adaptive accompanist
+    implementation("androidx.compose.material3:material3:1.0.0-beta03")
```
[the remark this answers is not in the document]
Gurupreet: Just updated to latest Rc 01 released today. I think for this sample I plan to keep latest/experimental versions to showcase new material releases. If main gonna be updated automatically I think that's fine then.

bentrengrove (approved):

Reply/app/build.gradle.kts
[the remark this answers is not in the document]
bentrengrove: We try to keep the main branch on stable libraries so once m3 goes stable, I would recommend working on this in the compose-latest branch which uses alphas etc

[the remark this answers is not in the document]
Gurupreet: Got it will keep latest versions on those and stable on main 💯

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
alexvanyo: I think passing it down would be best, `LocalContext.current as Activity` might work here, but could break if `LocalContext` was a context wrapper, and also makes it harder to test (it'd be easier to explicitly pass in a list)

PR #993, opened 2022-10-07
Updating Material3 to RC01

florina-muntenescu:


bentrengrove (changes requested):
Dependencies are centrally managed and then copied into each sample. You should upgrade it in the global versions file in `scripts/` or else it will be overriden the next time the samples are upgraded.

JetNews/gradle/libs.versions.toml
```diff
@@ -28,9 +28,9 @@ compileSdk = "33"
 compose = "1.3.0-rc01"
 compose-compiler = "1.3.2"
 # @pin to be updated manually - check JetChat tests
-compose-material3 = "1.0.0-beta03"
+compose-material3 = "1.0.0-rc01"
 # @pin to be updated manually - check JetChat tests
```
astamato, line 32: should we remove the disclaimer?

bentrengrove: I'm not sure who added it but I would think so

bentrengrove (approved):
Thanks for fixing that up

PR #996, opened 2022-10-10
[Jetchat] Improve UserInputText textState using rememberSaveable

ekoo:
## Step to reproduce
1. Open in large device / tablet
2. Text something in UserInputText
3. Open other apps in split screen mode and try to resize it
4. Text in UserInputText has been reset

## Expectation
Text in UserInputText survive from resizable windows

## Solution
Use rememberSaveable(stateSaver = TextFieldValue.Saver) instead remember to restore text state after configuration change


astamato (approved):
LGTM

PR #1003, opened 2022-10-21
Update Compose version

JolandaVerhoef:


Crane/gradle/libs.versions.toml
```diff
@@ -41,7 +37,7 @@ jdkDesugar = "1.1.5"
 junit = "4.13.2"
 kotlin = "1.7.20"
 maps-compose = "2.5.3"
-material = "1.7.0-beta01"
+material = "1.8.0-alpha01"
```
ricknout, line 40: I _think_ 1.8.0-alpha02 will be going out for ADS, but shouldn't be a blocker if it's later than the AndroidX releases

ricknout (approved):

Crane/gradle/libs.versions.toml
JolandaVerhoef: I'll do this later, wrote a reminder to myself

PR #1015, opened 2022-11-01
Add "Screenshots" H2 for screenshot indexing

tunjid:


tunjid:
For the screenshot to be properly indexed by the sample indexer, it needs to be under a `Schematics` or `Screenshots` H2.

PR #1017, opened 2022-11-04
[All] Upgrade to compose-latest to 1.4.0-alpha01

bentrengrove:


Crane/gradle/libs.versions.toml
```diff
@@ -27,6 +27,7 @@ coil = "2.2.0"
 # @keep
 compileSdk = "33"
 compose-compiler = "1.3.2"
+compose = "1.4.0-alpha01" # Prefer to use the BOM if you aren't on a prerelease version
```
riggaroo, line 30: @JolandaVerhoef  - would we be able to merge compose-latest if we need the bom to work on main?

JolandaVerhoef: Well, defining one compose version will not work anymore anyway as the whole point is that the version will start to differ. So it might become:

compose-ui: 1.4.0-alpha05
compose-runtime: 1.6.0-alpha01
compose-whatever: 4.1.9-rc05

My suggestion would be, don't update all samples to all latest versions every two weeks. Instead, if we want one of the samples to use a new feature, explicitly override its version.

PR #1026, opened 2022-11-17
[Jetsurvey] Simplify single/multiple choice questions

JolandaVerhoef:
There was a lot of overlap in these questions, even though the designs are very similar. I consolidated all versions and updated it according to the designs.

this CL follows up on #1024 so please merge that before this.


riggaroo (approved):
Yay, less code! Nice work 🎉

PR #1029, opened 2022-11-21
[jetsurvey] Create custom slider

JolandaVerhoef:
Screenshot 2022-11-21 at 18 30 44


Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/SliderQuestion.kt
```diff
@@ -50,60 +64,130 @@ fun SliderQuestion(
...
+            startText = stringResource(possibleAnswer.startText),
+            neutralText = stringResource(possibleAnswer.neutralText),
+            endText = stringResource(possibleAnswer.endText)
+        )
+    }
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+private fun DiscreteSlider(
```
riggaroo, line 83: This should probably take a modifier optional parameter

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/SliderQuestion.kt
```diff
@@ -50,60 +64,130 @@ fun SliderQuestion(
...
+    ticks: Int = 5,
+    onValueChangeFinished: (() -> Unit)? = null,
+    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() }
+) {
+    require(ticks >= 3) { "ticks should be >= 3" }
+    Column {
+        Slider(
+            value = value.toFloat(),
+            onValueChange = { onValueChange(it.roundToInt()) },
+            modifier = modifier.padding(horizontal = 21.dp),
```
riggaroo, line 99: Why 21dp? Its not a multiple of 8 :)

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/SliderQuestion.kt
```diff
@@ -50,60 +64,130 @@ fun SliderQuestion(
...
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+private fun DiscreteSlider(
+    value: Int,
+    onValueChange: (Int) -> Unit,
+    startText: String,
+    neutralText: String,
+    endText: String,
+    modifier: Modifier = Modifier,
+    ticks: Int = 5,
```
riggaroo, line 90: Why 5 as default? I dont know if it should have a default value here

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/SliderQuestion.kt
```diff
@@ -50,60 +64,130 @@ fun SliderQuestion(
...
+        Row(
+            horizontalArrangement = Arrangement.SpaceBetween,
+            modifier = Modifier.fillMaxWidth()
+        ) {
+            ProvideTextStyle(
+                MaterialTheme.typography.labelMedium.copy(textAlign = TextAlign.Center)
+            ) {
+                Text(
+                    text = startText,
+                    modifier = Modifier.widthIn(max = 60.dp)
```
riggaroo, line 116: I think this should be a percentage of the width, ie number of ticks, divided by overall size

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/SliderQuestion.kt
```diff
@@ -50,60 +64,130 @@ fun SliderQuestion(
...
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+private fun DiscreteSlider(
+    value: Int,
+    onValueChange: (Int) -> Unit,
+    startText: String,
+    neutralText: String,
+    endText: String,
```
riggaroo, line 88: What about passing in a list of options with their labels, and then returning that option as the onValueChanged? If I configured this to have 8 ticks instead of 3, how would this be spaced out?

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/SliderQuestion.kt
```diff
@@ -114,3 +198,24 @@ fun SliderQuestionPreview() {
...
 }
+
+@Preview(name = "default")
+@Preview(name = "large font", fontScale = 2f)
+@Preview(name = "dark", uiMode = UI_MODE_NIGHT_YES)
+@Preview(name = "wide", widthDp = 600)
+@Preview(name = "small", widthDp = 300)
+@Composable
+private fun DiscreteSliderPreview() {
+    var sliderValue by remember { mutableStateOf(3) }
```
riggaroo, line 209: Maybe add another preview slider with more ticks and more text :)

riggaroo (approved):
Looking good so far :) Can you update the slider to take in a list of ordered options though instead of ticks and separate text? I think it'd be more reusable. 
I'm approving to unblock you though, as its cool as is :)

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/SliderQuestion.kt
simona-anomis: I guess this also depends on whether Jolo decides to go with a customizable slider or hardcode the choice amount to 3

simona-anomis (approved):

PR #1035, opened 2022-11-25
[JetSurvey] Add a group question

angiesasmita:

### DIFF
--- a/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/Survey.kt
+++ b/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/Survey.kt
@@ -75,7 +75,7 @@ sealed class PossibleAnswer {
     data class Group(
         @ArrayRes val items: Int,
         @ArrayRes val groups: Int
-    ): PossibleAnswer()
+    ) : PossibleAnswer()
 }
 
 sealed class Answer<T : PossibleAnswer> {
@@ -86,7 +86,7 @@ sealed class Answer<T : PossibleAnswer> {
 
     data class Action(val result: SurveyActionResult) : Answer<PossibleAnswer.Action>()
     data class Slider(val answerValue: Float) : Answer<PossibleAnswer.Slider>()
-    data class Group(val answerGrouping: Map<String, List<Badge>>): Answer<PossibleAnswer.Group>()
+    data class Group(val answerGrouping: Map<String, List<Badge>>) : Answer<PossibleAnswer.Group>()
 }
 
 /**
--- a/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/Survey.kt
+++ b/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/Survey.kt
@@ -17,8 +17,10 @@
 package com.example.compose.jetsurvey.survey
 
 import android.net.Uri
+import androidx.annotation.ArrayRes
 import androidx.annotation.DrawableRes
 import androidx.annotation.StringRes
+import com.example.compose.jetsurvey.survey.question.Badge
 
 data class SurveyResult(
     val library: String,
@@ -69,6 +71,11 @@ sealed class PossibleAnswer {
         @StringRes val neutralText: Int,
         val defaultValue: Float = 5.5f
     ) : PossibleAnswer()
+
+    data class Group(
+        @ArrayRes val items: Int,
+        @ArrayRes val groups: Int
+    ): PossibleAnswer()
 }
 
 sealed class Answer<T : PossibleAnswer> {
@@ -79,6 +86,7 @@ sealed class Answer<T : PossibleAnswer> {
 
     data class Action(val result: SurveyActionResult) : Answer<PossibleAnswer.Action>()
     data class Slider(val answerValue: Float) : Answer<PossibleAnswer.Slider>()
+    data class Group(val answerGrouping: Map<String, List<Badge>>): Answer<PossibleAnswer.Group>()
 }
 
 /**
--- a/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/Survey.kt
+++ b/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/Survey.kt
@@ -72,7 +72,10 @@ sealed class PossibleAnswer {
         val defaultValue: Float = 5.5f
     ) : PossibleAnswer()
 
-    data class Group(
+    /**
+     * Requires user to completely divide the [items] up between the various distinct [groups].
+     */
+    data class Grouping(
         @ArrayRes val items: Int,
         @ArrayRes val groups: Int
     ) : PossibleAnswer()
@@ -86,7 +89,7 @@ sealed class Answer<T : PossibleAnswer> {
 
     data class Action(val result: SurveyActionResult) : Answer<PossibleAnswer.Action>()
     data class Slider(val answerValue: Float) : Answer<PossibleAnswer.Slider>()
-    data class Group(val answerGrouping: Map<String, List<Badge>>) : Answer<PossibleAnswer.Group>()
+    data class Grouping(val answerGrouping: Map<String, List<Badge>>) : Answer<PossibleAnswer.Grouping>()
 }
 
 /**
--- a/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/GroupingQuestion.kt
+++ b/Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/GroupingQuestion.kt
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.example.compose.jetsurvey.survey.question
 
 import androidx.compose.foundation.BorderStroke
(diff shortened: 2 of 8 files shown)


Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/Survey.kt
```diff
@@ -69,6 +71,11 @@ sealed class PossibleAnswer {
         @StringRes val neutralText: Int,
         val defaultValue: Float = 5.5f
     ) : PossibleAnswer()
+
+    data class Group(
```
JolandaVerhoef, line 75: Consider calling this "Grouping" to be clearer that this is not indicating a single group. 

Also, this might be a good place to add a comment explaining what this type of question is, something like: "Asks user to divide the [items] up between the various [groups]."

I know the other PossibleAnswer types don't have documentation yet, but this might set a good trend :)

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/GroupingQuestion.kt
```diff
@@ -0,0 +1,281 @@
...
+    val initialBadgeList =
+        stringArrayResource(possibleAnswer.items).mapIndexed { index, item ->
+            Badge(index, item)
+        }
+    val badgeGroupList =
+        stringArrayResource(possibleAnswer.groups)
+            // Make sure no group name uses reserved name
+            .filter { it != STARTING_ZONE_BADGE_GROUP }
+            .distinct()
+    var groupAnswer by remember {
```
JolandaVerhoef, line 71: Consider using `rememberSaveable` so user doesn't lose their progress on configuration changes.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/GroupingQuestion.kt
```diff
@@ -0,0 +1,281 @@
...
+                width = 1.dp,
+                color = MaterialTheme.colorScheme.outline
+            ),
+        ) {
+            Text(
+                text = badge.text,
+                modifier = Modifier.padding(8.dp)
+            )
+        }
+    }
```
JolandaVerhoef, line 280: Please add previews for this question.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/components/Draggable.kt
```diff
@@ -0,0 +1,187 @@
...
+    val state = LocalDragTargetInfo.current
+    val dragPosition = state.dragPosition
+    val dragOffset = state.dragOffset
+    var isItemInTheDropZone by remember { mutableStateOf(false) }
+
+    val hasDragEndInDropZone by remember {
+        derivedStateOf { isItemInTheDropZone && state.hasDragEnd }
+    }
+    if (hasDragEndInDropZone) {
+        state.hasDragEndInAnyDropZone = true
```
riggaroo, line 168: Our advice is to never write to state in a composition as it can cause unintended side effects https://developer.android.com/jetpack/compose/performance/bestpractices#avoid-backwards

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/components/OverflowContainer.kt
```diff
@@ -0,0 +1,70 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.layout.Layout
+import androidx.compose.ui.unit.IntOffset
+import androidx.compose.ui.unit.LayoutDirection
+
+/**
+ * Custom layout which allow content to be placed linearly and wraps when overflow
+ */
+@Composable
+fun OverflowContainer(
```
riggaroo, line 30: What about using a FlowLayout from accompanist https://google.github.io/accompanist/flowlayout/ ?

Jetsurvey/app/src/main/res/values/strings.xml
```diff
@@ -82,6 +83,23 @@
     <string name="back_to_future">Back to the future</string>
     <string name="outbreak">Outbreak</string>
 
+    <!-- question 7-->
+    <string name="group_food">Group the following food items in terms of frequency of consumption?</string>
```
riggaroo, line 87:
```suggestion
    <string name="group_food">Group the following food items in terms of frequency of consumption:</string>
```

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/components/Draggable.kt
```diff
@@ -0,0 +1,187 @@
...
+    // Check if any item dropped and this is the item
+    if (state.hasDragEndInAnyDropZone && hasItemDraggingEnd) {
+        hasItemDraggingEnd = false
+        state.hasDragEndInAnyDropZone = false
+        onItemDropped(item)
+    }
+
+    if (isItemDragging) {
+        state.dataToTransfer = item
+    }
```
riggaroo, line 114: Same thing here - I dont believe we should be writing to state in the composition

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/components/Draggable.kt
```diff
@@ -0,0 +1,187 @@
...
+import androidx.compose.ui.unit.IntSize
+import androidx.compose.ui.unit.center
+
+class DragTargetInfo<T> {
+    var isDragging: Boolean by mutableStateOf(false)
+    var hasDragEnd: Boolean by mutableStateOf(false)
+    var hasDragEndInAnyDropZone: Boolean by mutableStateOf(false)
+    var dragPosition by mutableStateOf(Offset.Zero)
+    var dragOffset by mutableStateOf(Offset.Zero)
+    var draggableComposable by mutableStateOf<(@Composable (Boolean) -> Unit)?>(null)
```
riggaroo, line 47: We should confirm if this is something we should be advising people to do - haven't seen storing of a composable in a mutableStateOf before. Wondering if this isn't a use case for movableContentOf instead..

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/GroupingQuestion.kt
```diff
@@ -0,0 +1,281 @@
...
+
+    Column(modifier) {
+        Text(
+            text = badgeGroup,
+            style = MaterialTheme.typography.headlineSmall,
+            modifier = Modifier.padding(bottom = 8.dp)
+        )
+        DropZone { isDraggingInDropZone, hasDragEndInDropZone ->
+            if (hasDragEndInDropZone) {
```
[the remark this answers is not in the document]
riggaroo: (see the link for calling state changes inside the composition, i believe the same thing applies here)

riggaroo (changes requested):
This looks great so far! Really nice example :) I have a couple of questions around the setting of state values and calling callbacks within the composition of items - I don't think its advised. As well as if this is not a case for movableContent - I don't have all the answers myself :D so might be worth asking around

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/GroupingQuestion.kt
```diff
@@ -0,0 +1,281 @@
...
+    var badgeList by remember {
+        mutableStateOf(
+            groupAnswer[badgeGroup] ?: emptyList()
+        )
+    }
+    val onItemDropped: (Badge) -> Unit = {
+        badgeList = badgeList.minus(it)
+    }
+
+    val state = LocalDragTargetInfo.current
```
bentrengrove, line 149: I think you could wrap a lot of this code up in a state holder, it would remove the need for the callbacks and the composition local I believe. It looks like you are using Composition as a way to call the callback which isn't quite right, you should have your state hoisted high enough that when it's changed the relevant components recompose.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/components/Draggable.kt
bentrengrove, line 114: Replace this with `val updatingItem = rememberUpdatedState(item)`

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/components/Draggable.kt
```diff
@@ -0,0 +1,187 @@
...
+
+    Box(
+        modifier = modifier
+            .onGloballyPositioned {
+                currentPosition = it.positionInWindow()
+            }
+            .pointerInput(Unit) {
+                detectDragGestures(
+                    onDragStart = {
+                        isItemDragging = true
```
bentrengrove, line 124: You can set your drag state here directly with the new updating state, this will remove the need for the recomposition with isItemDragging in order to set that state.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/components/Draggable.kt
```diff
@@ -0,0 +1,187 @@
...
+}
+
+val LocalDragTargetInfo = compositionLocalOf { DragTargetInfo<Any?>() }
+
+@Composable
+fun Draggable(
+    modifier: Modifier = Modifier,
+    content: @Composable BoxScope.() -> Unit
+) {
+    val state = remember { DragTargetInfo<Any?>() }
```
bentrengrove, line 58: A stronger type here rather than `Any?` would be ideal

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/Survey.kt
angiesasmita: Thanks :) Added in the next commit

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/GroupingQuestion.kt
angiesasmita: Added in the next commit!

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/question/components/Draggable.kt
```diff
@@ -0,0 +1,187 @@
...
+    var dragPosition by mutableStateOf(Offset.Zero)
+    var dragOffset by mutableStateOf(Offset.Zero)
+    var draggableComposable by mutableStateOf<(@Composable (Boolean) -> Unit)?>(null)
+    var dataToTransfer by mutableStateOf<T?>(null)
+}
+
+val LocalDragTargetInfo = compositionLocalOf { DragTargetInfo<Any?>() }
+
+@Composable
+fun Draggable(
```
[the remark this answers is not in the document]
angiesasmita: Changed in the next commit.

Jetsurvey/app/src/main/res/values/strings.xml
angiesasmita: Done

PR #1044, opened 2022-12-12
[Crane] Location input state is restored on process/activity recreation

astamato:
Fixing "Choose Destination" state restoration.

image


arriolac (approved):
LGTM

PR #1053, opened 2022-12-26
[Jetsurvey] Added QuestionDescription, simplified title usage

florina-muntenescu:


IanGClifton:
I think this is pretty similar to https://github.com/android/compose-samples/pull/1058/files#diff-78105390f84dd782b10e7ba24251542b83b39913d3d5635ce269bbcadca3561b, though I kept the spacer before the title instead of putting it into that composable. I don't think I have a strong opinion on either direction. I'll reach out via email since you should be OOO haha

PR #1058, opened 2022-12-27
[Jetsurvey] Simplification of survey screens

### LINKED ISSUE #1054: [Jetsurvey]: Tapping the survey title closes the survey

### Sample app

Jetsurvey

### What happened?

Tapping the title (e.g., "1 of 6") will close the survey; this is because the X has a modifier to make it take the whole width. The top bar should be using CenterAlignedTopAppBar and then the ButtonIcon/Icon should not use the full width.

### Relevant logcat output

_No response_

### Code of Conduct

IanGClifton:
This eliminates the survey repository and related data models in favor of simplifying the code base. Each of the different question types has a generic composable such as MultipleChoiceQuestion, which takes all of the relevant strings/resources to generate the UI. The actual survey data is baked into a specific composable in Survey.kt, such as FreeTimeQuestion, instead of coming from a data model.

The SurveyViewModel is now much more explicit with methods for each question such as onFreeTimeResponse. In a production app, you'd likely want this to be more flexible to handle server-side questions, but this code pattern makes the sample easier to skim. Every response is maintained in SurveyViewModel as a separate state. This increases verbosity in favor of better readability.

This also appears to fix the issue of photos not working on API 21, though I'm not sure what specific change fixed it. I verified on a Nexus 4 running Android 5.1.1.

Note: This does not animate between the final survey question and the survey results. When the work to move the app to Compose Navigation is done, the results will be a separate destination that is navigated to and animated that way.

This fixes #1054.


Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyFragment.kt
```diff
@@ -62,79 +73,151 @@ class SurveyFragment : Fragment() {
...
+                            ) { targetState ->
+
+                                when (targetState.surveyQuestion) {
+                                    SurveyQuestion.FREE_TIME -> {
+                                        FreeTimeQuestion(
+                                            modifier,
+                                            viewModel.freeTimeResponse,
+                                        ) { selected, answer ->
+                                            viewModel.onFreeTimeResponse(selected, answer)
+                                        }
```
JolandaVerhoef, line 145: Simplify by passing a method reference:
```
FreeTimeQuestion(
  modifier,
  viewModel.freeTimeResponse,
  viewModel::onFreeTimeResponse
)
```

Same goes for the other question types.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyFragment.kt
```diff
@@ -62,79 +73,151 @@ class SurveyFragment : Fragment() {
...
+                                    )
                                 }
-                            )
-                            is SurveyState.Result -> SurveyResultScreen(
-                                result = targetState,
-                                onDonePressed = {
-                                    activity?.onBackPressedDispatcher?.onBackPressed()
+                            ) { targetState ->
+
+                                when (targetState.surveyQuestion) {
```
JolandaVerhoef, line 138: The level of nesting here is quite extreme and makes it difficult to follow along. Would it be possible to extract a part of this method?

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyFragment.kt
```diff
@@ -62,79 +73,151 @@ class SurveyFragment : Fragment() {
...
+                    val isSurveyComplete by viewModel.isSurveyComplete.observeAsState(false)
+
+                    if (isSurveyComplete) {
+                        // Note, results are hardcoded for this demo; in a complete app, you'd
+                        // likely send the survey responses to a backend to determine the
+                        // result and then pass them here.
+                        SurveyResultScreen(
+                            title = stringResource(R.string.survey_result_title),
+                            subtitle = stringResource(R.string.survey_result_subtitle),
+                            description = stringResource(R.string.survey_result_description),
```
JolandaVerhoef, line 88: What is the reasoning behind passing these string resources and making them dynamic? They're not backed by any data, and this is the only calling spot. For simplicity, I'd consider just hard-coding them in `SurveyResultScreen`.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyViewModel.kt
```diff
@@ -17,125 +17,156 @@
...
+
+    private val _feelingAboutSelfiesResponse = mutableStateOf<Float?>(null)
+    val feelingAboutSelfiesResponse: Float?
+        get() = _feelingAboutSelfiesResponse.value
+
+    private val _selfieUriResponse = mutableStateOf<Uri?>(null)
+    val selfieUriResponse: Uri?
+        get() = _selfieUriResponse.value
+
+    // ----- Survey status exposed as LiveData -----
```
JolandaVerhoef, line 69: I'm surprised to see a `LiveData` here. Is there a valid reason to use `LiveData` vs a Compose `State`? If so, consider adding a larger comment explaining why you're using the two different versions in this sample.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
```diff
@@ -19,216 +19,74 @@ package com.example.compose.jetsurvey.survey
...
-        }
-        Spacer(modifier = Modifier.height(8.dp))
-        OutlinedButton(onClick = onDoNotAskForPermissions) {
-            Text(stringResource(R.string.do_not_ask_permissions))
-        }
-    }
-}
 
+/**
+ * Creates a Column that is vertically scrollable and contains the title, the directions (if a
```
JolandaVerhoef, line 38: Don't use "create" as a verb, as a composable does not _create_ a Column, it _defines_ or _is_ a column. Also, not sure if this comment is too specific. What about:

```
A scrollable container with the question's title, direction, and dynamic content.
```

Then add lines for each parameter, explaining that direction will be omitted when null.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
```diff
@@ -19,216 +19,74 @@ package com.example.compose.jetsurvey.survey
...
+ * string resource is passed), and the content
+ */
 @Composable
-private fun QuestionContent(
-    question: Question,
-    answer: Answer<*>?,
-    onAnswer: (Answer<*>) -> Unit,
-    onAction: (Int, SurveyActionType) -> Unit,
-    modifier: Modifier = Modifier
+fun QuestionWrapper(
```
JolandaVerhoef, line 42: Rename file to be `QuestionWrapper`

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
```diff
@@ -19,216 +19,74 @@ package com.example.compose.jetsurvey.survey
...
+        Spacer(Modifier.height(18.dp))
+
+        content()
     }
 }
 
 @Composable
-private fun QuestionTitle(@StringRes title: Int) {
-    Row(
+fun QuestionTitle(@StringRes title: Int) {
```
JolandaVerhoef, line 66: Make composables private that are only extracted for readability and not reusability. Goes for the whole project.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
```diff
@@ -19,216 +19,74 @@ package com.example.compose.jetsurvey.survey
...
-        Text(
-            text = stringResource(id = title),
-            style = MaterialTheme.typography.titleMedium,
-            color = MaterialTheme.colorScheme.onSurface.copy(alpha = slightlyDeemphasizedAlpha),
-            modifier = Modifier
-                .fillMaxWidth()
-                .padding(vertical = 24.dp, horizontal = 16.dp)
-        )
-    }
+            .padding(vertical = 24.dp, horizontal = 16.dp)
```
JolandaVerhoef, line 77: Consider passing this padding from the parent. Conceptually, a QuestionTitle should not know where it's placed in its parent and thus not care about this padding. Having all these paddings defined in the parent makes it easier to reason about how all children relate to each other.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
JolandaVerhoef, line 42: I'm wondering if this QuestionWrapper is really necessary, or if it's premature optimization and makes our survey less flexible. For example, what if I want to show a scrollable grid in one of the questions (or a lazy column for that matter), it would not work because there's already a scrollable column defined here.

What about making a reusable `QuestionTitle` and `QuestionDirections`. You can also save the modifier chains and reuse them, e.g. `questionTitleModifier`.

JolandaVerhoef (changes requested):

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/Survey.kt
```diff
@@ -17,83 +17,102 @@
...
- */
-enum class SurveyActionType { PICK_DATE, TAKE_PHOTO, SELECT_CONTACT }
-
-sealed class SurveyActionResult {
-    data class Date(val dateMillis: Long) : SurveyActionResult()
-    data class Photo(val uri: Uri) : SurveyActionResult()
-    data class Contact(val contact: String) : SurveyActionResult()
+@Composable
+fun FreeTimeQuestion(
+    modifier: Modifier = Modifier,
```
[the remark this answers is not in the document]
IanGClifton: I think there was only one where I had done this right, but I believe they're all fixed now!

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyFragment.kt
IanGClifton: Good call! Updated!

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyFragment.kt
IanGClifton: I was originally going to have them come from the ViewModel and pass them through the Composables, but it didn't really make sense given they were still hardcoded. I like your idea of hardcoding them in SurveyResultScreen instead, so I've done that.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyViewModel.kt
IanGClifton: I was originally thinking it would be good to show that either can be used, but now I think that it's just more confusing that way. I'm not sure a comment would really make that more clear, so I switched them to State also.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
IanGClifton: Updated!

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
IanGClifton: Great suggestion, done!

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
IanGClifton: Updated `QuestionTitle` and `QuestionDirections` to be private.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
IanGClifton: I played around with moving this to the parent, but it ends up a little odd. The problem is that the title has a background color that's padded and then the text goes in that. Mentally, I map that background + text as the `QuestionTitle` so the padding is part of that. Another option would be the title knows about the background but the padding is passed in, but that feels like it means the parent knows too much about what the child looks like. I started to think about it from the perspective of "if QuestionTitle is shown elsewhere, should it be with the same background and padding" and I _think_ the answer is yes, so it makes sense to have those defined in the QuestionTitle. What do you think?

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/SurveyQuestions.kt
IanGClifton: That's actually how I originally did it (which is why QuestionTitle and QuestionDirections weren't private). Since Column is simple, I didn't mind having that portion repeated in each question, but then I realized I needed it to be a scrollable Column. Repeating that modifier (even if extracted) made it feel like I was requiring each question to build the UI with the same building blocks in the same way. Each time I was updating a question, I was basically just copying and pasting the same chunks from another question.

Maybe if we have cases where there's another layout like a grid, we'd rename this to something like ScrollableColumnQuestion and the new one might be VerticalGridQuestion or something like that? In that case, the QuestionTitle and QuestionDirections composables might need to be different anyway.

As a side note, the reason I went with a scrollable Column vs a LazyColumn was because the majority of each question is visible on most phones, so the lazy behavior wasn't a performance win. If we had a question with a dozen potential answers, then a LazyColumn would make more sense for that question, but then it gets back to whether this should be in a wrapper or in the individual questions or something else.

JolandaVerhoef (approved):

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/survey/QuestionWrapper.kt
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright 2020 The Android Open Source Project
```
JolandaVerhoef, line 2: nit: update your copyright year :)

IanGClifton: Don't mind me just doing a Google search for what year it is....

PR #1059, opened 2022-12-28
[Jetsurvey] Changed to use Navigation Compose

### LINKED ISSUE #1038: [JetSurvey]: Migrate to Compose navigation

### Is this a feature request for one of the samples?

- [X] Yes, this is a specific request related to this samples repo.

### Sample app

Jetsurvey

### Describe the problem

It is not Jetpack compose friendly, please migrate it to 100% Jetpack compose, without any XML file or fragment.

### Describe the solution

Use Jetpack Compose with Navigation Compose

### Additional context

_No response_

### Code of Conduct

IanGClifton:
This eliminates all fragments in favor of navigating at the composable
level. I also updated the screen-level composables to take in explicit
lambdas rather than event handlers to follow best practices.
    
One notable change here is that there is no longer a fragment from which
to launch the camera. Instead, the PhotoQuestion uses
rememberLauncherForActivityResult and selfies are now stored in the
app's cache directory, which means no file permission is required.

Fixes #1038


IanGClifton:
This depends on #1058 landing first but fixes #1038 when landed.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/Navigation.kt
```diff
@@ -16,28 +16,124 @@
...
+) {
+    NavHost(
+        navController = navController,
+        startDestination = WELCOME_ROUTE,
+    ) {
+        composable(WELCOME_ROUTE) {
+            val welcomeViewModel: WelcomeViewModel = viewModel(factory = WelcomeViewModelFactory())
+            WelcomeScreen(
+                onSignInSignUp = { email ->
+                    welcomeViewModel.handleContinue(
```
arriolac, line 66: Should this logic should be handled within `WelcomeScreen` and `JetsurveyNavHost`? This way, `JetsurveyNavHost` is only concerned with navigation events (i.e. once sign in/up completes and ViewModels would only be scoped within individual screens. Same comment for other screens.

arriolac (commented):
How satisfying was it to remove all those Fragments?? 😛 

I just had one comment about scoping the ViewModel within individual screens so that the NavHost doesn't need to concern itself with that logic.

Jetsurvey/app/src/main/java/com/example/compose/jetsurvey/Navigation.kt
IanGClifton: That makes sense. I'll create a route composable for each route that has the logic for getting the view model and any setup needed for the individual screen.

arriolac (approved):
LGTM

PR #1074, opened 2023-01-26
Update NetworkImage.kt to support compose preview

pedrosax:
Fixed reference to placeholder image so that the compose preview in UI screens (eg Onboarding.kt, FeaturedCourses.kt, CourseDetails.kt) display a preview that indicates that it is working.


bentrengrove:
Looks like it just needs formatting. `./scripts/format.sh` to auto fix

bentrengrove (approved):

PR #1082, opened 2023-02-14
Move all plugins to version catalog

zsmb13:
Updates the build files to use version catalog declarations for all plugin usages.

(Small bonus: moves Jetcaster's Room usage from kapt to KSP.)


riggaroo (approved):
Thanks Marton!

PR #1083, opened 2023-02-17
Remove unused build configuration from project level build files

zsmb13:


scripts/libs.versions.toml
```diff
@@ -27,7 +27,6 @@ coil = "2.2.0"
 # @keep
 compileSdk = "33"
 compose-compiler = "1.4.0"
-compose-snapshot = "-"
```
zsmb13, line 30: I believe this is currently unused and can be removed, but let me know if it serves a purpose.

riggaroo (approved):

PR #1086, opened 2023-02-28
[Jetchat] Fixing top bar overlap

astamato:
Fixing top bar in Jetchat's Profile screen where half of the top bar is not properly visible.
Bug introduced [here](https://github.com/android/compose-samples/compare/77fc82e64ddefaaf6481aaecfaf0e9ea9bc65199..07b0a268889ff09c2872e7efe3f7d9e1f7330e7a) with the change in `JetchatAppBar`.

before | after

image

### DIFF
--- a/Jetchat/app/src/main/java/com/example/compose/jetchat/components/JetchatAppBar.kt
+++ b/Jetchat/app/src/main/java/com/example/compose/jetchat/components/JetchatAppBar.kt
@@ -19,6 +19,7 @@
 package com.example.compose.jetchat.components
 
 import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.RowScope
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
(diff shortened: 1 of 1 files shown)


Jetchat/app/src/main/java/com/example/compose/jetchat/components/JetchatAppBar.kt
```diff
@@ -43,21 +44,23 @@ fun JetchatAppBar(
...
-                modifier = Modifier
-                    .size(64.dp)
-                    .clickable(onClick = onNavIconPressed)
-                    .padding(16.dp)
-            )
-        }
-    )
+    Box {
+        CenterAlignedTopAppBar(
+            modifier = modifier,
```
bentrengrove, line 49: Should this modifier be on the parent Box now?

bentrengrove (approved):
Seems a bit weird that this is needed but lgtm

Jetchat/app/src/main/java/com/example/compose/jetchat/components/JetchatAppBar.kt
astamato: Yes this solution comes across as a hack. @JolandaVerhoef will try adding a `Scaffold` to the app which should be the correct way to solve the overlap issue.

Issue #1109, opened 2023-04-06
[Bug]: samples won't work in Android Studio Flamingo because `kotlinOptions.jvmTarget` isn't set


obask:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### Is this an issue related to one of the samples?

- [X] Yes, this is a specific issue related to this samples repo.

### Sample app

JetNews

### What happened?

When upgrading any sample to kotlin 1.8.0 or AGP 8.0.0:
```
Execution failed for task ':app:compileDebugKotlin'.
> 'compileDebugJavaWithJavac' task (current target is 1.8) and 'compileDebugKotlin' task (current target is 17) jvm target compatibility should be set to the same Java version.
  Consider using JVM toolchain: https://kotl.in/gradle/jvm/toolchain
```

This could be solved by adding this to `app/build.gradle.kts`:
```
    kotlinOptions {
        jvmTarget = '1.8'
    }
```
Or by `changing compileOptions.targetCompatibility` to `JavaVersion.VERSION_17`

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

astamato:
CC https://github.com/android/compose-samples/pull/1114

noufil:
'compileProdReleaseJavaWithJavac' task (current target is 1.8) and 'kaptGenerateStubsProdReleaseKotlin' task (current target is 17) jvm target compatibility should be set to the same Java version.
 
 
 Nothing is working for me.

noufil:
kotlin{
IjvmToolchain(18)
}

above code did not help

obask:
You just need to have the same version of kotlin JDK and Java target JDK. The exact fix may vary depending on which gradle task you are changing.
Here is an example https://github.com/android/compose-samples/pull/1110

noufil:
this answer is useless I tried all the way nothing is working

Caused by: org.gradle.api.GradleException: 'compileProdReleaseJavaWithJavac' task (current target is 1.7) and 'kaptGenerateStubsProdReleaseKotlin' task (current target is 17) jvm target compatibility should be set to the same Java version.

now both versions and error is also saying the same yet error exist.

argenkiwi:
> I tried all the way nothing is working
> 
[...]

PR #1110, opened 2023-04-07
Set kotlinOptions.jvmTarget to 1.8 #1109

obask:
...to make samples compatible with AGP 8.0.0
and Android Studio Flamingo

### DIFF
--- a/Crane/app/build.gradle.kts
+++ b/Crane/app/build.gradle.kts
@@ -83,6 +83,10 @@ android {
         targetCompatibility = JavaVersion.VERSION_1_8
     }
 
+    kotlinOptions {
+        jvmTarget = "1.8"
+    }
+
     buildFeatures {
         compose = true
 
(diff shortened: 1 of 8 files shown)


obask:
related issue: https://issuetracker.google.com/issues/277272832

astamato:
As per latest main, samples are already compatible with AGP 8.0.0, why do we need this? cc @mlykotom

obask:
No it's not necessary, because samples have been updated to JDK17 which is default for AGP 8.0 and kotlin 1.8.0

obask:
There are some minor inconsistencies, but these aren't important:
Jetsurvey is setting java 17 explicitly while other samples rely on the default value:
https://github.com/android/compose-samples/blob/main/Jetsurvey/app/build.gradle.kts#L52
```
    kotlinOptions {
        jvmTarget = JavaVersion.VERSION_17.toString()
    }
```

Issue #1122, opened 2023-04-30
[Reply]: Ripple not clipped when tapping Email list item


JolandaVerhoef:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### Is this an issue related to one of the samples?

- [X] Yes, this is a specific issue related to this samples repo.

### Sample app

Reply

### What happened?

See screenshot. The `Card` should clip the ripple when the user taps it.

<img width="367" alt="Screenshot 2023-04-30 at 12 39 42" src="https://user-images.githubusercontent.com/6952116/235365220-9fecc4d7-65ef-4e37-90d2-42708bb76741.png">

This would be easy to solve by using the `Card` override that has an `onClick` parameter. However, I'm also working on adding long-press support to the card, which is not easy to combine with that version of `Card`. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1123, opened 2023-04-30
[Reply] Allow multi-select of emails and highlight the opened email

JolandaVerhoef:
Allow multi-select of emails, by:
- tapping the avatar
- long-pressing the card

A selected email (during multi-select) is highlighted with a different background and a checkmark instead of the Avatar. The opened email is highlighted with a different container color.

This PR also fixes issue #1122 (ripple not clipped)

Screenshot_20230430_135836

### DIFF
--- a/Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
+++ b/Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
@@ -58,161 +58,163 @@ import com.google.accompanist.adaptive.TwoPane
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun ReplyInboxScreen(
-  contentType: ReplyContentType,
-  replyHomeUIState: ReplyHomeUIState,
-  navigationType: ReplyNavigationType,
-  displayFeatures: List<DisplayFeature>,
-  closeDetailScreen: () -> Unit,
-  navigateToDetail: (Long, ReplyContentType) -> Unit,
-  modifier: Modifier = Modifier
+    contentType: ReplyContentType,
+    replyHomeUIState: ReplyHomeUIState,
+    navigationType: ReplyNavigationType,
+    displayFeatures: List<DisplayFeature>,
+    closeDetailScreen: () -> Unit,
+    navigateToDetail: (Long, ReplyContentType) -> Unit,
+    modifier: Modifier = Modifier
 ) {
-  /**
-   * When moving from LIST_AND_DETAIL page to LIST page clear the selection and user should see LIST screen.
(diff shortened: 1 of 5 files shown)


Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
```diff
@@ -69,20 +75,31 @@ fun ReplyInboxScreen(
         }
     }
 
+    var selectedEmailIds by remember { mutableStateOf(setOf<Long>()) }
+    val swapEmailSelection = { id: Long ->
+        if (selectedEmailIds.contains(id))
+            selectedEmailIds -= id
+        else selectedEmailIds += id
+    }
```
riggaroo, line 83: Could this form part of `ReplyHomeUIState`?

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
```diff
@@ -69,20 +75,31 @@ fun ReplyInboxScreen(
         }
     }
 
+    var selectedEmailIds by remember { mutableStateOf(setOf<Long>()) }
+    val swapEmailSelection = { id: Long ->
```
Gurupreet, line 79: Does the swap mean user would be able to swap/drag to order the emails?

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
```diff
@@ -69,20 +75,31 @@ fun ReplyInboxScreen(
         }
     }
 
+    var selectedEmailIds by remember { mutableStateOf(setOf<Long>()) }
```
astamato, line 78: sounds like you would like to remembersave this

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
JolandaVerhoef: Done!

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
JolandaVerhoef: No, sorry that was a naming issue. I renamed to "toggleEmailSelection"

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
JolandaVerhoef: Done

astamato (approved):
Nice one!

PR #1124, opened 2023-05-04
Fix FilterScreen Preview

parishsu:
Wrap a Theme so that the Preview render correctly


parishsu (commented):
Fix format violations

parishsu (commented):
Fix formating again

PR #1126, opened 2023-05-06
[Reply] Using navigation 2.6.0-alpha06, removed unused dependency on …

florina-muntenescu:
…material, using stable version of benchmark junit


JolandaVerhoef:
Adapting versions in 1 toml file will not work in the future, as we use a common versions file, which is automatically copied to all samples. Instead, we should adapt the top level versions file and run the script.

florina-muntenescu:
> Adapting versions in 1 toml file will not work in the future, as we use a common versions file, which is automatically copied to all samples. Instead, we should adapt the top level versions file and run the script.

Addressed

JolandaVerhoef (approved):

PR #1136, opened 2023-05-13
[RefactorOnly] Removing unnecessary annotations and imports

idiotleon:
Removing unnecessary annotations and imports.
Correcting some typos of the comment.
No logic change at all.

Sorry, not trying to be picky :).


florina-muntenescu:
Thanks for the PR. Would you be able to finish it? Ensure that spotless is passing and make this ready for review?

PR #1140, opened 2023-06-02
[Jetchat] Add recording audio button

JolandaVerhoef:
See video.

Demonstrating:
- Cancel on horizontal swipe
- Only canceling when not too far up
- Tooltip when tapping the icon

https://github.com/android/compose-samples/assets/6952116/98c001f8-ec0f-4222-a953-528d531cb799

### DIFF
--- a/Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
+++ b/Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
@@ -53,8 +53,8 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import com.example.compose.jetchat.R
-import kotlinx.coroutines.launch
 import kotlin.math.abs
+import kotlinx.coroutines.launch
 
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
--- a/Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
+++ b/Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
@@ -1,0 +1,164 @@
+/*
+ * Copyright 2023 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.example.compose.jetchat.conversation
+
+import androidx.compose.animation.animateColor
+import androidx.compose.animation.core.Spring
+import androidx.compose.animation.core.animateFloat
+import androidx.compose.animation.core.spring
+import androidx.compose.animation.core.tween
+import androidx.compose.animation.core.updateTransition
+import androidx.compose.foundation.background
+import androidx.compose.foundation.gestures.detectDragGesturesAfterLongPress
+import androidx.compose.foundation.gestures.detectTapGestures
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.aspectRatio
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.sizeIn
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Mic
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.Icon
+import androidx.compose.material3.LocalContentColor
+import androidx.compose.material3.RichTooltipBox
+import androidx.compose.material3.RichTooltipState
+import androidx.compose.material3.Text
+import androidx.compose.material3.contentColorFor
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.graphicsLayer
+import androidx.compose.ui.input.pointer.pointerInput
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.unit.dp
+import kotlinx.coroutines.launch
+import kotlin.math.abs
+
+val swipeToCancelThreshold: () -> Float
+    @Composable get() = with(LocalDensity.current) { { 200.dp.toPx() } }
+
+private val verticalThreshold: () -> Float
+    @Composable get() = with(LocalDensity.current) { { 80.dp.toPx() } }
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun RecordButton(
+    recording: Boolean,
+    swipeOffset: MutableState<Float>,
+    onStartRecording: () -> Boolean,
+    onFinishRecording: () -> Unit,
+    onCancelRecording: () -> Unit,
+    modifier: Modifier = Modifier
+) {
+    val transition = updateTransition(targetState = recording, label = "record")
+    val scale = transition.animateFloat(
+        transitionSpec = { spring(Spring.DampingRatioMediumBouncy, Spring.StiffnessLow) },
+        label = "record-scale",
+        targetValueByState = { rec -> if (rec) 2f else 1f }
+    )
+    val containerAlpha = transition.animateFloat(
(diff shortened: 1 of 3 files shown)


Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/UserInput.kt
```diff
@@ -369,6 +393,7 @@ private fun NotAvailablePopup(onDismissed: () -> Unit) {
 val KeyboardShownKey = SemanticsPropertyKey<Boolean>("KeyboardShownKey")
 var SemanticsPropertyReceiver.keyboardShownProperty by KeyboardShownKey
 
+@OptIn(ExperimentalAnimationApi::class)
```
riggaroo, line 396: AnimatedContent shouldn't be experimental anymore, can you double check if you need this?

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
```diff
@@ -0,0 +1,165 @@
...
+                    horizontalSwipeProgress.value = 0f
+                    offsetY = 0f
+                    onStartRecording()
+                },
+                onDragCancel = { onCancelRecording() },
+                onDragEnd = { onFinishRecording() },
+                onDrag = { _, dragAmount ->
+                    horizontalSwipeProgress.value += dragAmount.x
+                    offsetY += dragAmount.y
+                    val offsetX = horizontalSwipeProgress.value
```
riggaroo, line 155: nit: this extra allocation can be removed and just use dragAmount.x

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
```diff
@@ -0,0 +1,165 @@
...
+                        onStartRecording,
+                        onFinishRecording,
+                        onCancelRecording,
+                    )
+            )
+        }
+    }
+}
+
+private fun Modifier.recordGesture(
```
riggaroo, line 131: nit: maybe should be named `voiceRecordingGesture`, or something similar.. Reading this initially I thought it meant recording the gesture.

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
```diff
@@ -0,0 +1,165 @@
...
+                    alpha = containerAlpha.value
+                    scaleX = scale.value; scaleY = scale.value
+                }
+                .clip(CircleShape)
+                .background(LocalContentColor.current)
+        )
+        val scope = rememberCoroutineScope()
+        val tooltipState = remember { RichTooltipState() }
+        RichTooltipBox(
+            text = { Text("Touch and hold to record") },
```
riggaroo, line 107: Extract Strings into strings.xml files please 🙏🏻

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
```diff
@@ -0,0 +1,165 @@
...
+                        onCancelRecording,
+                    )
+            )
+        }
+    }
+}
+
+private fun Modifier.recordGesture(
+    horizontalSwipeProgress: MutableState<Float>,
+    swipeToCancelThreshold: Float,
```
riggaroo, line 133: could you pass those functions as defaults? Feels like this is exposing quite a bit of extra detail that may not be needed. Or is this something that could be taken from the container size? How does it work with a larger screen?

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
```diff
@@ -0,0 +1,165 @@
...
+    modifier: Modifier = Modifier
+) {
+    val transition = updateTransition(targetState = recording, label = "record")
+    val scale = transition.animateFloat(
+        transitionSpec = { spring(Spring.DampingRatioMediumBouncy, Spring.StiffnessLow) },
+        label = "record-scale",
+        targetValueByState = { rec -> if (rec) 2f else 1f }
+    )
+    val containerAlpha = transition.animateFloat(
+        transitionSpec = { tween(2000) },
```
riggaroo, line 78: nit: Any reason why 2000 here and 200 in the color?

riggaroo:
Looks cool! 😎  What about pulsing the size of purple background as you record? Just a suggestion, looks fun as is!

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/UserInput.kt
JolandaVerhoef: In Compose Animation 1.4.3 (latest stable release) it's still marked as experimental.

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
JolandaVerhoef: not sure but it sure looks nice?

riggaroo: good enough for me then 😄

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
JolandaVerhoef: The values should be fixed no matter the size of the screen (it's basically just large enough values to make sure you're being intentional with your cancel affordance).

So what I was struggling with here is that you need screen density to get those fixed values. Now I can make my modifier stateful using `composed` but I don't think we should be doing that.

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
riggaroo: oooh, missed that!

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/UserInput.kt
riggaroo: Yeah, 1.5.0 then, bummer.

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
JolandaVerhoef: Done

riggaroo (approved):

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
```diff
@@ -0,0 +1,169 @@
...
+    @Composable get() = with(LocalDensity.current) { { 200.dp.toPx() } }
+
+private val verticalThreshold: () -> Float
+    @Composable get() = with(LocalDensity.current) { { 80.dp.toPx() } }
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun RecordButton(
+    recording: Boolean,
+    swipeOffset: MutableState<Float>,
```
bentrengrove, line 68: This is normally considered an anti-pattern, use a lambda to forward the state without reading it instead but I think you are doing it because you are mutating it? Doesn't that break UDF? Should it instead be a onSwipeChanged callback for mutation?

bentrengrove: That works as well but feels like overkill for one float. It's not that different to Slider and they just use the normal value and onValueChanged callback.

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/Slider.kt;l=150?q=Slider.kt

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/UserInput.kt
```diff
@@ -390,49 +417,151 @@ private fun UserInputText(
...
+                .align(Alignment.CenterStart)
+                .padding(start = 32.dp),
+            text = stringResource(R.string.textfield_hint),
+            style = MaterialTheme.typography.bodyLarge.copy(color = disableContentColor)
+        )
+    }
+}
+
+@Composable
+private fun RecordingIndicator(swipeOffset: State<Float>) {
```
bentrengrove, line 505: `swipeOffset` should be a lambda here not State. The pattern exists to make testing and previews easier to write because you don't need to alloc a state object and can just pass in something like `{ 0.5f }`

Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/RecordButton.kt
```diff
@@ -0,0 +1,168 @@
...
+        )
+        val scope = rememberCoroutineScope()
+        val tooltipState = remember { RichTooltipState() }
+        RichTooltipBox(
+            text = { Text(stringResource(R.string.touch_and_hold_to_record)) },
+            tooltipState = tooltipState
+        ) {
+            Icon(
+                Icons.Default.Mic,
+                contentDescription = "Record voice message",
```
riggaroo, line 111: extract to strings.xml?

PR #1143, opened 2023-06-08
replaced BackPressHandler with BackHandler

Abhimanyu14:


Jetchat/app/src/main/java/com/example/compose/jetchat/conversation/UserInput.kt
```diff
@@ -126,7 +127,9 @@ fun UserInput(
 
     // Intercept back navigation if there's a InputSelector visible
     if (currentInputSelector != InputSelector.NONE) {
-        BackPressHandler(onBackPressed = dismissKeyboard)
+        BackHandler {
```
androidmahi, line 130: enabled value is missed here?

Abhimanyu14: It is `true` by default.

PR #1149, opened 2023-06-23
[Jetsnack] Fixes #1137 - LazyColumn last element partially visible

### LINKED ISSUE #1137: [Jetsnack]: Navigate case LazyColumn last element partially visible.

### Sample app

Jetsnack

### What happened?

1. Scroll a LazyColumn to bottom. 
2. Click on any element in the LazyColumn. It navigate me to the detail page. 
3. Press back button, it back to the LazyColumn page but not at bottom position. The last element in LazyCoumn is partially visible, I expect it is fully visible.

here's my bug reproduction screen recording.

https://github.com/android/compose-samples/assets/78646351/1bca7436-0afb-4b07-9aa7-9bae3aa6e7a8

### Relevant logcat output

_No response_

### Code of Conduct

rool78:
Fixes #1137 

Current:
[Jetsnack-1137-current.webm](https://github.com/android/compose-samples/assets/54546499/5bce1f5f-3066-4b9d-9867-345c52cfdb81)

Fix:
[Jetsnack-1137-fix.webm](https://github.com/android/compose-samples/assets/54546499/b90e4f52-6df3-44df-a74f-7bc50b002d56)

### DIFF
--- a/Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
+++ b/Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
@@ -16,17 +16,11 @@
 
 package com.example.jetsnack.ui
 
-import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.systemBarsPadding
 import androidx.compose.material.SnackbarHost
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.unit.dp
 import androidx.navigation.NavBackStackEntry
 import androidx.navigation.NavGraphBuilder
 import androidx.navigation.NavType
@@ -37,7 +31,6 @@ import androidx.navigation.navigation
 import com.example.jetsnack.ui.components.JetsnackScaffold
 import com.example.jetsnack.ui.components.JetsnackSnackbar
 import com.example.jetsnack.ui.home.HomeSections
-import com.example.jetsnack.ui.home.JetsnackBottomBar
 import com.example.jetsnack.ui.home.addHomeGraph
 import com.example.jetsnack.ui.snackdetail.SnackDetail
 import com.example.jetsnack.ui.theme.JetsnackTheme
@@ -47,15 +40,6 @@ fun JetsnackApp() {
     JetsnackTheme {
         val appState = rememberJetsnackAppState()
         JetsnackScaffold(
-            bottomBar = {
-                if (appState.shouldShowBottomBar) {
-                    JetsnackBottomBar(
-                        tabs = appState.bottomBarTabs,
-                        currentRoute = appState.currentRoute!!,
-                        navigateToRoute = appState::navigateToBottomBarRoute
-                    )
-                }
-            },
             snackbarHost = {
                 SnackbarHost(
                     hostState = it,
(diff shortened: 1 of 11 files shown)


style96 (commented):
I checked changes. It worked well with api v28 phone and v33 phone.

rool78:
Thanks @style96! I also saw that you solved it in this [PR](https://github.com/android/compose-samples/pull/1150). Maybe your solution is simpler, wdyt? Let's also see @florina-muntenescu thoughts 🙂

riggaroo:
Thanks @rool78 and @style96 for the detailed pull requests and discussion. Appreciate your work here, after looking at this with the team, we believe the correct fix would be to remove the `shouldShowBottomBar` variable all together, and placing the Scaffold (and therefore Bottom navigation) in each Composable that requires it instead. 

The dialog approach, although it fixes the problem, its a patch that'd be exposed again later if a new composable is introduced that does a similar thing to the current SnackDetail composable. 
Passing around padding values is also not great for long term maintenance of the screens, you'd need to keep track of where or where not to use the values. 

Please let us know if either of you would like to update or submit a new PR for the change I've described :) 

Thanks again for the help!

rool78:
Thanks @riggaroo. Makes sense! I can update this PR

rool78:
PR was updated with @riggaroo suggested approach, a review would be appreciated

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -41,15 +40,6 @@ fun JetsnackApp() {
...
-                if (appState.shouldShowBottomBar) {
-                    JetsnackBottomBar(
-                        tabs = appState.bottomBarTabs,
-                        currentRoute = appState.currentRoute!!,
-                        navigateToRoute = appState::navigateToBottomBarRoute
-                    )
-                }
-            },
             snackbarHost = {
                 SnackbarHost(
```
riggaroo, line 44: The Snackbars are now obscuring the content of BottomBar, I think it would be good to include the SnackbarHost per Scaffold too, where required.

rool78: Good catch! It's because scaffoldState is managed at UI tree root level (JetsnackApp) and we removed the bottom bar there. In my last commit I'm using `rememberJetsnackAppState` in Cart composable to access scaffoldSate, I think is a bit overkill as I don't need access to the navController (anyway it's solving the problem). I also don't like passing the scaffoldState down the UI tree. Maybe I could refactor `JetsnackAppState` removing scaffoldState from it and create a custom `JetsnackScaffoldState`. Something like this, what do you think?
```
/**
 * Remember and created an instance of [JetsnackScaffoldState]
 */
@Composable
fun rememberJetsnackScaffoldState(
    scaffoldState: ScaffoldState = rememberScaffoldState(),
    snackbarManager: SnackbarManager = SnackbarManager,
    resources: Resources = resources(),
    coroutineScope: CoroutineScope = rememberCoroutineScope()
    ): JetsnackScaffoldState = remember(scaffoldState, snackbarManager, resources, coroutineScope) {
        JetsnackScaffoldState(scaffoldState, snackbarManager, resources, coroutineScope)
}

/**
 * Responsible for holding [ScaffoldState], handles the logic of showing snackbar messages
 */
@Stable
class JetsnackScaffoldState(
    val scaffoldState: ScaffoldState,
    private val snackbarManager: SnackbarManager,
    private val resources: Resources,
    coroutineScope: CoroutineScope
) {
    // Process snackbars coming from SnackbarManager
    init {
        coroutineScope.launch {
            snackbarManager.messages.collect { currentMessages ->
                if (currentMessages.isNotEmpty()) {
                    val message = currentMessages[0]
                    val text = resources.getText(message.messageId)
                    // Notify the SnackbarManager so it can remove the current message from the list
                    snackbarManager.setMessageShown(message.id)
                    // Display the snackbar on the screen. `showSnackbar` is a function
                    // that suspends until the snackbar disappears from the screen
                    scaffoldState.snackbarHostState.showSnackbar(text.toString())
                }
            }
        }
    }
}
```

riggaroo: Yeah the proposed change will be better I agree.

riggaroo:
@rool78 thanks so much - you need to run spotless to make the build pass :)

rool78:
@riggaroo done! could we relaunch workflows? :)

riggaroo (approved):
Thanks for the contritbution.

akaSourav:
Seems like placing Bottombar in each Scaffold ruins the bottomnavigationbar animation. The Animation isn't smooth anymore.

PR #1167, opened 2023-07-24
[Crane] Implement updates to transition animation api under Crane's Home screen

MikeAde1:
** The label parameter was added to color animation transitions for the non-editable input field on the Home screen. Adding this label makes it easier to inspect this transition in the Animation Preview. 

**   Replaced the experimental “with” extension function used alongside the fade-in animation with the stable “togetherWith” function.

No major logic was altered.

### DIFF
--- a/Crane/app/src/main/res/values/strings.xml
+++ b/Crane/app/src/main/res/values/strings.xml
@@ -39,6 +39,8 @@
     <string name="select_destination_to_caption">To</string>
     <string name="select_dates">Select Dates</string>
     <string name="error_max_people">Error: We don\'t support more than %d people</string>
+    <string name="tint_transition_spec">tintTransitionSpec</string>
+    <string name="tint_transition">tintTransition</string>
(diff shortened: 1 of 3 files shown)


Crane/app/src/main/res/values/strings.xml
```diff
@@ -39,6 +39,8 @@
     <string name="select_destination_to_caption">To</string>
     <string name="select_dates">Select Dates</string>
     <string name="error_max_people">Error: We don\'t support more than %d people</string>
+    <string name="tint_transition_spec">tintTransitionSpec</string>
+    <string name="tint_transition">tintTransition</string>
```
mlykotom, line 43: This doesn't have to be part of the strings.xml as this information is just for the tooling, not for the app.
You can just keep the strings directly in the code.

mlykotom (changes requested):

mlykotom (approved):

mlykotom (approved):

PR #1168, opened 2023-07-26
Jetcast: take account of non-itune and <item> categories

257:


257:
rebased, please re-open.

PR #1175, opened 2023-09-07
Adjustment in the logic of drawer menu and in the logo layout in CraneApp

FelipeIgansi:
In this PR I changed the logic of some parts of the app to simplify and clarify what the code is doing.
I saw that the layout of the menu was quite tight (not much space), and I adjusted this.


bentrengrove:
Thank you for the contribution, unfortunately we aren't doing design changes to the samples right now so I will close this off. I think I prefer the spacer approach the existing sample code is using as well rather than the padding modifier, either way as it is just preference I don't think it is worth the change.

FelipeIgansi:
In this change, I aimed to change the visual part of the menu, as it was not as shown in the print of CreneApp. 
As for Spacer, I personally think it is more interesting at times to use Modifier + padding informing which side will have the space, as I think it is more readable.

And of course, I understand that to maintain the project standard it makes more sense to maintain Spacer.

PR #1185, opened 2023-09-21
[JetLagged] Add custom usage of Predictive back to Jetlagged home screen sample. 

riggaroo:
Add predictive back handling using `PredictiveBackHandler` from `activity-compose`. The gesture first shows the drawer in the case of the drawer being closed, when its open, it propagates the gesture back to the home gesture, which is enabled by setting `android:enableOnBackInvokedCallback="true"` in `AndroidManifest.xml`

https://github.com/android/compose-samples/assets/9973046/bc39217d-3314-4123-bf6b-0c0c19f09514


muhammedheajneh:
riggaroo/jetlagged-predictive-back

PR #1230, opened 2023-11-27
Fix background parallax 

mlykotom:
This used to be a feature, but then when replacing with LazyRow, it was dropped.
Let's fix it + use the deferred modifier to properly read scroll state.

### DIFF
--- a/Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Gradient.kt
+++ b/Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Gradient.kt
@@ -21,12 +21,14 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.border
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.composed
+import androidx.compose.ui.draw.drawBehind
 import androidx.compose.ui.draw.drawWithContent
 import androidx.compose.ui.graphics.BlendMode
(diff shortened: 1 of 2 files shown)


Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Gradient.kt
```diff
@@ -45,20 +47,37 @@ fun Modifier.offsetGradientBackground(
     colors: List<Color>,
     width: Float,
     offset: Float = 0f
-) = background(
+) = this then background(
```
bentrengrove, line 50: I think the `this then` here is double adding the modifier because you have implicitly captured `this` already by not doing `Modifier.background` but just `background` which is actually `this.background`

We probably need another lint check for this as well!

bentrengrove (approved):
Nice! Thanks for doing this

Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Gradient.kt
mlykotom: I'm actually not sure when you're supposed to use which. Happy to write the lint check!

bentrengrove: You basically just have to use `this` somewhere in your implementation. By just writing `background` you've done it

PR #1231, opened 2023-12-06
[Jetlagged] Migrate background AGSL modifier to use Modifier.Node

riggaroo:
Following our new guidance, the `Modifier.yellowBackground()` was migrated to use `Modifier.Node`. https://developer.android.com/jetpack/compose/custom-modifiers#animating_modifier


bentrengrove (approved):
LGTM

PR #1258, opened 2024-02-14
Update actions dependencies

riggaroo:
Need an action that updates dependencies on actions :P


arriolac:
> Need an action that updates dependencies on actions :P

Looks like that is a thing 🤖  https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#enabling-dependabot-version-updates-for-actions

Issue #1260, opened 2024-02-15
[FR]: Reply is laid out edge-to-edge in 2 & 3-button navigation mode


manabu-nakamura:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is this a feature request for one of the samples?

- [X] Yes, this is a specific request related to this samples repo.

### Sample app

Reply

### Describe the problem

Reply is not laid out edge-to-edge in 2 & 3-button navigation mode (it is not drawn behind the system navigation bar in 2 & 3-button navigation mode).
![Screenshot_20240215_100721](https://github.com/android/compose-samples/assets/71050561/05a556b8-d26c-4585-967e-cc43c7a65eac)

### Describe the solution

Reply is laid out edge-to-edge in 2 & 3-button navigation mode (it is drawn behind the system navigation bar in 2 & 3-button navigation mode).
![Screenshot_20240215_100808](https://github.com/android/compose-samples/assets/71050561/50708619-e1c6-40b6-8eee-173cad1d19b1)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1261, opened 2024-02-15
Reply is laid out edge-to-edge in 2 & 3-button navigation mode

manabu-nakamura:
#1260

(before) Reply is not laid out edge-to-edge in 2 & 3-button navigation mode (it is not drawn behind the system navigation bar in 2 & 3-button navigation mode).
Screenshot_20240215_100721
(after) Reply is laid out edge-to-edge in 2 & 3-button navigation mode (it is drawn behind the system navigation bar in 2 & 3-button navigation mode).
Screenshot_20240215_100808

### DIFF
--- a/Reply/app/src/main/java/com/example/reply/ui/MainActivity.kt
+++ b/Reply/app/src/main/java/com/example/reply/ui/MainActivity.kt
@@ -19,7 +19,6 @@ package com.example.reply.ui
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
-import androidx.activity.enableEdgeToEdge
 import androidx.activity.viewModels
 import androidx.compose.material3.windowsizeclass.ExperimentalMaterial3WindowSizeClassApi
 import androidx.compose.material3.windowsizeclass.WindowSizeClass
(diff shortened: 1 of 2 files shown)


Reply/app/src/main/java/com/example/reply/ui/MainActivity.kt
```diff
@@ -42,6 +43,7 @@ class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         enableEdgeToEdge()
         super.onCreate(savedInstanceState)
+        ViewCompat.setOnApplyWindowInsetsListener(window.decorView) { _, insets -> insets }
```
yaraki, line 46: This gets the inset and return it without doing anything. It seems to me that it doesn't do what the before/after screenshots do.
Is PR missing some commits?

manabu-nakamura: No, it isn't.

manabu-nakamura: https://github.com/android/compose-samples/pull/1265/commits/a1e7c597eb85cec716f5e83575a18b283278f6e2:
Screenshot_20240220_150514

Issue #1262, opened 2024-02-15
[FR]: Jetsnack is laid out edge-to-edge in 2 & 3-button navigation mode


manabu-nakamura:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is this a feature request for one of the samples?

- [X] Yes, this is a specific request related to this samples repo.

### Sample app

Jetsnack

### Describe the problem

Jetsnack is not laid out edge-to-edge in 2 & 3-button navigation mode (it is not drawn behind the system navigation bar in 2 & 3-button navigation mode).
![Screenshot_20240215_103236](https://github.com/android/compose-samples/assets/71050561/de5858e9-1105-4039-a5e6-fabf922a3221)

### Describe the solution

Jetsnack is laid out edge-to-edge in 2 & 3-button navigation mode (it is drawn behind the system navigation bar in 2 & 3-button navigation mode).
![Screenshot_20240215_103317](https://github.com/android/compose-samples/assets/71050561/8c255455-3986-4a1d-bbc3-fc6e06b4757c)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1263, opened 2024-02-15
Jetsnack is laid out edge-to-edge in 2 & 3-button navigation mode

manabu-nakamura:
#1262

(before) Jetsnack is not laid out edge-to-edge in 2 & 3-button navigation mode (it is not drawn behind the system navigation bar in 2 & 3-button navigation mode).
Screenshot_20240215_103236
(after) Jetsnack is laid out edge-to-edge in 2 & 3-button navigation mode (it is drawn behind the system navigation bar in 2 & 3-button navigation mode).
Screenshot_20240215_103317

### DIFF
--- a/Jetsnack/app/src/main/java/com/example/jetsnack/ui/MainActivity.kt
+++ b/Jetsnack/app/src/main/java/com/example/jetsnack/ui/MainActivity.kt
@@ -20,11 +20,13 @@ import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.enableEdgeToEdge
+import androidx.core.view.ViewCompat
 
 class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         enableEdgeToEdge()
         super.onCreate(savedInstanceState)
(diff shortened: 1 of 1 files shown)


Jetsnack/app/src/main/java/com/example/jetsnack/ui/MainActivity.kt
```diff
@@ -20,11 +20,13 @@ import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.enableEdgeToEdge
+import androidx.core.view.ViewCompat
 
 class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         enableEdgeToEdge()
         super.onCreate(savedInstanceState)
+        ViewCompat.setOnApplyWindowInsetsListener(window.decorView) { _, insets -> insets }
```
yaraki, line 29: This gets the inset and return it without doing anything. It seems to me that it doesn't do what the before/after screenshots do.
Is PR missing some commits?

manabu-nakamura: No, it isn't.

Issue #1264, opened 2024-02-15
[FR]: Owl is laid out edge-to-edge in 2 & 3-button navigation mode


manabu-nakamura:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is this a feature request for one of the samples?

- [X] Yes, this is a specific request related to this samples repo.

### Sample app

Owl

### Describe the problem

Owl is not laid out edge-to-edge in 2 & 3-button navigation mode (it is not drawn behind the system navigation bar in 2 & 3-button navigation mode).
![Screenshot_20240215_105930](https://github.com/android/compose-samples/assets/71050561/adb98861-18db-4a32-a7c2-6f2bfa8c025d)

### Describe the solution

Owl is laid out edge-to-edge in 2 & 3-button navigation mode (it is drawn behind the system navigation bar in 2 & 3-button navigation mode).
![Screenshot_20240215_110222](https://github.com/android/compose-samples/assets/71050561/1062bbc6-d91b-4908-9017-b7b78c5c04a6)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1265, opened 2024-02-15
Owl is laid out edge-to-edge in 2 & 3-button navigation mode

manabu-nakamura:
#1264

(before) Owl is not laid out edge-to-edge in 2 & 3-button navigation mode (it is not drawn behind the system navigation bar in 2 & 3-button navigation mode).
Screenshot_20240215_105930
(after) Owl is laid out edge-to-edge in 2 & 3-button navigation mode (it is drawn behind the system navigation bar in 2 & 3-button navigation mode).
Screenshot_20240215_110222

### DIFF
--- a/Owl/app/src/main/java/com/example/owl/ui/MainActivity.kt
+++ b/Owl/app/src/main/java/com/example/owl/ui/MainActivity.kt
@@ -16,13 +16,14 @@
 
 package com.example.owl.ui
 
+import android.content.res.Configuration
+import android.graphics.Color
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.SystemBarStyle
 import androidx.activity.compose.setContent
 import androidx.activity.enableEdgeToEdge
 import androidx.core.content.ContextCompat
-import androidx.core.view.ViewCompat
 import com.example.owl.R
 
 class MainActivity : ComponentActivity() {
@@ -31,10 +32,13 @@ class MainActivity : ComponentActivity() {
         enableEdgeToEdge(
             statusBarStyle = SystemBarStyle.dark(
                 ContextCompat.getColor(this, R.color.immersive_sys_ui)
-            )
+            ),
+            navigationBarStyle = if (resources.configuration.uiMode and Configuration.UI_MODE_NIGHT_MASK == Configuration.UI_MODE_NIGHT_YES)
+                SystemBarStyle.dark(Color.TRANSPARENT)
+            else
+                SystemBarStyle.light(Color.TRANSPARENT, Color.TRANSPARENT)
         )
         super.onCreate(savedInstanceState)
-        ViewCompat.setOnApplyWindowInsetsListener(window.decorView) { _, insets -> insets }
 
         setContent {
             OwlApp { finish() }
--- a/Owl/app/src/main/java/com/example/owl/ui/MainActivity.kt
+++ b/Owl/app/src/main/java/com/example/owl/ui/MainActivity.kt
@@ -17,7 +17,6 @@
 package com.example.owl.ui
 
 import android.graphics.Color
-import android.os.Build
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.SystemBarStyle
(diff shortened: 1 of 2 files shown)


Owl/app/src/main/java/com/example/owl/ui/MainActivity.kt
```diff
@@ -33,6 +34,7 @@ class MainActivity : ComponentActivity() {
             )
         )
         super.onCreate(savedInstanceState)
+        ViewCompat.setOnApplyWindowInsetsListener(window.decorView) { _, insets -> insets }
```
yaraki, line 37: This gets the inset and return it without doing anything. It seems to me that it doesn't do what the before/after screenshots do.
Is the PR missing some commits?

manabu-nakamura: No, it isn't.

manabu-nakamura: In portrait orientation, the listener in MDC Catalog app only returns the passed parameter, too. This causes that it is drawn behind the system navigation bar in 2 & 3-button navigation mode.

yaraki: I see, so this is for disabling the contrast enforcement for the 3-button navbar. That is a side effect of preventing the original `onApplyWindowInsets` of the `decorView`.
https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/com/android/internal/policy/DecorView.java;l=1030
The problem is that it also prevents other things done in that method. For example, `R.color.immersive_sys_ui` set above won't be used.

If you want the 3-button nav to be transparent, you should use the `navigationBarStyle` parameter in `enableEdgeToEdge`.

manabu-nakamura: I used it.

manabu-nakamura: I don't understand that window.isNavigationBarContrastEnforced = true only when navigationBarStyle = SystemBarStyle.auto() (window.isNavigationBarContrastEnforced = false when navigationBarStyle = SystemBarStyle.dark() or SystemBarStyle.light()).
https://github.com/androidx/androidx/blob/androidx-main/activity/activity/src/main/java/androidx/activity/EdgeToEdge.kt#L315-L316

yaraki: The default appearance of 3-button navigation is to have a translucent scrim with `isNavigationBarContrastEnforced = true`. It is still considered edge-to-edge.
If you really want to make it transparent, you can do so, but that's optional.

manabu-nakamura: Thank you. But, I don't understand that window.isNavigationBarContrastEnforced = false when navigationBarStyle = SystemBarStyle.dark() or SystemBarStyle.light().

Does this PR still need fixing?

P.S.
Should we use the navigationBarStyle parameter in enableEdgeToEdge instead of ViewCompat.setOnApplyWindowInsetsListener() inside other apps (Jetchat, Jetsnack, and Reply)?

yaraki: `SystemBarStyle.auto` sets up with the recommended style (transparent gesture nav and translucent 3-button nav). The color parameters are for backward compatibility, and not suitable for color customization.

If you really want the bar to be transparent on 3-button nav, then you can use either `SystemBarStyle.light` or `SystemBarStyle.dark`. This app seems to change the background between the light/dark modes, so you need to call the right one based on the mode. I would leave this `enableEdgeToEdge` as is, and add a second call in `Themes.kt` because the mode `isDarkMode` is available there.

I don't recommend using `ViewCompat.setOnApplyWindowInsetsListener` for suppressing the DecorView's default behavior.

And the same goes for all the other PRs.

manabu-nakamura: Thank you. I tried writing it.

manabu-nakamura: https://github.com/android/compose-samples/pull/1265/commits/a1e7c597eb85cec716f5e83575a18b283278f6e2:
Screenshot_20240220_150514

PR #1274, opened 2024-02-27
[Jetcaster] Improve scrolling in Home screen

arriolac:
Changes:
* Support scrolling entire content in home screen
* Fix `PreviewHomeContent`
* Remove non-screen composable view model usages

It would be great to also support collapsing the toolbar while scrolling but that might be better accomplished when migrating Jetcaster to Material3.

[jetcaster_scrolling.webm](https://github.com/android/compose-samples/assets/463186/a0334c6f-d3ac-4d00-ab7e-daccd7180212)


Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -198,72 +211,106 @@ fun HomeContent(
...
-                    modifier = Modifier.fillMaxWidth()
-                )
+                        if (isRefreshing) {
+                            // TODO show a progress indicator or similar
+                        }
 
-                if (featuredPodcasts.isNotEmpty()) {
-                    Spacer(Modifier.height(16.dp))
+                        stickyHeader {
+                            if (homeCategories.isNotEmpty()) {
```
bentrengrove, line 261: nit: It should be the other way around I think?

```
if (homeCategories.isNotEmpty()) {
   stickyHeader {}
}
```

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -198,72 +211,106 @@ fun HomeContent(
...
+                                    onTogglePodcastFollowed = onTogglePodcastFollowed
+                                )
+                            }
+                        }
+                    }
                 }
             }
         }
+    }
+}
```
bentrengrove, line 290: nit: This is super nested, consider breaking it apart into multiple composables where it makes sense

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -198,72 +211,106 @@ fun HomeContent(
...
-            ) {
-                // Draw a scrim over the status bar which matches the app bar
-                Spacer(
-                    Modifier
-                        .background(appBarColor)
-                        .fillMaxWidth()
-                        .windowInsetsTopHeight(WindowInsets.statusBars)
-                )
+            val scrimColor = MaterialTheme.colors.primary.copy(alpha = 0.38f)
+            Scaffold(
```
bentrengrove, line 215: Do we even need the Scaffold? Looks like it is sitting in a Column already that maybe could just be used seeing as its just a TopAppBar and then LazyColumn

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
arriolac: Good catch. Fixed it

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
arriolac: I was thinking a Scaffold would be good if we want the top app bar to dismiss while scrolling. Since I haven't implemented that yet (requires M3) I'll remove the Scaffold for now 👍

PR #1284, opened 2024-03-14
[Reply] Contrast aware updates.

jwill:
Added colors for standard, medium, and high in both light and dark themes. 
Added helper function to select contrast level based on values from SysUI controls. 
Added more variance in application of surface container roles to better indicate elevation and emphasis.


Reply/app/src/main/java/com/example/reply/ui/components/ReplyEmailThreadItem.kt
```diff
@@ -107,31 +111,32 @@ fun ReplyEmailThreadItem(
...
+                        containerColor = MaterialTheme.colorScheme.surfaceBright
                     )
                 ) {
                     Text(
                         text = stringResource(id = R.string.reply),
-                        color = MaterialTheme.colorScheme.onSurfaceVariant
+                        color = MaterialTheme.colorScheme.onSurface
                     )
                 }
+                Spacer(modifier = Modifier.width(8.dp))
```
IanGClifton, line 129: Nit: This spacing is a combination of line 115 and line 129. I'd combine them to just do 12dp in one place, probably 115, but I don't have a strong preference whether it's that or a spacer with no horizontalArrangement

Reply/app/src/main/java/com/example/reply/ui/theme/Theme.kt
```diff
@@ -15,95 +15,332 @@
...
+            window.statusBarColor = replyColorScheme.primary.toArgb()
+            WindowCompat.getInsetsController(window, view).isAppearanceLightStatusBars = darkTheme
+        }
     }
 
     MaterialTheme(
         colorScheme = replyColorScheme,
-        typography = replyTypography,
-        shapes = shapes,
+        typography = replyTypography, shapes = shapes,
```
IanGClifton, line 343: Nit: Should still be separate lines

Reply/app/src/main/java/com/example/reply/ui/theme/Theme.kt
```diff
@@ -15,95 +15,332 @@
...
+
+    MaterialTheme(
+        colorScheme = replyColorScheme,
+        typography = replyTypography, shapes = shapes,
+        content = content
+    )
+}
+
 @Composable
 fun ReplyTheme(
```
IanGClifton, line 317: Is the ReplyTheme still used anywhere?

IanGClifton (commented):
Very minor feedback; I'm assuming all the colors are perfect and beautiful in every way

Reply/app/src/main/java/com/example/reply/ui/theme/Theme.kt
jwill: It was being used in previews but I've replaced all those references and removed ReplyTheme. Done.

Reply/app/src/main/java/com/example/reply/ui/theme/Theme.kt
jwill: Done.

Reply/app/src/main/java/com/example/reply/ui/components/ReplyEmailThreadItem.kt
jwill: Done.

IanGClifton (approved):

PR #1287, opened 2024-03-19
[Jetcaster] Migrate from M2 to M3.

arriolac:
Upgrading Jetcaster from M2 to M3. This migration changes one of the original purposes of the sample which was to demonstrate dynamic theming. We would still like to demonstrate dynamic theming but using a more modern approach using Material's [material-color-utilities](https://github.com/material-foundation/material-color-utilities).
Screenshot_20240319_131815


Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -403,28 +368,33 @@ fun HomeCategoryTabIndicator(
...
-        FollowedPodcastCarouselItem(
-            podcastImageUrl = podcast.imageUrl,
-            podcastTitle = podcast.title,
-            onUnfollowedClick = { onPodcastUnfollowed(podcast.uri) },
-            lastEpisodeDateText = lastEpisodeDate?.let { lastUpdated(it) },
-            modifier = Modifier
-                .padding(4.dp)
-                .fillMaxSize()
+    val lastIndex = items.size - 1
+    LazyRow(
```
bentrengrove, line 378: Why was this changed? I assume the old one was to achieve snapping behavior, which should be possible without HorizontalPager if you need. But also there is a new Carousel in M3 which might be good also, I haven't actually looked into it though

Jetcaster/app/src/main/java/com/example/jetcaster/ui/theme/Color.kt
```diff
@@ -16,37 +16,31 @@
...
-    primary = Yellow800,
-    onPrimary = Color.Black,
-    primaryVariant = Yellow800,
-    secondary = Yellow800,
-    onSecondary = Color.Black,
-    error = Red300,
-    onError = Color.Black
-)
+// /**
+// * Return the fully opaque color that results from compositing [onSurface] atop [surface] with the
```
bentrengrove, line 27: Delete?

bentrengrove (approved):

bentrengrove:
The status bar color doesn't look correct in your screenshot, should be white.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/theme/Color.kt
arriolac: Deleted. Thanks

arriolac:
> The status bar color doesn't look correct in your screenshot, should be white.

That change is intentional — though the status bar should match `surface` color. Updating that

bentrengrove:
Does the contrast of the text on the status bar color look correct now? Hard to tell from the code

arriolac:
> Does the contrast of the text on the status bar color look correct now? Hard to tell from the code

Ahh good catch. I didn't notice the status bar text contrast. Looks like there was a bug in `JetcasterTheme`.

PR #1289, opened 2024-03-20
[Jetcaster] Refactor: create domain models

arriolac:
Added the following domain layer objects and corresponding test cases:
* `GetLatestFollowedEpisodesUseCase`
* `PodcastCategoryFilterUseCase`
* `FilterableCategoriesUseCase`

This also required creating interfaces for repository objects to enable creating fakes for testing. Introducing these changes improves readability/maintainability of `HomeViewModel` and encourages splitting up responsibilities to smaller easily testable contained [domain] objects.


Jetcaster/core/src/main/java/com/example/jetcaster/core/data/repository/CategoryStore.kt
```diff
@@ -26,32 +26,67 @@ import com.example.jetcaster.core.data.database.model.PodcastCategoryEntry
 import com.example.jetcaster.core.data.database.model.PodcastWithExtraInfo
 import kotlinx.coroutines.flow.Flow
 
+interface CategoryStore {
```
mlykotom, line 29: I would say you don't need to have this abstraction, because this abstraction is provided by room by having the DAO abstract classes. Basically what you can have is implement `TestCategoriesDao` (and others) and have the simpler logic there. While the DAO class is annotated, from the perspective of the tests it's just a pure abstract class.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/discover/Discover.kt
```diff
@@ -38,23 +38,19 @@ import androidx.compose.ui.res.stringResource
...
-data class DiscoverViewState(
-    val categories: List<Category> = emptyList(),
-    val selectedCategory: Category? = null
-)
-
 fun LazyListScope.discoverItems(
-    discoverViewState: DiscoverViewState,
-    podcastCategoryViewState: PodcastCategoryViewState,
+    filterableCategories: List<FilterableCategory>,
+    podcastCategoryFilterResult: PodcastCategoryFilterResult,
```
mlykotom, line 48: should we consider enabling strong skipping mode @bentrengrove ?

Jetcaster/core/src/main/java/com/example/jetcaster/core/data/domain/FilterableCategoriesUseCase.kt
```diff
@@ -0,0 +1,50 @@
...
+import kotlinx.coroutines.flow.onEach
+
+/**
+ * Use case for categories that can be used to filter podcasts.
+ */
+class FilterableCategoriesUseCase(
+    private val categoryStore: CategoryStore
+) {
+    operator fun invoke(
+        selectedCategoryStateFlow: MutableStateFlow<Category?>
```
mlykotom, line 34: I'm not a fan of mutating the stateflow this way as a side effect of this use case.
If I understand this correctly, the `selectedCategoryStateFlow` might not be needed in this use case anyway. We could have `onEach{}` in the viewmodel that uses this use case and mutate that `selectedCategoryFlow` there.

Jetcaster/core/src/main/java/com/example/jetcaster/core/data/domain/GetLatestFollowedEpisodesUseCase.kt
```diff
@@ -0,0 +1,52 @@
...
+                    flowOf(emptyList())
+                } else {
+                    combine(
+                        followedPodcasts.map { p ->
+                            episodeStore.episodesInPodcast(p.podcast.uri, 5)
+                        }
+                    ) { allEpisodes ->
+                        allEpisodes.toList().flatten().sortedByDescending { it.episode.published }
+                    }
+                }
```
mlykotom, line 50: I found this quite complicating and I thought whether it wouldn't be simpler and more performant if we had a new query that would find all of the queries at once.

```
@Query(
        """
        SELECT * FROM episodes WHERE podcast_uri IN (:podcastUri)
        ORDER BY datetime(published) DESC
        LIMIT :limit
        """
    )
    abstract fun episodesForMultiplePodcastUri(podcastUri: List<String>, limit: Int): Flow<List<EpisodeToPodcast>>
```
    
    The only difference is now to figure out the limit, because before it was `Number of items * 5` (which technically could be a lot)
    
    (I haven't tested this)

The callsite is then just 
```
  .flatMapLatest { followedPodcasts ->
      episodeStore.episodesForMultiplePodcastUri(
          followedPodcasts.map { it.podcast.uri },
          25
      )
  }
```

mlykotom (commented):
Chiming in as I was around and this was interesting 😊

Jetcaster/core/src/main/java/com/example/jetcaster/core/data/repository/CategoryStore.kt
arriolac: I originally went with this approach but then decided to create an interface for the `*Store` objects because creating `Test*Dao` objects was too time-consuming :/ This would probably be the right approach though if I were writing tests for all layers.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/discover/Discover.kt
arriolac: Do we demonstrate this right now in a sample? Definitely open to enabling strong skipping mode!

Jetcaster/core/src/main/java/com/example/jetcaster/core/data/domain/FilterableCategoriesUseCase.kt
arriolac: Agreed that it would be good to remove this side effect. Will update.

Jetcaster/core/src/main/java/com/example/jetcaster/core/data/domain/GetLatestFollowedEpisodesUseCase.kt
arriolac: Thanks - that is simpler. Will update

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/discover/Discover.kt
bentrengrove: It's not in any samples yet, you are welcome to add it. I am planning on doing it in the future to all samples (or it will just be switched on by default)

arriolac:
Thanks for your review @mlykotom ! Feedback addressed in [92c1b86](https://github.com/android/compose-samples/pull/1289/commits/92c1b863e1fad856baa2a9df70e50de9cb32407f)

Jetcaster/core/src/main/java/com/example/jetcaster/core/data/domain/FilterableCategoriesUseCase.kt
mlykotom: I realised the problem might be more in depth with how you modelled the object. 

The `FilterableCategory` has now `isSelected` flag, which I think doesn't represent the real state as (if I read it right) only one category can be selected. This way we add complexity to all of the domain models, where in reality this is just a UI layer selection. 

I think things would be simplified if you passed to the composables `selectedCategory: Category?` as well and just had  `categories: List<Category>` instead of `filterableCategories`.

Forgive me if I don't see something. I see now basically what I suggest was sort of there before 🤔 😅 .

mlykotom: Ok, now I understand your point that you wanted to have this functionality be testable and actually part of the domain state of your app.

I still think there could be improvement instead.

Instead of having `isSelected` in each category, I would suggest having something like 
```
data class FilterableModel(
    val categories: List<Category>,
    val selectedCategory: Category?
)
```

This way the use case might be still quite simple without the lambda and would still contain the logic to select the first category. Something like:

```
        categoryStore.categoriesSortedByPodcastCount()
            .map {
                FilterableModel(
                    categories = it,
                    // If we haven't got a selected category yet, select the first
                    selectedCategory = selectedCategory ?: it.firstOrNull()
                )
            }
```

The view state in the viewmodel would either be as before with the `DiscoverViewState`, or as is now, but you would add `selectedCategory:Category?` all the way down where it's needed.

The use case tests can be simplified this way as well without the side effects:

```
    @Test
    fun whenNoSelectedCategory_onEmptySelectedCategoryInvoked() = runTest {
        val filterableCategories = useCase(null).first()
        assertEquals(
            filterableCategories.categories[0],
            filterableCategories.selectedCategory
        )
    }

    @Test
    fun whenSelectedCategory_correctFilterableCategoryIsSelected() = runTest {
        val selectedCategory = testCategories[2]
        val filterableCategories = useCase(selectedCategory)

        assertEquals(filterableCategories.first().selectedCategory, selectedCategory)
    }
```


What do you think? 😊

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/discover/Discover.kt
```diff
@@ -84,20 +81,21 @@ private val emptyTabIndicator: @Composable (List<TabPosition>) -> Unit = {}
...
-    categories: List<Category>,
-    selectedCategory: Category,
+    filterableCategoriesModel: FilterableCategoriesModel,
     onCategorySelected: (Category) -> Unit,
     modifier: Modifier = Modifier
 ) {
-    val selectedIndex = categories.indexOfFirst { it == selectedCategory }
+    val selectedIndex = filterableCategoriesModel.categories.indexOfFirst {
+        filterableCategoriesModel.selectedCategory == it
+    }
```
mlykotom, line 90: You can simplify this to standard `indexOf` 

```suggestion
    val selectedIndex = filterableCategoriesModel.categories.indexOf(filterableCategoriesModel.selectedCategory)
```

Or you could even encapsulate it into the `FilterableCategoriesModel` as a `val selectedCategoryIndex: Int = categories.indexOf(selectedCategory)`

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/discover/Discover.kt
```diff
@@ -38,23 +38,21 @@ import androidx.compose.ui.res.stringResource
...
+    filterableCategoriesModel: FilterableCategoriesModel,
+    podcastCategoryFilterResult: PodcastCategoryFilterResult,
     navigateToPlayer: (String) -> Unit,
     onCategorySelected: (Category) -> Unit,
     onTogglePodcastFollowed: (String) -> Unit,
 ) {
-    if (discoverViewState.categories.isEmpty() || discoverViewState.selectedCategory == null) {
+    if (filterableCategoriesModel.categories.isEmpty() ||
+        filterableCategoriesModel.selectedCategory == null
+    ) {
```
mlykotom, line 55: I would suggest encapsulating this logic into the `FilterableCategoriesModel` something like 
```
val isEmptyState = categories.isEmpty() || selectedCategory == null
```

and therefore here you can check just for that boolean without worrying what's the real implementation

Jetcaster/core/src/main/java/com/example/jetcaster/core/data/domain/FilterableCategoriesUseCase.kt
```diff
@@ -0,0 +1,39 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+
+/**
+ * Use case for categories that can be used to filter podcasts.
+ */
+class FilterableCategoriesUseCase(
+    private val categoryStore: CategoryStore
+) {
+    operator fun invoke(selectedCategory: Category?): Flow<FilterableCategoriesModel> =
```
mlykotom, line 31: NIT: maybe name it `previouslySelectedCategory` to be more clear why this parameter exists?

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/HomeViewModel.kt
```diff
@@ -129,23 +75,29 @@ class HomeViewModel(
                 selectedHomeCategory,
                 podcastStore.followedPodcastsSortedByLastEpisode(limit = 20),
                 refreshing,
-                discover,
-                podcastCategory,
-                libraryEpisodes
+                _selectedCategory.flatMapLatest { selectedCategory ->
+                    filterableCategoriesUseCase(selectedCategory).onEach {
+                        _selectedCategory.value = it.selectedCategory
```
mlykotom, line 80: I'm a bit afraid of this, because if this wasn't a `StateFlow`, which automatically uses `distinctUntilChanged`, then this would cause infinite loop.  I think we're mixing kind of two states `_selectedCategory` and `HomeViewState` in one viewmodel. Anyway, I think it can be beyond this PR to change that.

What I would change for readability though is to just have as part of the combine lambda, so you don't have to have tne `onEach` because that what happens anyway:
```

_selectedCategory.value = filterableCategories.selectedCategory

HomeViewState(
...
```

mlykotom (approved):
There are some comments left, but in general LGTM

Jetcaster/core/src/main/java/com/example/jetcaster/core/data/domain/FilterableCategoriesUseCase.kt
arriolac: It's technically not the previous selection. I'll add a javadoc on the invoke method to clarify its purpose.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/HomeViewModel.kt
arriolac: The original intent of this PR is to simplify the responsibility of `HomeViewModel` without changing the implementation (so mostly refactoring and moving code and not changing the underlying implementation). I agree that the current implementation can be improved though so let's address that separately.

PR #1292, opened 2024-03-21
[Jetcaster] Handle no subscribed podcasts empty state

arriolac:
Changes:
* Handles empty state when there are no subscribed podcasts (library / discover tab hidden when no podcasts are followed)
* Reverting `FollowedPodcasts` to use `HorizontalPager` so that the list of episodes changes as the user swipes through podcasts
 
[Screen_recording_20240321_163113.webm](https://github.com/android/compose-samples/assets/463186/0b90b672-7f04-48dd-b94a-7347e159d91e)

### DIFF
--- a/Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
+++ b/Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
@@ -14,11 +14,14 @@
  * limitations under the License.
  */
 
+@file:OptIn(ExperimentalFoundationApi::class)
+
 package com.example.jetcaster.ui.home
 
 import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
@@ -26,7 +29,6 @@ import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.WindowInsetsSides
-import androidx.compose.foundation.layout.aspectRatio
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
@@ -40,8 +42,10 @@ import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.windowInsetsPadding
 import androidx.compose.foundation.layout.windowInsetsTopHeight
 import androidx.compose.foundation.lazy.LazyColumn
-import androidx.compose.foundation.lazy.LazyRow
-import androidx.compose.foundation.lazy.itemsIndexed
+import androidx.compose.foundation.pager.HorizontalPager
+import androidx.compose.foundation.pager.PageSize
+import androidx.compose.foundation.pager.PagerState
+import androidx.compose.foundation.pager.rememberPagerState
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.AccountCircle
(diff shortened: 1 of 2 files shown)


Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -174,7 +188,15 @@ fun Home(
...
     onTogglePodcastFollowed: (String) -> Unit,
+    onLibraryPodcastSelected: (Podcast?) -> Unit
 ) {
+    // Effect that changes the home category selection when there are no subscribed podcasts
+    LaunchedEffect(key1 = featuredPodcasts) {
+        if (featuredPodcasts.isEmpty()) {
+            onHomeCategorySelected(HomeCategory.Discover)
+        }
+    }
+
```
bentrengrove, line 199: I think this should just be in the viewmodel where you combine your state

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -243,11 +266,18 @@ private fun HomeContent(
     onCategorySelected: (Category) -> Unit,
     navigateToPlayer: (String) -> Unit,
     onTogglePodcastFollowed: (String) -> Unit,
+    onLibraryPodcastSelected: (Podcast?) -> Unit
 ) {
+    val pagerState = rememberPagerState { featuredPodcasts.size }
+    LaunchedEffect(pagerState.currentPage, featuredPodcasts) {
+        val podcast = featuredPodcasts.getOrNull(pagerState.currentPage)
+        onLibraryPodcastSelected(podcast?.podcast)
+    }
```
bentrengrove, line 275:
```suggestion
LaunchedEffect(pagerState, featuredPodcasts) {
        snapshotFlow { pagerState.currentPage }
            .collect {
                val podcast = featuredPodcasts.getOrNull(pagerState.currentPage)
                onLibraryPodcastSelected(podcast?.podcast)
            }
    }
```

bentrengrove: Slightly more efficient

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -367,34 +397,54 @@ fun HomeCategoryTabIndicator(
...
+        state = pagerState,
+        modifier = modifier.onSizeChanged { size ->
+            // TODO: this is not quite performant since it requires 2 passes to compute the content
+            // padding. This should be revisited once a carousel component is available.
+            // Alternatively, version 1.7.0-alpha05 of Compose Foundation supports `snapPosition`
+            // which solves this problem and avoids this calculation altogether. Once 1.7.0 is
+            // stable, this implementation can be updated.
+            horizontalPadding = with(density) {
+                (size.width.toDp() - FEATURED_PODCAST_IMAGE_WIDTH_DP) / 2
+            }
```
bentrengrove, line 426: `BoxWithConstraints` might be more desirable here because this two frame padding calculation is visible to the user. It seems weird to me you have to do this in the first place though, I would have thought you could do some combination of pageSpacing, custom page size and horizontal padding to get it to centre

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -367,34 +397,54 @@ fun HomeCategoryTabIndicator(
...
 ) {
-    // TODO: Update this component to a carousel once better support is available
-    val lastIndex = items.size - 1
-    LazyRow(
-        modifier = modifier,
+    val coroutineScope = rememberCoroutineScope()
+
+    var horizontalPadding by remember { mutableStateOf(0.dp) }
+    val density = LocalDensity.current
+    val screenWidth = LocalConfiguration.current.screenWidthDp.dp
```
bentrengrove, line 415: This line isn't used

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -367,34 +397,54 @@ fun HomeCategoryTabIndicator(
...
+            }
+        },
         contentPadding = PaddingValues(
-            start = Keyline1,
-            top = 16.dp,
-            end = Keyline1,
+            horizontal = horizontalPadding,
+            vertical = 16.dp,
+        ),
+        pageSize = PageSize.Fixed(180.dp)
```
bentrengrove, line 432: Wouldn't this be the image width? Currently its the same as the image height

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
arriolac: Good call

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
arriolac: Thanks - will change to `BoxWithConstraints`. I was also hoping to accomplish the same effect without needing the parent's size but unfortunately, that's needed. `snapPolicy` is really what we want here but that's still in alpha.

PR #1293, opened 2024-03-22
[Jetcaster] Add mock support for playing an episode

arriolac:
Introduces `EpisodePlayer` interface and `MockEpisodePlayer` conforming class to mock playing an episode. This enables play/pause and the slider to update as you would expect in a real podcast app. 

[Screen_recording_20240322_160937.webm](https://github.com/android/compose-samples/assets/463186/09a4e7bd-7d99-4381-8ec2-9b2af3bfefac)

#1292 should be reviewed and merged first.

### DIFF
--- a/Jetcaster/core/src/main/java/com/example/jetcaster/core/data/database/dao/EpisodesDao.kt
+++ b/Jetcaster/core/src/main/java/com/example/jetcaster/core/data/database/dao/EpisodesDao.kt
@@ -40,10 +40,10 @@ abstract class EpisodesDao : BaseDao<Episode> {
         """
         SELECT episodes.* FROM episodes
         INNER JOIN podcasts ON episodes.podcast_uri = podcasts.uri
-        WHERE episodes.uri = :uri
+        WHERE episodes.uri = :episodeUri
         """
     )
-    abstract fun episodeAndPodcast(uri: String): Flow<EpisodeToPodcast>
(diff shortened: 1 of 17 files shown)


kul3r4 (approved):

Jetcaster/core/src/main/java/com/example/jetcaster/core/data/database/dao/EpisodesDao.kt
```diff
@@ -36,6 +36,15 @@ abstract class EpisodesDao : BaseDao<Episode> {
...
     abstract fun episode(uri: String): Flow<Episode>
 
+    @Query(
+        """
+        SELECT episodes.* FROM episodes
+        INNER JOIN podcasts ON episodes.podcast_uri = podcasts.uri
+        WHERE episodes.uri = :uri
+        """
+    )
+    abstract fun episodeAndPodcast(uri: String): Flow<EpisodeToPodcast>
```
chikoski, line 46: This is not a blocker, but we could clarify what the uri parameter points at by changing the parameter name to podcastUri

chikoski (approved):

Jetcaster/core/src/main/java/com/example/jetcaster/core/data/database/dao/EpisodesDao.kt
arriolac: This is actually supposed to be `episodeUri`. But case in point, the variable name could be more descriptive. Will update.

PR #1296, opened 2024-03-26
[Jetcaster]: Podcast screen

arriolac:
Changes:
* Added podcast details screen. 
* Added external models `PodcastInfo`, `EpisodeInfo` and `CategoryInfo` so that composables don't depend on database entities (`Podcast`, `Episode` and `Category) directly.

Note that design is not final yet but will address that in another PR.

image

### DIFF
--- a/Jetcaster/app/src/main/java/com/example/jetcaster/ui/podcast/PodcastDetailsScreen.kt
+++ b/Jetcaster/app/src/main/java/com/example/jetcaster/ui/podcast/PodcastDetailsScreen.kt
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.example.jetcaster.ui.podcast
 
 import androidx.compose.foundation.background
--- a/Jetcaster/app/src/main/java/com/example/jetcaster/ui/podcast/PodcastDetailsScreen.kt
+++ b/Jetcaster/app/src/main/java/com/example/jetcaster/ui/podcast/PodcastDetailsScreen.kt
@@ -154,7 +154,7 @@ fun PodcastDetailsContent(
             EpisodeListItem(
                 episode = episode,
                 podcast = podcast,
-                onClick = { navigateToPlayer(episode) },
+                onClick = navigateToPlayer,
                 onQueueEpisode = onQueueEpisode,
                 modifier = Modifier.fillMaxWidth(),
                 showPodcastImage = false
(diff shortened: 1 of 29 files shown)


Jetcaster/app/src/main/java/com/example/jetcaster/ui/podcast/PodcastDetailsScreen.kt
```diff
@@ -0,0 +1,336 @@
...
+        Row(
+            verticalAlignment = Alignment.Bottom,
+            modifier = Modifier.fillMaxWidth()
+        ) {
+            AsyncImage(
+                model = ImageRequest.Builder(LocalContext.current)
+                    .data(podcast.imageUrl)
+                    .crossfade(true)
+                    .build(),
+                contentDescription = stringResource(id = R.string.cd_podcast_image),
```
IanGClifton, line 184: This isn't useful as a content description. If one isn't available from the model, then I'd prefer null here so that screen readers can use the title below instead of repeating a generic image description.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/podcast/PodcastDetailsScreen.kt
```diff
@@ -0,0 +1,336 @@
...
+
+@Composable
+fun PodcastDetailsHeaderItemButtons(
+    isSubscribed: Boolean,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier
+) {
+    Row(modifier.padding(top = 16.dp)) {
+        Button(
+            onClick = onClick
```
IanGClifton, line 263: You should double-check this but I think this is going to result in TalkBack repeating itself. Since the icon and the text convey the same info, it's probably better to use null for the content description of the icon and maybe merge the semantics at the button.

IanGClifton (changes requested):
Now that there are models for the UI, I'd update PreviewData to use those directly. For example, it's a little weird to have 

```
val PreviewEpisodeToPodcasts = listOf(
    EpisodeToPodcast().apply {
        episode = PreviewEpisodes.first()
        _podcasts = PreviewPodcasts
    }
)
```

and then in the preview itself, it uses asPodcastCategoryEpisode to convert everything. I think PreviewData shouldn't need com.example.jetcaster.core.data.database.* imports anymore.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/podcast/PodcastDetailsScreen.kt
arriolac: Thanks for the a11y call out. `null` makes sense here since it's a decorative element.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/podcast/PodcastDetailsScreen.kt
arriolac: Makes sense to set the icon to `null` and have the Button merge semantics.

arriolac:
> Now that there are models for the UI, I'd update PreviewData to use those directly. For example, it's a little weird to have
> 
> ```
> val PreviewEpisodeToPodcasts = listOf(
>     EpisodeToPodcast().apply {
>         episode = PreviewEpisodes.first()
>         _podcasts = PreviewPodcasts
>     }
> )
> ```
> 
> and then in the preview itself, it uses asPodcastCategoryEpisode to convert everything. I think PreviewData shouldn't need com.example.jetcaster.core.data.database.* imports anymore.

Done in 1c94c881

arriolac:
This is ready for review again. PTAL @IanGClifton when you have a moment.

PR #1298, opened 2024-03-28
Add search functionality to Jetcaster

chikoski:
This pull request adds search functionality for Jetcaster. More specifically, the following changes are applied:

- Add methods implementing podcast search feature to PodcastDao and PodcastStore
- Add the search screen to the TV app

### DIFF
--- a/Jetcaster/tv-app/src/main/java/com/example/jetcaster/tv/ui/search/SearchScreenViewModel.kt
+++ b/Jetcaster/tv-app/src/main/java/com/example/jetcaster/tv/ui/search/SearchScreenViewModel.kt
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


kul3r4 (approved):

Jetcaster/tv-app/src/main/java/com/example/jetcaster/tv/ui/search/SearchScreenViewModel.kt
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private val searchConditionFlow =
+        combine(keywordFlow, selectedCategoryListFlow) { keyword, selectedCategories ->
+            SearchCondition(keyword, selectedCategories)
+        }
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    private val searchResultFlow = searchConditionFlow.flatMapLatest {
+        podcastStore.searchPodcastByTitleAndCategories(it.keyword, it.selectedCategories)
```
arriolac, line 59: I believe this will still trigger a search when the keyword is the empty string. Do we want that (i.e. still show results if there is no input)? Or should the empty case be handled another way?

arriolac (approved):
LGTM

Jetcaster/tv-app/src/main/java/com/example/jetcaster/tv/ui/search/SearchScreenViewModel.kt
chikoski: It is an intended behavior to trigger the search with empty string. I expect that all podcasts belonging to the selected category show up when the empty string is input.

PR #1303, opened 2024-03-29
[Jetcaster] Adding SupportingPaneScaffold to Home

arriolac:
Adding `SupportingPaneScaffold` to the home screen of Jetcaster.

Remaining TODOs:
- [ ] Make expanded-width default to having the supporting pane be hidden
- [ ] Hide top app bar in supporting pane when 2 panes are shown

[Screen_recording_20240329_162553.webm](https://github.com/android/compose-samples/assets/463186/a5b185d3-de84-4962-a01f-d4976021dcfa)

### DIFF
--- a/Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
+++ b/Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
@@ -103,17 +103,16 @@ import com.example.jetcaster.ui.theme.JetcasterTheme
 import com.example.jetcaster.util.ToggleFollowPodcastIconButton
 import com.example.jetcaster.util.quantityStringResource
 import com.example.jetcaster.util.verticalGradientScrim
-import java.time.Duration
-import java.time.LocalDateTime
-import java.time.OffsetDateTime
 import kotlinx.collections.immutable.PersistentList
 import kotlinx.collections.immutable.toPersistentList
 import kotlinx.coroutines.launch
+import java.time.Duration
+import java.time.LocalDateTime
+import java.time.OffsetDateTime
 
 @OptIn(ExperimentalMaterial3AdaptiveApi::class)
 @Composable
 fun Home(
-    navigateToPodcastDetails: (PodcastInfo) -> Unit,
     navigateToPlayer: (EpisodeInfo) -> Unit,
     viewModel: HomeViewModel = viewModel()
 ) {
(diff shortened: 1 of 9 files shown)


Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -104,31 +112,68 @@ import kotlinx.collections.immutable.PersistentList
...
-            onHomeCategorySelected = viewModel::onHomeCategorySelected,
-            onCategorySelected = viewModel::onCategorySelected,
-            onPodcastUnfollowed = viewModel::onPodcastUnfollowed,
-            navigateToPodcastDetails = navigateToPodcastDetails,
-            onTogglePodcastFollowed = viewModel::onTogglePodcastFollowed,
-            onLibraryPodcastSelected = viewModel::onLibraryPodcastSelected,
-            onQueueEpisode = viewModel::onQueueEpisode,
-            modifier = Modifier.fillMaxSize()
-        )
+    val navigator = rememberSupportingPaneScaffoldNavigator<String>(
```
florina-muntenescu, line 123: All of these changes should be in the other Home composable, rather than here.

florina-muntenescu:
it looks like there's a padding between the 2 panes. can we remove that?

arriolac:
> it looks like there's a padding between the 2 panes. can we remove that?

Yep, I noticed that as well. Will remove

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
arriolac: I see you opened #1305. Let's address this issue in your PR

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -103,30 +111,66 @@ import kotlinx.collections.immutable.PersistentList
 import kotlinx.collections.immutable.toPersistentList
 import kotlinx.coroutines.launch
 
+@OptIn(ExperimentalMaterial3AdaptiveApi::class)
 @Composable
 fun Home(
     navigateToPodcastDetails: (PodcastInfo) -> Unit,
```
jdkoren, line 117: Studio says this parameter is no longer used

jdkoren (approved):
On tablet in landscape, I see a back arrow icon in the supporting pane after tapping on a podcast. Tapping this icon does nothing while in landscape, so it would be good to remove it -- happy to do that in a follow up PR if you prefer

arriolac:
> On tablet in landscape, I see a back arrow icon in the supporting pane after tapping on a podcast. Tapping this icon does nothing while in landscape, so it would be good to remove it -- happy to do that in a follow up PR if you prefer

Yeah another PR would be great. I think the top bar should be hidden altogether when showing two panes

PR #1304, opened 2024-04-01
Add a screen to show the details of the selected episode

chikoski:
This pull request adds a screen to show the details of the selected episode. 

 In addition to that, some improvement on focus management is also included in this pull request as follows:

- Set focusRestorer to the following components:
  - PodcastRow
  - EpisodeRow
- Change the tabRow in the discover screen to get the current tab selected when d-pad focus enters to the row
- Change the Library screen to get the first podcast in the PodcastRow selected when users navigate to the screen


arriolac (approved):
LGTM.

Note that PR #1296 introduces new UI models. You may want to consider using those once that PR is merged so that your UI doesn't depend directly on database models. No action needed, more of an FYI

PR #1307, opened 2024-04-02
Fix: a build error happening on TV app module

chikoski:
This PR fixes a build issue happening on tv-app module.  The issue was cased by the resolution failure of BasicTextField2.

More specifically the following change happen by this PR:

- Switching from BasicTextFiled2 to BasicTextField
- Modification on DiscoverScreenViewModel to remove line breaks causing a layout glitch
- Passing supportingContent parameter to ListItem instead of putting everything in the content parameter


arriolac (approved):
LGTM

PR #1309, opened 2024-04-02
[Jetcaster] Handling no podcast selected and extracting state for home

florina-muntenescu:


Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -123,14 +149,42 @@ fun Home(
     BackHandler(enabled = navigator.canNavigateBack()) {
         navigator.navigateBack()
     }
+
+    val homeState = HomeState(
+        windowSizeClass = windowSizeClass,
```
mlykotom, line 154: NIT: could be simplified by having just `isCompact: Boolean`, which would also potentially recompose less when changing the window size

PR #1311, opened 2024-04-02
[Jetcaster] Add Hilt to app, tv and wear modules

arriolac:
Added Hilt to Jetcaster app module. Note that `Graph` is still in the codebase since TV and Wear still depend on it - both should also be migrated to use Hilt.


Jetcaster/core/src/main/java/com/example/jetcaster/core/data/di/Graph.kt
jdkoren, line 1: Is Graph still used anywhere, or can it be removed?

arriolac: TV and Wear still depend on it so those modules would need to be migrated to also use Hilt.

FYI @chikoski @kul3r4

chikoski: Thanks for the heads up! Actually, it is really helpful for me!

arriolac: Went ahead and updated TV and Wear to use Hilt and I also removed Graph.kt in the process [176e656](https://github.com/android/compose-samples/pull/1311/commits/176e65681112ca6a20778d89f13ea22aeae8a7a8)

PR #1314, opened 2024-04-04
[Jetsnack] Adding shared elements to Jetsnack 🎉

riggaroo:
Adding shared elements between Home feed and SnackDetail.

Introduced `SnackSharedElementKey` to create a unique key for the shared element to match. 

Using `Modifier.sharedBounds()` on the Card container, and `Modifier.sharedElement()` on the Image. 

Using `Modifier.sharedBounds()` with `Modifier.animateEnterExit()` on Text. 

The above modifiers require being in an `AnimatedVisibilityScope` and `SharedTransitionScope`, in order to use the correct scope at the correct point, we save the scopes in a CompositionLocal and retrieve it when required. 

https://github.com/android/compose-samples/assets/9973046/2d0cf1d9-4ef2-4051-a2e9-b3255d45ac78


Jetsnack/app/src/main/java/com/example/jetsnack/model/Snack.kt
```diff
@@ -41,183 +42,183 @@ val snacks = listOf(
         price = 299
     ),
     Snack(
-        id = 2L,
+        id = Random.nextLong(),
```
jdkoren, line 45: Cupcake above still has a fixed ID. Do these IDs need to be random? Also, isn't it now possible (albeit very improbable) that two snacks end up with the same ID?

riggaroo: Its for preview purposes to be able to query for a certain snack. yes it is somewhat possible, but this is fake mock data so I dont think its worth the hassle of making it more robust.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Snacks.kt
```diff
@@ -186,101 +212,261 @@ fun SnackItem(
...
+                        .sharedBounds(
+                            rememberSharedContentState(
+                                key = SnackSharedElementKey(
+                                    snackId = snack.id,
+                                    origin = snackCollectionId.toString(),
+                                    type = SnackSharedElementType.Image
+                                )
+                            ),
+                            animatedVisibilityScope = animatedVisibilityScope,
+                            boundsTransform = snackDetailBoundsTransform
```
yongsuk44, line 246: I have a question. 🤔

I think `sharedBounds()` was used in `SnackItem` to maintain the visual continuity of space and size, as it involves both an image and text. If there were only an image, would it be appropriate to use `sharedElement()` instead?

riggaroo: It depends, if you have an Async image like this in this case, you'd want to use shared bounds as you may be loading up a higher resolution image.  If you are using a standard image that you dont expect to be different, then yes you can use sharedElement :)

yongsuk44: May I interpret it such that sharedBounds is used when the resolution of a component is determined dynamically, and sharedElement is employed when it's static? 

Additionally, as I have a keen interest in this topic, would you kindly recommend any references I could consult?

riggaroo: So if you were to apply some other transformations to the image, for example a black and white filter, you'd need to use sharedBounds.

yongsuk44: @riggaroo Thank you for your assistance. 😀
I understand the difference.

yongsuk44:
Will values be added to the 'element' tab later, or is it normal for it to be empty. In my opinion, it would be convenient for debugging to display the values applied to parameters other than the key in sharedContentState.

스크린샷 2024-05-21 오후 5 34 40

JolandaVerhoef (approved):

PR #1321, opened 2024-04-05
 Show single pane on phone in landscape

jdkoren:


Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -142,6 +154,69 @@ private fun <T> ThreePaneScaffoldNavigator<T>.isMainPaneHidden(): Boolean {
...
+private fun getExcludedVerticalBounds(posture: Posture, hingePolicy: HingePolicy): List<Rect> {
+    return when (hingePolicy) {
+        HingePolicy.AvoidSeparating -> posture.separatingVerticalHingeBounds
+        HingePolicy.AvoidOccluding -> posture.occludingVerticalHingeBounds
+        HingePolicy.AlwaysAvoid -> posture.allVerticalHingeBounds
+        else -> emptyList()
+    }
+}
+
+private fun androidx.window.core.layout.WindowSizeClass.isCompact(): Boolean =
```
jdkoren, line 216: There are currently two different `WindowSizeClass` symbols, one in `androidx-window` (used by the `adaptive` deps), and one in `material3-window-size-class`. I think things are moving towards the `andriodx-window` one, and I have a mind to migrate all usages of the other, but not sure if this PR is the right place for that.

arriolac: Thanks for calling that out. Yeah we can make that migration on a separate PR

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -142,6 +154,69 @@ private fun <T> ThreePaneScaffoldNavigator<T>.isMainPaneHidden(): Boolean {
     return scaffoldValue[SupportingPaneScaffoldRole.Main] == PaneAdaptedValue.Hidden
 }
 
+@OptIn(ExperimentalMaterial3AdaptiveApi::class)
+fun calculateScaffoldDirective(
```
arriolac, line 158: Would be good to add a comment on how this differs from the default `calculateScaffoldDirective`

arriolac (approved):

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
jdkoren: Done

PR #1324, opened 2024-04-08
Add the player screen to TV app

chikoski:
This pull request adds the player screen to the TV app. In related to that, EpisodePlayer interface is updated so that EpisodePlayer implementations can play the given PlayerEpisode.

### DIFF
--- a/Jetcaster/tv-app/src/main/java/com/example/jetcaster/tv/ui/component/Background.kt
+++ b/Jetcaster/tv-app/src/main/java/com/example/jetcaster/tv/ui/component/Background.kt
@@ -91,7 +91,7 @@ internal fun Background(
 }
 
 @Composable
-internal fun WithBackground(
+internal fun BackgroundContainer(
     playerEpisode: PlayerEpisode,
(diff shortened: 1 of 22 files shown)


Jetcaster/tv-app/src/main/java/com/example/jetcaster/tv/ui/component/Background.kt
```diff
@@ -56,3 +89,24 @@ internal fun Background(
             }
     )
 }
+
+@Composable
+internal fun WithBackground(
```
arriolac, line 94: This should be named with a noun per our naming API guidance: https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#naming-unit-composable-functions-as-entities

How about `EpisodeBackground` or `EpisodeContainer`?

arriolac (commented):
Please also make sure to indicate in a comment that the player screen is mocked and not actually playing any media.

arriolac:
#1326 should fix the build failure

arriolac (approved):

PR #1325, opened 2024-04-08
[Wear] Enable play/pause on the Player

kul3r4:
Adds logic for enabling the button play/pause.

### DIFF
--- a/Jetcaster/wear/src/main/java/com/example/jetcaster/ui/player/PlayerScreen.kt
+++ b/Jetcaster/wear/src/main/java/com/example/jetcaster/ui/player/PlayerScreen.kt
@@ -37,16 +37,18 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalView
+import androidx.compose.ui.res.stringResource
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.wear.compose.material.MaterialTheme
+import com.example.jetcaster.R
 import com.google.android.horologist.audio.ui.VolumeUiState
 import com.google.android.horologist.audio.ui.VolumeViewModel
 import com.google.android.horologist.audio.ui.rotaryVolumeControlsWithFocus
 import com.google.android.horologist.compose.rotaryinput.RotaryDefaults
 import com.google.android.horologist.media.ui.components.PodcastControlButtons
 import com.google.android.horologist.media.ui.components.background.ArtworkColorBackground
-import com.google.android.horologist.media.ui.components.display.LoadingMediaDisplay
+import com.google.android.horologist.media.ui.components.controls.SeekButtonIncrement
 import com.google.android.horologist.media.ui.components.display.TextMediaDisplay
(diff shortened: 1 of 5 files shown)


Jetcaster/wear/src/main/java/com/example/jetcaster/ui/player/PlayerScreen.kt
```diff
@@ -80,10 +83,20 @@ private fun PlayerScreen(
     PlayerScreen(
         mediaDisplay = {
             if (playerUiState != null) {
-                playerUiState.episodePlayerState.currentEpisode?.let {
+                if (playerUiState.episodePlayerState.currentEpisode != null) {
```
arriolac, line 86: 2 null checks are being done here. You can simplify this to:

```
val episode = playerUiState.episodePlayerState.currentEpisode
if (episode != null) {
     TextMediaDisplay(
         title = episode.podcastName,
         subtitle = episode.title
     )
} else {
    TextMediaDisplay(
          title = it.podcastName,
          subtitle = it.title
          title = "Nothing to play",
          subtitle = ""
      )
}
```

Jetcaster/wear/src/main/java/com/example/jetcaster/ui/player/PlayerScreen.kt
```diff
@@ -80,10 +83,20 @@ private fun PlayerScreen(
...
+                                title = episode.podcastName,
+                                subtitle = episode.title
+                            )
+                        }
+                    }
+                } else {
                     TextMediaDisplay(
-                        title = it.podcastName,
-                        subtitle = it.title
+                        title = "Nothing to play",
```
arriolac, line 98: This string should be added to `strings.xml` and read from `stringResource`

Jetcaster/wear/src/main/java/com/example/jetcaster/ui/player/PlayerScreen.kt
kul3r4: In my check I am also checking that the tile is not "" as I want to show "Nothing to play" instead (data class PlayerEpisode sets the title to "")

arriolac: How about?

```
val episode = playerUiState.episodePlayerState.currentEpisode
if (episode != null && episode.title.isNotEmpty()) {
     TextMediaDisplay(
         title = episode.podcastName,
         subtitle = episode.title
     )
} else {
    TextMediaDisplay(
          title = it.podcastName,
          subtitle = it.title
          title = "Nothing to play",
          subtitle = ""
      )
}
```

Jetcaster/wear/src/main/java/com/example/jetcaster/ui/player/PlayerScreen.kt
```diff
@@ -72,6 +74,7 @@ fun PlayerScreen(
 @Composable
 private fun PlayerScreen(
     playerUiState: PlayerUiState?,
```
yschimke, line 76: PlayerUiState can be non nullable.

We might want to change the model to reflect no loaded item, rather then "" being no item.

kul3r4: thanks, removed the ?

PR #1335, opened 2024-04-10
Unify WindowSizeClass usage

jdkoren:


arriolac (commented):
Can we also remove the M3 WSC dependency in `build.gradle` and `libs.versions.toml` to prevent using the wrong one? 

Also please rebase against `jetcaster/all_form_factors` so that the build passes

jdkoren:
> Can we also remove the M3 WSC dependency in `build.gradle` and `libs.versions.toml` to prevent using the wrong one?

Done

arriolac (approved):

PR #1338, opened 2024-04-12
[Jetcaster] Add ItemCount and Current Position Indicator to FollowedPodcasts

yongsuk44:
image


arriolac:
Thanks for this! I know that it's not very clear that there are multiple pages in that screen that you can swipe through so a solution like this is desirable. We have been working on some new designs however in the `jetcaster/all_form_factors` branch which we intend to merge into `main` soon. We've intentionally left out a page indicator in favor of a design that display multiple album art:

portrait-phone

Going to close this PR since it will be merged into `main` soon.

yongsuk44:
@arriolac got it. Your work seems to look better. 😀

PR #1339, opened 2024-04-12
[Jetcaster] Handling of HTML Elements in PodcastSummary on Foldable Devices

yongsuk44:
There are cases where the Summary Text does not contain HTML elements. In these instances, it is considered that the performance overhead is negligible, therefore no additional processing was implemented

| before | after |
| ------- | -----|
| HTML-TAG|html-refactor|


arriolac:
Thanks! I'm aware of this issue but I think it would be good to use the new `parseAsHtml` method in 1.7.0-alpha06. Since Jetcaster doesn't depend on 1.7.0 _yet_, this is a good interim solution.

As mentioned in #1338, we have been working on `jetcaster/all_form_factors` as the soon-to-be state of `Jetcaster` on `main`. Would you mind changing the base branch of this PR to `jetcaster/all_form_factors` instead?

yongsuk44:
> Would you mind changing the base branch of this PR to jetcaster/all_form_factors instead?

@arriolac Done :)

yongsuk44:
sry, I found the conflicting parts and have made the corrections. 🥲

Jetcaster/app/src/main/java/com/example/jetcaster/ui/player/PlayerScreen.kt
```diff
@@ -618,32 +619,31 @@ private fun PodcastInformation(
     title: String,
     name: String,
     summary: String,
+    modifier: Modifier = Modifier,
     titleTextStyle: TextStyle = MaterialTheme.typography.headlineSmall,
     nameTextStyle: TextStyle = MaterialTheme.typography.displaySmall,
 ) {
     Column(
-        horizontalAlignment = Alignment.CenterHorizontally,
```
arriolac, line 625: This line should be reverted

yongsuk44: I see.

yongsuk44:
An error occurred in Jetchat's instrumentation test during the CI build earlier. 
Should I address it? 😅

PR #1340, opened 2024-04-12
[Jetcaster] Update player styling

arriolac:
Changes:
* Added tests for `MockEpisodePlayer`
* Update styling to match design in `PlayerScreen`
* Supporting adding to queue from `PlayerScreen`


Jetcaster/app/src/main/java/com/example/jetcaster/ui/player/PlayerViewModel.kt
```diff
@@ -99,4 +100,14 @@ class PlayerViewModel @Inject constructor(
...
         episodePlayer.rewindBy(duration)
     }
+
+    fun onAddToQueue() {
+        uiState.episodePlayerState.currentEpisode?.let {
+            episodePlayer.addToQueue(it)
+        }
+    }
+
+    fun onPlay(episode: PlayerEpisode) {
```
kul3r4, line 110: curious why in `addToQueue` you don't have a parameter and use the `uiState.episodePlayerState.currentEpisode` while in `onPlay` you pass a parameter.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/player/PlayerViewModel.kt
```diff
@@ -99,4 +100,14 @@ class PlayerViewModel @Inject constructor(
     fun onRewindBy(duration: Duration) {
         episodePlayer.rewindBy(duration)
     }
+
+    fun onAddToQueue() {
+        uiState.episodePlayerState.currentEpisode?.let {
+            episodePlayer.addToQueue(it)
```
kul3r4, line 106: are we adding every time the user taps the button? (just clarifying if you were going to add some logic on the episodePlayer for throttling or keep it as it is)

Jetcaster/app/src/main/java/com/example/jetcaster/ui/player/PlayerViewModel.kt
arriolac: Context: I was experimenting with the `MockEpisodePlayer#play(PlayerEpisode)` API to support swiping through episodes. But, I ended up not adding that feature to this PR since there are a couple of tricky edge cases. For that feature, you need to explicitly specify which episode you would like to play whereas the no-arg `onPlay` method is implicit (i.e. you want to play the current episode). 

For clarity, I can rename this to `onEpisodeChanged` or something like that. However, since this method is not used yet I'll remove it.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/player/PlayerViewModel.kt
arriolac: Yep, we are adding the same episode to the queue as the user taps and there's no throttling/limiting behavior at the moment. If we want that we can add that logic in the `EpisodePlayer` but I would say that's out-of-scope for now.

Jetcaster/core/src/test/kotlin/com/example/jetcaster/core/player/MockEpisodePlayerTest.kt
```diff
@@ -0,0 +1,140 @@
...
+        mockEpisodePlayer.currentEpisode = PlayerEpisode(
+            uri = "currentEpisode",
+            duration = Duration.ofSeconds(60)
+        )
+        testEpisodes.forEach { mockEpisodePlayer.addToQueue(it) }
+
+        mockEpisodePlayer.play()
+        advanceTimeBy(100)
+
+        // TODO override next?
```
jdkoren, line 120: Is this comment needed?

Jetcaster/core/src/test/kotlin/com/example/jetcaster/core/player/MockEpisodePlayerTest.kt
```diff
@@ -0,0 +1,140 @@
...
+        testEpisodes.forEach { mockEpisodePlayer.addToQueue(it) }
+
+        mockEpisodePlayer.play()
+        advanceTimeBy(duration.toMillis() + 1)
+
+        assertEquals(testEpisodes.first(), mockEpisodePlayer.currentEpisode)
+    }
+
+    @Test
+    fun whenNextQueueEmpty_doesNothing() {
```
jdkoren, line 66: Minor nit about naming: this combines the "given" and "when" of the test, which reads a little strange. I would separate them with another underscore, e.g. `whenNext_queueIsEmpty_doesNothing()`.

Same for other tests in this file.

jdkoren (changes requested):

Jetcaster/core/src/test/kotlin/com/example/jetcaster/core/player/MockEpisodePlayerTest.kt
arriolac: Ahh yeah that's much better. Will update

Jetcaster/core/src/test/kotlin/com/example/jetcaster/core/player/MockEpisodePlayerTest.kt
arriolac: Remove

bentrengrove (approved):

chikoski (approved):

PR #1342, opened 2024-04-15
[All] Add plugins to root gradle

mlykotom:
Ading the plugins to the root level build.gradle without applying them directly. This is because some plugins might have different versions. For example, Baseline Profile Gradle Plugin had a dependency on AGP x, without specifying it here, Gradle would pick some dependency (which one? 🤔 ) and a conflict would be raised.


arriolac (approved):

bentrengrove:
Also important for #1354

ThRiLKiLKuLt (commented):
Should've went to  college

PR #1345, opened 2024-04-15
[Jetcaster] Merge `jetcaster/all_form_factors` into `main`

arriolac:
The new Jetcaster sample now demonstrates how to build an app with Compose such that it supports the following form factors: phones, tablets, foldables, TV and Wear.

readme_cover


Jetcaster/README.md
```diff
@@ -3,60 +3,34 @@
...
-### Status: 🚧 In progress 🚧
-
-Jetcaster is still in the early stages of development, and as such only one screen has been created so far. However,
-most of the app's architecture has been implemented, as well as the data layer, and early stages of dynamic theming.
-
-
 ## Screenshots
 
-<img src="docs/screenshots.png"/>
+![readme_cover](https://github.com/android/compose-samples/assets/10263978/a58ab950-71aa-48e0-8bc7-85443a1b4f6b)
```
kul3r4, line 16: reminder that the Wear one does not have the download function right now, I can upload some screenshots if you like?

Jetcaster/README.md
```diff
@@ -3,60 +3,34 @@
...
 
 ## Features
 
-This sample contains 2 screens so far: the home screen, and a player screen.
+This sample has 3 components: the home screen, the podcast details screen, and the player screen
```
kul3r4, line 20: We could potentially have different features for form factors, e.g. Wear does not deal with multiple panels. We can do in another PR.

kul3r4 (approved):

jdkoren (approved):

Jetcaster/README.md
arriolac: Yes, we can describe the differences of each form factor in more detail. I agree that we should tackle that in a separate PR.

Jetcaster/README.md
arriolac: Good call out. I don't think we'll be adding that feature at all, yeah? If so, we can update the screenshot to reflect the actual implementation on Wear.

chikoski (approved):

bentrengrove (approved):

PR #1348, opened 2024-04-17
[Jetcaster] Add functionality to change the speed of the player

arriolac:
So that the player speed can be changed on Wear.

### DIFF
--- a/Jetcaster/core/src/main/java/com/example/jetcaster/core/player/EpisodePlayer.kt
+++ b/Jetcaster/core/src/main/java/com/example/jetcaster/core/player/EpisodePlayer.kt
@@ -17,8 +17,8 @@
 package com.example.jetcaster.core.player
 
 import com.example.jetcaster.core.model.PlayerEpisode
-import kotlinx.coroutines.flow.StateFlow
 import java.time.Duration
+import kotlinx.coroutines.flow.StateFlow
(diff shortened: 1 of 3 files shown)


kul3r4 (approved):

Jetcaster/core/src/main/java/com/example/jetcaster/core/player/EpisodePlayer.kt
```diff
@@ -85,4 +85,9 @@ interface EpisodePlayer {
      * Rewinds a currently played episode by a given time interval specified in [duration].
      */
     fun rewindBy(duration: Duration)
+
+    /**
+     * Changes the speed of which the player increments
+     */
+    fun changePlaySpeed(duration: Duration)
```
bentrengrove, line 92: Total nit, this should be `setPlayerSpeed` and probably could just be a public var

bentrengrove (approved):
Minor nit

Jetcaster/core/src/main/java/com/example/jetcaster/core/player/EpisodePlayer.kt
arriolac: Definitely much better. Fixed in fbe75e28

PR #1349, opened 2024-04-17
[Jetcaster]: Suggest podcasts in 'Technology' category for Wear app

arriolac:
Changes:
* Show podcasts to subscribe to in the 'Technology' category when first launching Wear app
* Fix ADB and NiA feed URLs which fixes the issue of NiA and ADB not being fetched correctly on TV and Wear

### DIFF
--- a/Jetcaster/core/src/main/java/com/example/jetcaster/core/data/network/Feeds.kt
+++ b/Jetcaster/core/src/main/java/com/example/jetcaster/core/data/network/Feeds.kt
@@ -20,14 +20,18 @@ package com.example.jetcaster.core.data.network
  * A hand selected list of feeds URLs used for the purposes of displaying real information
  * in this sample app.
  */
+const val NowInAndroid = "https://feeds.libsyn.com/244409/rss"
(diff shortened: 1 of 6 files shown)


kul3r4 (approved):
thank you!

Jetcaster/core/src/main/java/com/example/jetcaster/core/data/network/Feeds.kt
```diff
@@ -20,14 +20,18 @@ package com.example.jetcaster.core.data.network
  * A hand selected list of feeds URLs used for the purposes of displaying real information
  * in this sample app.
  */
+const val NowInAndroid = "https://feeds.libsyn.com/244409/rss"
+const val AndroidDevelopersBackstage =
```
bentrengrove, line 24: Should be private?
```suggestion
private const val NowInAndroid = "https://feeds.libsyn.com/244409/rss"
private const val AndroidDevelopersBackstage =
```

bentrengrove (approved):

PR #1353, opened 2024-04-19
Minor UI tweak on Jetcaster/tv-app

chikoski:
This pull request aims to polish UI of the Jetcaster TV app.  More specifically, the following changes are applied:

- Switch app icon from default one to the proper Jetcaster icon
- Set a placeholder for thumbnails in case thumbnail is not available
- Change the search screen behavior to be consistent with the expected filter chip behavior


Jetcaster/tv-app/src/main/java/com/example/jetcaster/tv/ui/component/EpisodeCard.kt
```diff
@@ -87,7 +78,12 @@ private fun EpisodeThumbnail(
         scale = CardScale.None,
         modifier = modifier,
     ) {
-        AsyncImage(model = playerEpisode.podcastImageUrl, contentDescription = null)
+        AsyncImage(
```
arriolac, line 81: Thoughts about using `PodcastImage` in `designsystem`? Would be nice to reuse the same component across form factors. (side-note though: I do plan on update the loading indication from a circular progress to a shimmer effect)

Jetcaster/tv-app/src/main/java/com/example/jetcaster/tv/ui/component/thumbnailPlaceholder.kt
```diff
@@ -0,0 +1,32 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.SolidColor
+import androidx.compose.ui.graphics.painter.BrushPainter
+import androidx.tv.material3.ExperimentalTvMaterial3Api
+import androidx.tv.material3.MaterialTheme
+
+@OptIn(ExperimentalTvMaterial3Api::class)
+@Composable
+internal fun thumbnailPlaceholder(
```
arriolac, line 28: Thoughts about adding this to `designsystem` module?

arriolac (approved):
Change LGTM. I just have a couple of suggestions on how we can improve reusability across other form factors—image loading behavior specifically seems like something we can share.

Jetcaster/tv-app/src/main/java/com/example/jetcaster/tv/model/CategoryInfoList.kt
```diff
@@ -18,6 +18,27 @@ package com.example.jetcaster.tv.model
...
+            return CategoryInfoList(member)
+        }
+    }
+}
+
+private fun CategoryInfo.intoCategory(): Category {
+    return Category(id, name)
+}
+
+private fun Category.intoCategoryInfo(): CategoryInfo {
```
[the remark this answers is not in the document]
chikoski: Thanks! I was not aware of the method. Updated with it and removed the `intoCategoryInfo`.

Jetcaster/tv-app/src/main/java/com/example/jetcaster/tv/ui/component/thumbnailPlaceholder.kt
chikoski: I would, and it would be tricky as it refers to the `MaterialTheme` object from `tv-material` library. 

There are three possible options for this:

1. Leave the component in `tv-app` module
2. Remove the default parameter value from the function
3. Change the default parameter value to Color.Transparent from MaterialTheme.colorScheme.backgroundVariant

I would take the 3rd option to move the function to the `design` module. What do you think?

Jetcaster/tv-app/src/main/java/com/example/jetcaster/tv/ui/component/EpisodeCard.kt
chikoski: I would like to switch to `PodcastImage` from `AsyncImage`, and it is blocked as `PodcastImage` depends on `MaterialTheme` object from `androidx.compose.material3`, which is not a part of `tv-material` library. 

I believe we should remove the dependencies to the mobile specific components from the `designsystem` module, but It would need more discussion. I would defer it to another issue. So, I would keep current implementation in this pull request.

PR #1354, opened 2024-04-19
[All] Migrate to the Compose gradle plugin

bentrengrove:
Migrate all samples to the compose gradle plugin and Kotlin 2.0

Also enable strong skipping mode on all samples

### DIFF
--- a/JetLagged/build.gradle.kts
+++ b/JetLagged/build.gradle.kts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 The Android Open Source Project
+ * Copyright 2024 The Android Open Source Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/JetLagged/build.gradle.kts
+++ b/JetLagged/build.gradle.kts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 The Android Open Source Project
+ * Copyright 2020 The Android Open Source Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,10 +17,10 @@
 plugins {
     alias(libs.plugins.gradle.versions)
     alias(libs.plugins.version.catalog.update)
-
     alias(libs.plugins.android.application) apply false
     alias(libs.plugins.kotlin.android) apply false
     alias(libs.plugins.kotlin.parcelize) apply false
+    alias(libs.plugins.compose) apply false
 }
 
 apply("${project.rootDir}/buildscripts/toml-updater-config.gradle")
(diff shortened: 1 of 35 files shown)


Jetsnack/app/build.gradle.kts
```diff
@@ -86,6 +83,14 @@ android {
     }
 }
 
+composeCompiler {
+    // Not required in final Kotlin 2.0.0 release
+    suppressKotlinVersionCompatibilityCheck = "2.0.0-RC1"
```
CoreFloDev, line 88: That can be removed if Kotlin 2.0.0-RC2 is used

chris-hatton: Trying to use `2.0.0-RC2` results in:
```
Compose Multiplatform 1.6.2 doesn't support Kotlin 2.0.0-RC2. Please see https://github.com/JetBrains/compose-jb/blob/master/VERSIONING.md#kotlin-compatibility to know the latest supported version of Kotlin.
```
...where the linked page only references up to `2.0.0-RC1`.  Appreciate this is very much the bleeding edge of compose but it would be nice if the tables could be kept up-to-date with the RC's.

bentrengrove: We don't control the Jetbrains pages unfortunately. I'm updating this PR now for Kotlin 2.0 though which has been released

JetLagged/build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 The Android Open Source Project
+ * Copyright 2020 The Android Open Source Project
```
sabufung30, line 2: Isn't this should be 2024

jdkoren (approved):

gastsail:
I have a question that I cannot find anywhere, is there anything we need to take care about regarding BOM and compose versions?

This is because now that the compiler comes with kotlin 2.0 I wonder if previous versions of BOM or compose components will crash since we are not specifying the right compiler version.

Thanks

bentrengrove:
Compose library versions and Compose compiler versions are not linked, you can mix and match them. There is no need to match Compose 1.5.1 with Compiler 1.5.1

See this blog post for more details
https://android-developers.googleblog.com/2022/06/independent-versioning-of-Jetpack-Compose-libraries.html

gastsail:
> Compose library versions and Compose compiler versions are not linked, you can mix and match them. There is no need to match Compose 1.5.1 with Compiler 1.5.1
> 
> 
> 
> See this blog post for more details
> 
> https://android-developers.googleblog.com/2022/06/independent-versioning-of-Jetpack-Compose-libraries.html

Thanks a lot !

zeesibanda08-max:
NYC

PR #1355, opened 2024-04-19
[Jetcaster] Add tests for podcast category filter limit

yongsuk44:
Although this test case may become unnecessary once the `limit` parameters in each method of the `CategoryStore` are explicitly specified, I added it because it seems that there was no test code to verify the limits currently used in the `PodcastCategoryFilterUseCase`.


arriolac (approved):
Thanks!

PR #1358, opened 2024-04-24
Fix Compose Previews in the Reply sample

amaurymedeiros:
Currenly, Compose Previews don't support UIModeManager, so any previews using the ContrastAwareReplyTheme will fail to render when trying to apply contrast. While UIModeManager support is not added, we can work around this issue by not adding contrast in Compose Previews.


jwill:
Temp fix LGTM and is in the spirit of the original code I wrote. 
(Am not an official repo owner but hoping to speed up the process for whoever reviews this PR.)

bentrengrove (approved):

bentrengrove:
Needs a rebase on top of #1364 I think

arriolac:
Just merged #1364. Feel free to rebase against `main` instead to resolve the failed actions.

amaurymedeiros:
Rebase is done and checks have passed.

PR #1359, opened 2024-04-24
Refactor home to library

kul3r4:
- Refactor home to use states similar to other screens
- Removed Dialog for when the content is not available in Library as it was causing problem to navigation when dismissing being in a pager. Moreover this avoids flashing problems which have been observed.


Jetcaster/wear/src/main/java/com/example/jetcaster/ui/library/LibraryScreen.kt
```diff
@@ -0,0 +1,198 @@
...
+        contentPadding = ScalingLazyColumnDefaults.padding(
+            first = ScalingLazyColumnDefaults.ItemType.Text,
+            last = ScalingLazyColumnDefaults.ItemType.Chip,
+        ),
+    )
+
+    when (val s = uiState) {
+        is LibraryScreenUiState.Loading -> LoadingScreen(columnState = columnState)
+        is LibraryScreenUiState.NoSubscribedPodcast ->
+            ScreenScaffold(scrollState = columnState, modifier = modifier) {
```
arriolac, line 74: Nit: This can be set as it's own separate composable (e.g. `NoSubscribedPodcast`) to improve readability. Same comment for the `Ready` state below.

Jetcaster/wear/src/main/java/com/example/jetcaster/ui/library/LibraryScreen.kt
```diff
@@ -0,0 +1,198 @@
...
+                            )
+                        }
+                    }
+                }
+            }
+    }
+}
+
+@Composable
+fun LoadingScreen(columnState: ScalingLazyColumnState) {
```
arriolac, line 155: This composable (and the ones below) should also accept a modifier.

kul3r4: thanks fixed

arriolac (approved):

PR #1360, opened 2024-04-25
Switching to PodcastImage

chikoski:
This pull request switch the component to show thumbnails in tv apps from `AsyncImage` into `PodcastImage`. As a side effect of the migration,  Podcast`, `PodcastWithExtraInfo` and `EpisodeToPodcast` are replaced with their external models.


Jetcaster/designsystem/src/main/java/com/example/jetcaster/designsystem/component/PodcastImage.kt
```diff
@@ -22,14 +22,14 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.size
 import androidx.compose.material3.CircularProgressIndicator
-import androidx.compose.material3.MaterialTheme
```
arriolac, line 25: Thank you!

arriolac (approved):

PR #1361, opened 2024-04-25
Remove unused launcher icons

ithinkihaveacat:
Among other reasons to think they're unused, they're default Android icons rather than the custom "Jetcaster" one.

FYI @kul3r4


Jetcaster/wear/src/main/res/mipmap-xxxhdpi/ic_launcher.png
kul3r4, line 1: @ithinkihaveacat this was actually a Jetcaster icon. Did you run the sample to check that this one in particualt is not used?

jdkoren: Min SDK for this sample is 26, so presumably the one in `mipmap-anydpi-v26` would always be used?

jdkoren (approved):
Wear module seems to build and run fine with this change.

PR #1362, opened 2024-04-25
Switch to using local debug.keystore in all samples. 

riggaroo:
Switching to local debug.keystore to prevent issues with installing debug app on device.


JetLagged/app/build.gradle.kts
```diff
@@ -34,9 +34,9 @@ android {
     }
 
     signingConfigs {
```
jdkoren, line 36: Maybe we just remove this block altogether? I think Android Studio already defaults to the local key with these parameters for debug builds.

mlykotom: Question is if the `benchmark` buildType allows it, but I assume there's default `signingConfig` even if it's not defined.

riggaroo: Benchmark requires release build, and release requires a certificate :-/ Not sure how else to work around this

simona-anomis (approved):

PR #1363, opened 2024-04-25
[Jetcaster] Handle error and loading states

arriolac:
* Adding loading/error states to Home
* Adding loading/error states to PodcastImage

image


Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -230,14 +231,73 @@ private fun getExcludedVerticalBounds(posture: Posture, hingePolicy: HingePolicy
...
+            )
+        }
+    }
+}
+
+@Composable
+private fun HomeScreenLoading(modifier: Modifier = Modifier) {
+    Surface(modifier.fillMaxSize()) {
+        Box {
+            CircularProgressIndicator(
```
kul3r4, line 259: can you use instead some "loading components"? Like a grid/column of few grey items showing that content is loading?

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -230,14 +231,73 @@ private fun getExcludedVerticalBounds(posture: Posture, hingePolicy: HingePolicy
...
+@Composable
+private fun HomeScreenError(onRetry: () -> Unit, modifier: Modifier = Modifier) {
+    Surface(modifier = modifier) {
+        Column(
+            verticalArrangement = Arrangement.Center,
+            horizontalAlignment = Alignment.CenterHorizontally,
+            modifier = Modifier.fillMaxSize(),
+        ) {
+            Text(
+                text = "An error has occurred.",
```
kul3r4, line 275: Should this live in strings.xml?

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/HomeViewModel.kt
```diff
@@ -179,13 +183,21 @@ enum class HomeCategory {
...
+        val featuredPodcasts: PersistentList<PodcastInfo> = persistentListOf(),
+        val selectedHomeCategory: HomeCategory = HomeCategory.Discover,
+        val homeCategories: List<HomeCategory> = emptyList(),
+        val filterableCategoriesModel: FilterableCategoriesModel =
+            FilterableCategoriesModel(),
+        val podcastCategoryFilterResult: PodcastCategoryFilterResult =
+            PodcastCategoryFilterResult(),
+        val library: LibraryInfo = LibraryInfo(),
+    ) : HomeScreenUiState
+}
```
mlykotom, line 203: I would advocate against structuring the state this way, because it makes quite hard to work with in general.

1. usually if you already have content and an error occurs, you can just show a snackbar/dialog with the error, you don't have to dismiss the whole existing content

2. Similarly for the loading state, if you'd like to have pull to refresh and you already have data loaded, it's hard to show the loading in form of some non-blocking UI. In the previous case you can simply check for `isRefreshing` and for `data.isEmpty()` to decide whether to show the "blocking" loading or just progress bar or something

3. The state that user can interact with in this case is implemented as a separate flow (`_selectedCategory`) so it's not as a problem, but it might make sense that this is part of the one `HomeViewState` in which case you can't really use the `copy()` function on the data classes, because your `_state` is the abstract sealed class.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
arriolac: Yes, it definitely should. Will fix

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/HomeViewModel.kt
arriolac: These are all good points. The primary motivation for making this change was to have consistency with state handling across all form factors (this is how TV and Wear are also handled). I do agree with the idea of still showing content that is available and showing error/loading as part of the "ready" state. 

With the renewed purpose of Jetcaster in mind, I would rank consistency across FF higher than accomplishing an the ideal UX. That said, I think we need to revisit this architecture due to the shortcomings that you had mentioned. My suggestion is that we continue with the approach in this change but file an issue to track this work as it is a larger change that affects mobile/TV/Wear. We can use the GH issue to agree on an architecture that makes sense.

Thoughts? FYI @chikoski @kul3r4

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/Home.kt
arriolac: I think this relates to Tomas's comments below about intermediate loading/error states. I made a suggestion there on how to improve this so let's use that thread to discuss.

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/HomeViewModel.kt
mlykotom: Yeah, I understand this is probably beyond this PR as I've seen it used in other places as well.
I agree with putting a though about this independently of this PR. 👍🏼

mlykotom (approved):

Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/HomeViewModel.kt
arriolac: Filed #1376 to track this work

PR #1365, opened 2024-04-25
[Jetcaster]: Display latest episodes for all subscribed podcasts in library.

arriolac:


Jetcaster/app/src/main/java/com/example/jetcaster/ui/home/HomeViewModel.kt
```diff
@@ -68,6 +69,8 @@ class HomeViewModel @Inject constructor(
     // Holds the view state if the UI is refreshing for new data
     private val refreshing = MutableStateFlow(false)
 
+    private val followedPodcasts = podcastStore.followedPodcastsSortedByLastEpisode(limit = 10)
```
mlykotom, line 72: By reusing this and having it twice in the `combine` call below, this means it will subscribe 2 streams of data: one for directly `followedPodcasts` and one for the `flatmapLatest`. I would suggest not doing this.

You can keep one stream of data by combining the actual value with the stream, for example:
```
                followedPodcasts.flatMapLatest { podcasts ->
                    combine(
                        flowOf(podcasts),
                        episodeStore.episodesInPodcasts(
                            podcastUris = podcasts.map { it.podcast.uri },
                            limit = 20
                        ),
                        ::Pair
                    )
                }
```

arriolac: Ah, I meant to use `shareIn` here. I prefer that approach so we're not nesting as much as is easier to follow/read. Any thoughts on either approach?

mlykotom: I think you're right that the `flatmapLatest` might be less readable.

I'm starting to think this `combine` flow should be optimized, because when you start the screen the combine flow will always call it 3x, which means it sets 3x state with different value and therefore the first time this whole screen recomposes 3x, which can be very janky (it actually seems like it is).

1. The `selectedHomeCategory` can cause inner loop with`selectedHomeCategory.value =if (podcasts.isEmpty()) HomeCategory.Discover else homeCategory`. 

2. Similar for `_selectedCategory`, which is subscribed twice. `combine` operator will combine latest values from each, so these two can cause double call of downstream operators.

I think as in the other PR, this should be more thought of (I'd suggest making one `StateFlow` with the data class for user actions as well).

arriolac: I remember you calling this out in a previous PR - agreed that we should revisit this approach. I've opted in for the `shareIn` approach in 737e634f

Jetcaster/mobile/src/main/java/com/example/jetcaster/ui/home/HomeViewModel.kt
```diff
@@ -70,6 +73,9 @@ class HomeViewModel @Inject constructor(
     // Holds the view state if the UI is refreshing for new data
     private val refreshing = MutableStateFlow(false)
 
+    private val subscribedPodcasts = podcastStore.followedPodcastsSortedByLastEpisode(limit = 10)
```
mlykotom, line 76: NIT: I'd add a comment why this uses `shareIn` operator as when someone reads it in the future, they will think "this is unnecessary" and remove it :)

mlykotom (approved):

PR #1372, opened 2024-04-26
[Wear] Adds previews to queue and podcasts

kul3r4:
- Adds previews to queue and podcasts
- Fixes README which had some redundant `[` `]`

### DIFF
--- a/Jetcaster/README.md
+++ b/Jetcaster/README.md
@@ -71,7 +71,7 @@ The sample implements [Wear UX best practices for media apps][mediappsbestpracti
 - Display scrollbar on scrolling
 - Display the time on top of the screens
 
-The sample is built using the [Media Toolkit][[mediatoolkit]] which is an open source
+The sample is built using the [Media Toolkit][mediatoolkit] which is an open source
 project part of [Horologist][horologist] to ease the development of media apps on Wear OS built on top of Compose for Wear.
 It provides ready to use UI screens, such the [EntityScreen][entityscreen]
 that is used in this sample to implement many screens such as Podcast, LatestEpisodes and Queue. [Horologist][horologist] also provides
@@ -81,7 +81,7 @@ For simplicity, this sample uses a mock Player which is reused across form facto
 if you want to see an advanced Media sample built on Compose that uses Exoplayer and plays media content,
 refer to the [Media Toolkit sample][mediatoolkitsample].
 
-The [official media app guidance for Wear OS][ [wearmediaguidance]]
+The [official media app guidance for Wear OS][wearmediaguidance]
 advices to download content on the watch before listening to preserve power, this feature will be added to this sample in future iterations. You can
 refer to the [Media Toolkit sample][mediatoolkitsample] to learn how to implement the media download feature.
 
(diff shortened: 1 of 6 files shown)


Jetcaster/wear/src/main/java/com/example/jetcaster/PreviewData.kt
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package com.example.jetcaster
+
+import com.example.jetcaster.core.model.PlayerEpisode
+import com.example.jetcaster.core.model.PodcastInfo
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
+
+val PreviewPodcasts = listOf(
```
yongsuk44, line 24: Using [PreviewParameterProvider](https://developer.android.com/reference/kotlin/androidx/compose/ui/tooling/preview/PreviewParameterProvider) to explicitly supply parameters to Preview looks good.

kul3r4: thank you! Will take a look

kul3r4: done,thanks

yschimke (approved):

Jetcaster/wear/src/main/java/com/example/jetcaster/podcasts/PodcastsScreen.kt
```diff
@@ -135,40 +93,114 @@ fun PodcastsScreen(
...
+            PodcastsScreenState.Empty ->
+                PodcastScreenEmpty(onDismiss)
+            PodcastsScreenState.Loading ->
+                PodcastScreenLoading(columnState)
+        }
+    }
+}
+
+@Composable
+fun PodcastScreenLoaded(
```
arriolac, line 110: Nit: Make sure to add modifiers to these composables. Same in `QueueScreen`.

arriolac (approved):

Jetcaster/wear/src/main/java/com/example/jetcaster/podcasts/PodcastsScreen.kt
yschimke: Is it worth adding lint to fail on these?

yschimke: Actually maybe it won't for this, just based on order?

arriolac: > Is it worth adding lint to fail on these?

Error might be too aggressive—it should warn at the very least.

Also, if you are using live templates, typing "comp" in the editor will automatically create the following template for you so it's easy to add a composable with a default modifier (I know this was in AS preview and I think should be available now in the stable channel with jellyfish):

```
@Composable
fun <cursor here>(modifier: Modifier = Modifier) {
}
```

kul3r4: thanks, added

PR #1373, opened 2024-04-26
[Jetcaster]: Rename 'app' module to 'mobile'

arriolac:
Rename `app` module to `mobile` for a more descriptive name for the target form factor.


.github/workflows/build-sample.yml
```diff
@@ -9,6 +9,9 @@ on:
       path:
         required: true
         type: string
+      module:
+        default: "app"
```
mlykotom, line 13: is the default "app"? or "mobile"?

mlykotom (approved):
One comment about default, otherwise LGTM

.github/workflows/build-sample.yml
arriolac: The default should be "app" here since `build-sample.yml` is a shared workflow across all of our samples which all use "app" for the app module with the exception of Jetcaster

PR #1377, opened 2024-04-29
[Jetcaster]: HTML format episode summary in podcast detail.

arriolac:
Before:

After:

### DIFF
--- a/Jetcaster/designsystem/src/main/java/com/example/jetcaster/designsystem/component/HtmlTextContainer.kt
+++ b/Jetcaster/designsystem/src/main/java/com/example/jetcaster/designsystem/component/HtmlTextContainer.kt
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.example.jetcaster.ui.shared
+package com.example.jetcaster.designsystem.component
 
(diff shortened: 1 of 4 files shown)


kul3r4:
@arriolac this would be handy for Wear as well

arriolac:
> @arriolac this would be handy for Wear as well

Moved to `designsystem` so Wear can use it as well @kul3r4

Jetcaster/designsystem/src/main/java/com/example/jetcaster/designsystem/component/HtmlTextContainer.kt
```diff
@@ -0,0 +1,44 @@
...
+import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.buildAnnotatedString
+import androidx.core.text.HtmlCompat
+
+/**
+ * A container for text that should be HTML formatted. This container will handle building the
+ * annotated string from [text], and enable text selection if [text] has any selectable element.
+ */
+@Composable
+fun HtmlTextContainer(
```
bentrengrove, line 31: There is an API for this in Compose 1.7, consider upgrading or adding a TODO

bentrengrove (approved):

PR #1380, opened 2024-04-30
Remove building release on PRs

riggaroo:
When building from a fork pull request, the secrets are not shared. This PR removes the need to build a release since it wasn't used, named the signing config it to be a release.

### DIFF
--- a/.github/workflows/Crane.yaml
+++ b/.github/workflows/Crane.yaml
@@ -76,7 +76,7 @@ jobs:
           target: google_apis
           arch: x86
           disable-animations: true
-          script: ./gradlew connectedCheck --stacktrace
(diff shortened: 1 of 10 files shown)


kul3r4 (approved):

.github/workflows/Crane.yaml
```diff
@@ -76,7 +76,7 @@ jobs:
           target: google_apis
           arch: x86
           disable-animations: true
-          script: ./gradlew connectedCheck --stacktrace
+          script: ./gradlew connectedAndroidDebugTest --stacktrace
```
arriolac, line 79: Looks like this change causes Crane builds to fail https://github.com/android/compose-samples/actions/runs/8898242654/job/24435886140?pr=1380

arriolac (approved):

PR #1385, opened 2024-05-01
[Wear] Use a podcast specific scaffold

kul3r4:
- Inline scaffold
- Move all screens under UI package

### DIFF
--- a/Jetcaster/wear/src/main/java/com/example/jetcaster/WearApp.kt
+++ b/Jetcaster/wear/src/main/java/com/example/jetcaster/WearApp.kt
@@ -16,17 +16,20 @@
 
 package com.example.jetcaster
 
+import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.pager.rememberPagerState
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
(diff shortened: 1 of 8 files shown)


Jetcaster/wear/src/main/java/com/example/jetcaster/WearApp.kt
```diff
@@ -59,12 +40,8 @@ fun WearApp() {
     val navHostState = rememberSwipeDismissableNavHostState()
     val volumeViewModel: VolumeViewModel = viewModel(factory = VolumeViewModel.Factory)
 
-    // TODO remove from MediaPlayerScaffold
-    val snackBarManager: SnackbarManager = SnackbarManager()
-    val snackbarViewModel: SnackbarViewModel = SnackbarViewModel(snackBarManager)
-
     WearAppTheme {
-        MediaPlayerScaffold(
+        JetcasterScaffold(
```
yschimke, line 44: Does it add much to have it in JetcasterScaffold? Should you just inline it here?

yschimke (approved):

Jetcaster/wear/src/main/java/com/example/jetcaster/WearApp.kt
kul3r4: Good point, removed

PR #1392, opened 2024-05-03
[Jetcaster]: Create data-testing module.

arriolac:
Move repository mocks in a new separate module (`:data-testing`).


Jetcaster/settings.gradle.kts
```diff
@@ -45,3 +45,4 @@ include(
     ":wear"
 )
 enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
+include(":core:data-testing")
```
bentrengrove, line 48: Nit: Should be in the above include

bentrengrove (approved):

PR #1393, opened 2024-05-06
[Jetcaster]: Change alpha of prev/next buttons.

arriolac:
Change alpha of previous and next buttons in the player to help indicate if those actions are possible based on the state of the player (i.e. previous only possible while playing and next only possible when queue isn't empty).

Disabled:

Enabled:


Jetcaster/mobile/src/main/java/com/example/jetcaster/ui/player/PlayerScreen.kt
```diff
@@ -712,6 +713,9 @@ private fun PlayerSlider(
     }
 }
 
+private fun Modifier.invisibleWhen(predicate: () -> Boolean): Modifier =
```
jdkoren, line 716: Doesn't seem like this is being used in this change, but if you already have some intended uses, feel free to keep it.

jdkoren (approved):

Jetcaster/mobile/src/main/java/com/example/jetcaster/ui/player/PlayerScreen.kt
arriolac: Thanks, yep I meant to remove this.

PR #1395, opened 2024-05-07
[Jetcaster] Add MediaPlayer to Mobile Module Screen

yongsuk44:
I have implemented basic playback functionality through `MediaSession` and `MediaController`, although there are still features that need to be implemented.

---

[Screen_recording_20240508_031547.webm](https://github.com/android/compose-samples/assets/38935359/be5dba01-f8c8-4033-81a7-6fb139a9910e)


arriolac (commented):
I am a bit hesitant to integrate this as playing media is beyond the scope of the sample. As mentioned in the README, the primary goal of this sample is to showcase building UI with Compose across different form factors. So, adding non-UI related complexity may take away from that goal; further, adding real media playback has a maintenance cost that I'm not sure we are willing to shoulder. 

I'll check with my team and confirm the decision in another update. I do appreciate the (pretty significant) contribution though but want to be transparent about the implications of integrating this.

arriolac:
Confirmed that this is not a feature we want to integrate in Jetcaster. For Media API best practices, please have a look at [Socialite](https://github.com/android/socialite/) instead.

PR #1407, opened 2024-05-28
[Reply] Use NavigationSuite for nav UI components

jdkoren:


Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationComponents.kt
```diff
@@ -49,20 +50,119 @@ import androidx.compose.material3.NavigationRail
...
+        WindowHeightSizeClass.EXPANDED -> ReplyNavigationContentPosition.CENTER
+        else -> ReplyNavigationContentPosition.TOP
+    }
+
+    val drawerState = rememberDrawerState(initialValue = DrawerValue.Closed)
+    val coroutineScope = rememberCoroutineScope()
+    // Avoid opening the modal drawer when there is a permanent drawer or a bottom nav bar,
+    // but always allow closing an open drawer.
+    val gesturesEnabled =
+        drawerState.isOpen || navLayoutType == NavigationSuiteType.NavigationRail
```
alexvanyo, line 118: This logic does makes sense for consistency, where the drawer will remain open if it was open, but maybe an alternative would be to just hide it immediately if the navigation layout type isn't a navigation rail? Not sure which way would be better.

alexvanyo (approved):

Reply/gradle/libs.versions.toml
```diff
@@ -10,20 +10,25 @@ androidx-appcompat = "1.6.1"
 androidx-benchmark = "1.2.0"
 androidx-benchmark-junit4 = "1.2.1"
 androidx-compose-bom = "2024.05.00"
+androidx-compose-material3-adaptive = "1.0.0-beta01"
```
bentrengrove, line 13: You need to make all these changes in the scripts/libs.versions.toml file or else they will be overriden on the next dependency update

Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationComponents.kt
jdkoren: I tried something like that, and it looked fine when changing to a layout that isn't a rail. But when changing back to the rail the drawer opens again, and that's a bit jarring.

We can work around that with a `LaunchedEffect` that calls `drawerState.snapTo(Closed)`, but that feels more imperative than declarative.

Reply/gradle/libs.versions.toml
bentrengrove: All good! When I first looked it didn't look like they matched but maybe I was wrong

bentrengrove (approved):
LGTM

PR #1414, opened 2024-06-11
[All] Remove duplicate deps and use top-level libs.versions.toml as deps SoT.

arriolac:
Changes:

1. Removes duplicate dependencies from Glance and Wear in top-level libs.versions.toml
2. Duplicates top-level libs.versions.toml to sample-specific libs.versions.toml by running ./scripts/duplicate_version_config.sh
3. Update deps by running ./script/updateDeps.sh


bentrengrove (approved):

arriolac:
Looks like upgrading androidx-lifecycle to 2.8.1 causes tests to fail on Jetnews and Jetchat so I reverted the version back to 2.7.0 in [3c30d20](https://github.com/android/compose-samples/pull/1414/commits/3c30d2051ce9a343b391cdccf479e2ad1428d456).

PR #1420, opened 2024-06-27
Fixed TV release build

pflammertsma:


mlykotom (approved):

IanGClifton (approved):
Is there a bug to follow for getting "-keep class androidx.compose.ui.platform.AndroidCompositionLocals_androidKt { *; }" added on the androidx side so that apps don't need to manually add that?

riggaroo (approved):

PR #1422, opened 2024-06-28
Switch to using static images for jetsnack

riggaroo:
Due to unsplash source urls not being available anymore, we've switched to using static Images instead.


Jetsnack/app/src/main/res/drawable-nodpi/almonds.jpg
JolandaVerhoef, line 1: Any chance we can optimize these images by lowering their resolution, converting them to webp/png, or running a jpg optimization algorithm on them? 
332KB seems a bit big.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Snacks.kt
```diff
@@ -296,7 +298,7 @@ fun SnackImage(
     ) {
         AsyncImage(
             model = ImageRequest.Builder(LocalContext.current)
-                .data(imageUrl)
+                .data(imageRes)
                 .crossfade(true)
                 .build(),
             contentDescription = contentDescription,
```
JolandaVerhoef, line 304: Should you consider removing the placeholder? You removed the jpg so maybe you also want to remove the placeholder parameter here.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Snacks.kt
```diff
@@ -296,7 +298,7 @@ fun SnackImage(
     ) {
         AsyncImage(
```
JolandaVerhoef, line 299: Would you still need an asyncImage at all or would you consider using a typical `Image` composable instead?

riggaroo: AsyncImage loads the image off the main thread, which is better for performance.

Jetsnack/app/src/main/res/drawable-nodpi/almonds.jpg
riggaroo: I reduced these quite a bit already, started with large images like 8mb each :D Anyway I reduced a lot of these some more.

JolandaVerhoef (approved):

PR #1423, opened 2024-06-30
[Reply] Replaced deprecated codes and optimised performance

mubashirpa:
Replaced deprecated DockedSearchBar and removed unused codes from Reply to improve performance.


riggaroo (approved):
Thanks!

PR #1428, opened 2024-07-09
Replacement of tv-foudation from the dependencies of the TV app

chikoski:
This pull request replace the tv-foundation with Compose Foundation 1.7.0-beta04 as tv-foundation is going to be deprecated in the next release. 

In addition to that, this pull request fixes a crashing bug that happens on the discovery screen of the TV app, as a part of the replacement.


Jetcaster/tv/src/main/java/com/example/jetcaster/tv/ui/JetcasterApp.kt
```diff
@@ -126,7 +143,7 @@ private fun GlobalNavigationContainer(
                 ) {
                     Text(text = "Library")
                 }
-                Spacer(modifier = Modifier.weight(1f))
+                Spacer(modifier = modifier.weight(1f))
```
chikoski, line 146: Updated as `Modifier.weight(1f)`

Jetcaster/gradle/libs.versions.toml
JolandaVerhoef, line 1: All samples use the same main libs.version.toml file. It is duplicated every time we do a release cycle. So updating one sample's dependencies will not work, as the file will simply be overwritten next time the samples are updated.

Do we really need this update now, or can we wait for the stable 1.7.0 release?

chikoski: Thanks for the suggestion. It sounds we should wait the stable 1.7.0 release.  I will change the state of the PR to draft and come back when 1.7.0 becomes stable

chikoski:
@JolandaVerhoef Would you please take a look the updated change, which was blocked by stable 1.7.0 release?

JolandaVerhoef (approved):

PR #1432, opened 2024-07-16
[Jetcaster] Touch up the TV UI

bentrengrove:
Fix podcast art padding
Fix clipping on podcast screen
Fix navigation transition jumping around


riggaroo (approved):
thanks for fixing!

PR #1433, opened 2024-07-16
Add animating background gradient on JetSnack

riggaroo:
https://github.com/user-attachments/assets/f793fd86-fe65-4164-af68-9f1a9b41b184

### DIFF
--- a/Jetsnack/app/src/main/java/com/example/jetsnack/ui/snackdetail/SnackDetail.kt
+++ b/Jetsnack/app/src/main/java/com/example/jetsnack/ui/snackdetail/SnackDetail.kt
@@ -17,6 +17,12 @@
 package com.example.jetsnack.ui.snackdetail
 
 import android.content.res.Configuration
+import androidx.compose.animation.core.LinearEasing
+import androidx.compose.animation.core.RepeatMode
+import androidx.compose.animation.core.animateFloat
+import androidx.compose.animation.core.infiniteRepeatable
+import androidx.compose.animation.core.rememberInfiniteTransition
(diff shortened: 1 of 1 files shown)


Jetsnack/app/src/main/java/com/example/jetsnack/ui/snackdetail/SnackDetail.kt
```diff
@@ -112,11 +121,36 @@ fun SnackDetail(
...
+        end = Offset(offset + brushSize, offset + brushSize),
+        tileMode = TileMode.Mirror
+    )
     Spacer(
         modifier = Modifier
             .height(280.dp)
             .fillMaxWidth()
-            .background(Brush.horizontalGradient(JetsnackTheme.colors.tornado1))
+            .blur(40.dp)
+            .background(brush)
```
mlykotom, line 152: Do you think we could change this to custom drawing that would prevent recomposing this header and would just redraw?
image


```
            .drawWithCache {
                val brushSize = 400f
                val brush = Brush.linearGradient(
                    colors = brushColors,
                    start = Offset(offset, offset),
                    end = Offset(offset + brushSize, offset + brushSize),
                    tileMode = TileMode.Mirror
                )
                onDrawBehind {
                    drawRect(brush)
                }
            }
```

riggaroo: Good idea, thanks Tomas.

mlykotom (approved):
LGTM!

PR #1435, opened 2024-07-24
Jetlagged Home Screen - use Flows 

riggaroo:
Using FlowRow + FlowColumn combined with WindowSize classes to adapt better to different device sizes. 

image


JetLagged/app/src/main/java/com/example/jetlagged/MainActivity.kt
```diff
@@ -20,16 +20,20 @@ import android.os.Bundle
...
 import com.example.jetlagged.ui.theme.JetLaggedTheme
 
 class MainActivity : ComponentActivity() {
 
+    @OptIn(ExperimentalMaterial3WindowSizeClassApi::class)
     override fun onCreate(savedInstanceState: Bundle?) {
         enableEdgeToEdge()
         super.onCreate(savedInstanceState)
         setContent {
+            val windowSizeClass = calculateWindowSizeClass(this)
```
tiwiz, line 34: Another approach is to use [`currentWindowAdaptiveInfo()`](https://developer.android.com/reference/kotlin/androidx/compose/material3/adaptive/package-summary#currentWindowAdaptiveInfo()) from M3 inside your `Composable`, and this function has the huge pro to also give you the posture of the device.  Since you're using M3, I would suggest to use this function, so you don't have to pass the `WindowSizeClass` to every `Composable`, but you can calculate it only where you need :)

tiwiz (approved):

JetLagged/app/src/main/java/com/example/jetlagged/JetLaggedScreen.kt
```diff
@@ -17,44 +17,37 @@
...
+import com.example.jetlagged.ui.util.MultiDevicePreview
 
-@Preview(showBackground = true)
-@Preview(device = Devices.FOLDABLE, showBackground = true)
+@OptIn(ExperimentalLayoutApi::class)
+@MultiDevicePreview
 @Composable
 fun JetLaggedScreen(
+    windowSizeClass: WindowWidthSizeClass = WindowWidthSizeClass.Compact,
+    viewModel: JetLaggedHomeScreenViewModel = viewModel(),
```
dturner, line 50: `modifier` should be first optional param

dturner (approved):

PR #1437, opened 2024-08-02
Jetsnack - Migrate to Material 3

riggaroo:
Jetsnack sample is using Material 2 instead of Material 3, keeping most of the UI the same as previously. Material 3 is the preferred choice for Material. 

This pull request migrates to using M3, as well as fixes the filters screen selection to have a shared element transition to show and hide the dialog. 

https://github.com/user-attachments/assets/0121a41c-2aff-4bb0-abc3-d2075a8f2bd4

It also fixes build issues some machines were experiencing, where there is a mismatch in JVM versions by setting `jvmToolchain` version.


Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/cart/SwipeDismissItem.kt
```diff
@@ -16,52 +16,46 @@
...
-            directions = directions,
-            background = { background(offset) },
-            dismissContent = { content(isDismissed) }
-        )
-    }
+        state = dismissState,
+        backgroundContent = { background(/*offset*/) },
+        content = { content(/*isDismissed*/) }
+    )
+    /*   }*/
```
bentrengrove, line 60: Remove the commented out sections?

bentrengrove (approved):
Nice work!

Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/cart/SwipeDismissItem.kt
riggaroo: Thanks for reminding me about this one, I added it back based on new dismissed state information (progress percentage instead of DP)

PR #1443, opened 2024-08-12
Upgrade to new spotless version for all samples

riggaroo:
Upgrade to version 1.3.1 of Ktlint, migrating to the standard rules where applicable. This change is the result of running the new rules - its large since it hasn't been run in a long time unfortunately. 

**Two rules disabled as per Compose conventions:**
Composable function names, allow them to start with lowercase
 Allowing PascalCase naming on constants


JetLagged/.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*.{kt,kts}]
```
JolandaVerhoef, line 3: WDYT about commenting why we're adding these exceptions? I'm assuming it's so that the different linters / checkers are in sync?

JetLagged/app/build.gradle.kts
```diff
@@ -22,13 +23,22 @@ plugins {
 }
 
 android {
-    compileSdk = libs.versions.compileSdk.get().toInt()
+    compileSdk =
+        libs.versions.compileSdk
+            .get()
+            .toInt()
```
JolandaVerhoef, line 29: This looks a bit overly optimized, are we sure we want this level of formatting?

JetLagged/app/src/main/java/com/example/jetlagged/HomeScreenCards.kt
```diff
@@ -73,15 +73,16 @@ import com.example.jetlagged.ui.theme.SmallHeadingStyle
...
     val shape = RoundedCornerShape(24.dp)
     Card(
         shape = shape,
         colors = CardDefaults.cardColors(containerColor = Color.White),
-        modifier = modifier
-            .padding(8.dp),
-        border = BorderStroke(2.dp, borderColor)
+        modifier =
+            modifier
+                .padding(8.dp),
```
JolandaVerhoef, line 84: Again here, this doesn't look very readable, WDYT?

JetLagged/build.gradle.kts
```diff
@@ -14,13 +14,40 @@
...
+import com.diffplug.gradle.spotless.SpotlessExtension
+
 plugins {
     alias(libs.plugins.gradle.versions)
     alias(libs.plugins.version.catalog.update)
     alias(libs.plugins.android.application) apply false
     alias(libs.plugins.kotlin.android) apply false
     alias(libs.plugins.kotlin.parcelize) apply false
     alias(libs.plugins.compose) apply false
+    alias(libs.plugins.spotless)
```
dturner, line 26: One of the issues with Spotless is that it's not compatible with the Gradle configuration cache. This is why it should only be applied using an init script, rather than being part of the project build config.

dturner:
Following discussion, I thought it'd be helpful to describe the process for identifying and [disabling a ktlint rule](https://pinterest.github.io/ktlint/0.49.1/rules/configuration-ktlint/#disabled-rules) which is run through spotless.  

To disable a ktlint rule, you need to know 2 things: 

- the rule name
- the ruleset which it appears in 

The problem is that running `gradlew spotlessCheck` doesn't provide the names of the rules which are being violated (reported [here](https://github.com/diffplug/spotless/issues/1278), [here](https://github.com/diffplug/spotless/issues/1551) and [here](https://github.com/diffplug/spotless/issues/1278)). To find that out, you need to run `ktlint` directly (`brew install ktlint`). 

I ran it on the JetLagged source using `ktlint "app/src/**/*.kt"` after rolling back some of the formatting changes in this PR. This gives you the names of any rule violations. In my case, one of the rules being violated was `standard:multiline-expression-wrapping` meaning it's in the `standard` ruleset and is named `multiline-expression-wrapping`.
  
To disable it, add an entry to `.editorconfig` using the following format `ktlint_<ruleset>_<rulename> = disabled` so in this case it's `ktlint_standard_multiline-expression-wrapping = disabled`. 

BUT WAIT. Spotless is problematic because it doesn't work with configuration caching so you may find that it doesn't pick up the changes in `.editorconfig` ([as reported here](https://github.com/diffplug/spotless/issues/1924)). To solve this use `--no-configuration-cache` and `--no-daemon` flags, so: `./gradlew spotlessCheck --no-daemon --no-configuration-cache`.

PR #1448, opened 2024-08-26
Added predictive back animation to drawer

IanGClifton:
This uses a PredictiveBackHandler and collects the flow of back events to manipulate the translation. I created separate openDrawer and closeDrawer functions to reuse the code in both toggleDrawerState and in the PredictiveBackHandler implementation.

### DIFF
--- a/JetLagged/app/src/main/java/com/example/jetlagged/JetLaggedDrawer.kt
+++ b/JetLagged/app/src/main/java/com/example/jetlagged/JetLaggedDrawer.kt
@@ -16,6 +16,7 @@
 
 package com.example.jetlagged
 
+import android.os.SystemClock
 import androidx.activity.compose.PredictiveBackHandler
 import androidx.compose.animation.core.Animatable
 import androidx.compose.animation.core.calculateTargetValue
@@ -48,14 +49,16 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.graphicsLayer
 import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.input.pointer.util.VelocityTracker
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.util.lerp
-import kotlin.coroutines.cancellation.CancellationException
 import kotlinx.coroutines.launch
+import kotlin.coroutines.cancellation.CancellationException
 
 @Composable
 fun HomeScreenDrawer(windowSizeClass: WindowSizeClass) {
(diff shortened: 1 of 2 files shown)


JetLagged/app/src/main/java/com/example/jetlagged/JetLaggedDrawer.kt
```diff
@@ -79,18 +81,32 @@ fun HomeScreenDrawer(windowSizeClass: WindowSizeClass) {
...
-                    DrawerState.Closed
-                } else {
-                    DrawerState.Open
+            }
+        }
+        PredictiveBackHandler(drawerState == DrawerState.Open) { progress ->
+            try {
+                progress.collect { backEvent ->
+                    val targetSize = (drawerWidth - (drawerWidth * backEvent.progress))
+                    translationX.snapTo(targetSize)
```
riggaroo, line 105: I'm not sure why i didn't merge my PR on this, but i think it'd be nice to do a similar thing to what i was doing before - ie tracking the velocity and finishing the gesture with the velocity.
https://github.com/android/compose-samples/pull/1185 

Something like this is what I had:
```
PredictiveBackHandler(drawerState == DrawerState.Closed) { progress: Flow<BackEventCompat> ->
            // code for gesture back started
            try {
                progress.collect { backEvent ->
                    // code for progress
                    if (drawerState == DrawerState.Closed) {
                        translationX.snapTo(backEvent.progress * drawerWidth)
                        velocityTracker.addPosition(System.currentTimeMillis(), Offset(backEvent.touchX, backEvent.touchY))
                    }
                }
                // code for completion
                val endProgress = if (translationX.value > drawerWidth * 0.2f) {
                    drawerWidth
                } else {
                    0f
                }
                translationX.animateTo(endProgress, initialVelocity = velocityTracker.calculateVelocity().x)

                if (endProgress == 0f){
                    drawerState = DrawerState.Closed
                } else {
                    drawerState = DrawerState.Open
                }
            } catch (e: CancellationException) {
                // code for cancellation
            }
        }
```

IanGClifton: I added in the VelocityTracker, but I'm not sure I can tell the difference. I might have done something wrong, but if I log out the `backEvent.touchX` and `backEvent.touchY`, I see the same values repeated multiple times at the end of the back event, which could also be affecting the velocity calculation.

riggaroo (approved):
LGTM

JetLagged/app/src/main/java/com/example/jetlagged/JetLaggedDrawer.kt
riggaroo: Yeah maybe - well feel free to remove if it doesn't seem to affect it, maybe there is built in velocity calculations into the progress.

IanGClifton: Playing with it on a different device today, I _think_ it does have an impact, but it seems minimal. If I multiply the velocity by a large number, I see a difference between a really slow swipe and a faster one, so I'll leave the velocity in for now. Maybe at some point we can play with figuring out the right multiplier.

PR #1452, opened 2024-08-29
Fixed spacing for floating toolbar + back support

IanGClifton:
The floating toolbar seen in expanded width windows was implemented as a sticky header. The problem with that is it allows the toolbar to go behind the status bar on scroll due to edge-to-edge. For a real sticky header, you'd probably want to implement inset handling into the header so that it sticks just below any insets. In this case, that would mean the toolbar never actually moves, so I pulled it out of sticky header and just put it as a sibling with the LazyColumn. By adding the safeDrawing insets, this keeps it positioned at the top right of the safeDrawing area.

I also added a BackHandler for the drawer, so that pressing back closes the drawer. I don't see any clean way to implement predictive back with a modal navigation drawer.

### DIFF
--- a/JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
+++ b/JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsApp.kt
@@ -16,10 +16,10 @@
 
 package com.example.jetnews.ui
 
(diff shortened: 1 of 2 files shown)


JolandaVerhoef (approved):
LGTM!

PR #1454, opened 2024-08-30
Added predictive back support to the drawer

IanGClifton:
In M3 1.3, the drawerState can be passed to the ModalDrawerSheet to get predictive back animations without the need for a PredictiveBackHandler to be manually defined.


IanGClifton:
I've created this as a draft to validate it. Once there is a BOM that uses M3 1.3.0 stable, we can update all samples to that and I can push a new commit to this PR to remove the custom M3 version being set for this sample.

JetNews/gradle/libs.versions.toml
```diff
@@ -9,10 +9,7 @@ androidx-activity-compose = "1.9.0"
 androidx-appcompat = "1.7.0"
 androidx-benchmark = "1.2.4"
 androidx-benchmark-junit4 = "1.2.4"
-androidx-compose-bom = "2024.06.00"
```
riggaroo, line 12: I think this changes can be reverted right? You should also change all the time files at once and not individually if making a change to them.

riggaroo (changes requested):
Mostly lgtm, just the toml changes.

JetNews/gradle/libs.versions.toml
IanGClifton: Yeah, I made the temporary change to make sure it could build with pre-stable and then updated the bom to the same as main to verify it still worked correctly with the released libraries. I just merged with main to make it more clear though :)

riggaroo (approved):

PR #1455, opened 2024-08-30
Added predictive back support to Jetchat drawer

IanGClifton:
In M3 1.3, the drawerState can be passed to the ModalDrawerSheet to get predictive back animations without the need for a PredictiveBackHandler to be manually defined. I also moved the color settings to the correct place (ModalDrawerSheet) to make sure the drawer portions covered by insets would be colored correctly.

### DIFF
--- a/Jetchat/gradle/libs.versions.toml
+++ b/Jetchat/gradle/libs.versions.toml
@@ -10,8 +10,6 @@ androidx-appcompat = "1.7.0"
 androidx-benchmark = "1.2.4"
 androidx-benchmark-junit4 = "1.2.4"
 androidx-compose-bom = "2024.09.00"
-androidx-compose-latest = "1.7.0-beta07"
-androidx-compose-material3-adaptive = "1.0.0-beta04"
 androidx-constraintlayout = "1.1.0-alpha13"
 androidx-core-splashscreen = "1.0.1"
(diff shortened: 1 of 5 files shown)


IanGClifton:
I've created this as a draft to validate it. Once there is a BOM that uses M3 1.3.0 stable, we can update all samples to that and I can push a new commit to this PR to remove the custom M3 version being set for this sample.

Jetchat/gradle/libs.versions.toml
```diff
@@ -9,10 +9,7 @@ androidx-activity-compose = "1.9.0"
 androidx-appcompat = "1.7.0"
 androidx-benchmark = "1.2.4"
 androidx-benchmark-junit4 = "1.2.4"
-androidx-compose-bom = "2024.06.00"
```
bentrengrove, line 12: If you rebase, you should be able to pick this up from main

bentrengrove (approved):

Jetchat/gradle/libs.versions.toml
IanGClifton: That's what I tried at first, but then it was giving me errors about the updated AGP version.... which should have just worked without a problem, so I'm not sure if it was just a local caching issue.

PR #1459, opened 2024-09-06
Add type safe navigation

dturner:
Adding type safe navigation to Compose samples now that the stable API has been released.

I also had to add
```
kotlinOptions {
    jvmTarget = "17"
}
```

to avoid getting the following build error:

```
Inconsistent JVM-target compatibility detected for tasks 'compileDebugJavaWithJavac' (17) and 'compileDebugKotlin' (21).
```


Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationActions.kt
```diff
@@ -25,16 +25,15 @@ import androidx.compose.ui.graphics.vector.ImageVector
...
 import com.example.reply.R
+import kotlinx.serialization.Serializable
 
-object ReplyRoute {
-    const val INBOX = "Inbox"
-    const val ARTICLES = "Articles"
-    const val DM = "DirectMessages"
-    const val GROUPS = "Groups"
-}
+@Serializable data object InboxRoute
```
riggaroo, line 30: What about using a sealed class here? Might be useful for limiting what can be defined as a route?

riggaroo (approved):
Lgtm!

Reply/app/src/main/java/com/example/reply/ui/navigation/ReplyNavigationActions.kt
dturner: Good suggestion. Done.

PR #1464, opened 2024-09-10
[Jetcaster] Refactor to use fromHtml in AnnotatedString

yongsuk44:
HTML formatting support from compose 1.7.0

image


bentrengrove (approved):
Thank you for doing this!

PR #1466, opened 2024-09-12
Fixed the search bar text input

Sanu5:
fixed the no input form keyboard issue


mlykotom:
Thank you for your contribution! This was simulated with a local state in the search bar.

PR #1475, opened 2024-09-30
Refactor FAB with onScroll animation

jamilxt:
The large FAB looks weird with the design. So I've refactored it to use Extended FAB and added animation onScroll like Gmail does. 

### Old:

We can also consider showing/hiding the Search Bar & Bottom Navigation Bar on the scroll. I will try to do it next if this PR gets merged.
Thank you.


Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
```diff
@@ -113,20 +119,17 @@ fun ReplyInboxScreen(
...
-                    contentColor = MaterialTheme.colorScheme.onTertiaryContainer
-                ) {
-                    Icon(
-                        imageVector = Icons.Default.Edit,
-                        contentDescription = stringResource(id = R.string.compose),
-                        modifier = Modifier.size(28.dp)
-                    )
-                }
+                    contentColor = MaterialTheme.colorScheme.onTertiaryContainer,
+                    expanded = emailLazyListState.isScrollingUp()
```
bentrengrove, line 131: No need for an extension here, you can replace this with
```suggestion
                    expanded = emailLazyListState.lastScrolledBackward ||
                            !emailLazyListState.canScrollBackward
```

bentrengrove (changes requested):
Thanks for the submission! I am just checking with our designer whether this meets the Material spec that this sample is demonstrating

One minor comment as well, there is an easier way to determine if the fab should be expanded

Reply/app/src/main/java/com/example/reply/ui/ReplyListContent.kt
jamilxt: Thank you! I really appreciate it. It's been exciting to try Android development again after 4 years!

bentrengrove (approved):
LGTM!

bentrengrove:
Looks like it just needs formatting. You should be able to just run

`./gradlew --init-script buildscripts/init.gradle.kts spotlessApply`

jamilxt:
> Looks like it just needs formatting. You should be able to just run
> 
> `./gradlew --init-script buildscripts/init.gradle.kts spotlessApply`

I've applied spotless to format the code. 
Thank you.

PR #1479, opened 2024-10-03
Fix dependency updates

### LINKED ISSUE #1478: [Bug]: Updating dependencies adds non-existent version for org.jetbrains.kotlinx:kotlinx-collections-immutable
Steps to repro: 

Run `./JetNews/gradlew -p ./JetNews versionCatalogUpdate` (this command is taken from `scripts/updateDeps.sh`).

This will update the library version for `kotlinx_immutable` to `1.9.0` which doesn't exist (latest version is `0.3.8`). The build will then fail. 

Cause: 

In `libs.versions.toml` the `kotlinx_immutable` version variable is also used for `kotlinx-coroutines-test`. This is incorrect.

dturner:
Fixed #1478 by using the correct version variable for `kotlinx-coroutines-test`

Also changed the master project used for version updates from `JetNews` to `Reply`. 

Why? `JetNews` doesn't use the `kotlinx-serialization` plugin so this was automatically removed when running `versionCatalogUpdate` gradle task. This meant that `Reply` would fail to build because it depends on this plugin.

### DIFF
--- a/JetLagged/gradle/libs.versions.toml
+++ b/JetLagged/gradle/libs.versions.toml
@@ -43,7 +43,7 @@ jdkDesugar = "1.2.2"
 junit = "4.13.2"
 kotlin = "2.0.20"
 kotlinx-serialization-json = "1.7.3"
-kotlinx_immutable = "1.9.0"
+kotlinx_immutable = "0.3.8"
 ksp = "2.0.20-1.0.24"
 maps-compose = "3.1.1"
 # @keep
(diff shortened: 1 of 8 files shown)


JoseAlcerreca (approved):

dturner:
For some reason the instrumented tests are failing on the API 29 emulator. When I run locally all tests pass.

JolandaVerhoef (approved):

PR #1482, opened 2024-10-04
Fix AppBar padding and alignment issues

clarabayarri:
AppBar was missing a start padding that caused it to extend to the edge of the screen on Compact size classes.

Updated the SearchBar component usage away from the deprecated method version as an internal usage of Modifier.fillMaxWidth in the deprecated method was now causing the AppBar to always fill max width on all Window Size Classes (vs the intended end-aligned smaller version in non-Compact)

Phone - after
jetcaster_phone_after
Phone - before
jetcaster_phone_before
Tablet - after
jetcaster_tablet_after
Tablet - before
jetcaster_tablet_before


JolandaVerhoef (approved):
LGTM!

PR #1486, opened 2024-10-08
[Jetcaster] Refactor HomeContent to use grid only

yongsuk44:
https://github.com/user-attachments/assets/6cdda4fa-8416-4284-82f8-32dcc5032b89


jdkoren (approved):
Code looks good to me, but you may need to resolve conflicts again.

Issue #1487, opened 2024-10-09
[FR]: Import modules with gradle typesafe API feature


Xiryl:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is this a feature request for one of the samples?

- [X] Yes, this is a specific request related to this samples repo.

### Sample app

Jetcaster

### Describe the problem

In the jetcaster project, some modules are not imported with gradle's typesafe feature.
e.g. https://github.com/android/compose-samples/blob/bbf6ecaafc9541faa52d760ded69c4140f62b8fd/Jetcaster/glancewidget/build.gradle.kts#L59

### Describe the solution

Afflicted modules have been updated with the new API

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1488, opened 2024-10-09
[Jetcaster] Update gradle typesafe import

Xiryl:
reference to #1487

Afflicted modules have been updated with the new gradle typesafe import API


IanGClifton:
Thank you for the fix!

PR #1489, opened 2024-10-09
Jetcaster UI updates

simona-anomis:
Add to Jetcaster: shared elements, swipetodismiss, lazy list animations and removal of episode from queue

https://github.com/user-attachments/assets/95d6c54a-1f62-4bc9-8903-920296dd212b


Jetcaster/mobile/src/main/java/com/example/jetcaster/ui/home/category/PodcastCategory.kt
```diff
@@ -58,26 +63,44 @@ fun LazyListScope.podcastCategory(
...
-            episode = item.episode,
-            podcast = item.podcast,
-            onClick = navigateToPlayer,
-            onQueueEpisode = onQueueEpisode,
-            modifier = Modifier.fillParentMaxWidth()
-        )
+        val sharedTransitionScope = LocalSharedTransitionScope.current
+            ?: throw IllegalStateException("No SharedElementScope found")
+        val animatedVisibilityScope = LocalAnimatedVisibilityScope.current
+            ?: throw IllegalStateException("No SharedElementScope found")
```
riggaroo, line 83: You may want to do this a bit differently for this example since the Large screen versions wont need to have the scopes provided, you can probably build up a version that rather allows this to be nullable. 

We've explored a few different ways to do this - the least amount of code would probably be this kind of block:
```
.block {
    val animatedVisibilityScope = LocalNavAnimatedVisibilityScope.current
    if(this@with != null && animatedVisibilityScope != null)
        sharedElement(
            this@with.rememberSharedContentState(
                key = ChatSharedElementKey(
                    chatId = chat.chatWithLastMessage.id,
                    type = ChatSharedElementType.Image)
            ),
            animatedVisibilityScope = animatedVisibilityScope
        ) else Modifier
}
```
I also explored it with Jetsnack in creating a whole bunch of modifiers that are called `safe` versions of things. 
https://github.com/android/compose-samples/pull/1417/files#diff-db028e384f27d7d2dd08354e752447d4efddb1faaa8dafc3f43e7c73682e0270

Jetcaster/core/domain/src/main/java/com/example/jetcaster/core/model/EpisodeInfo.kt
```diff
@@ -43,3 +43,15 @@ fun Episode.asExternalModel(): EpisodeInfo =
...
+fun EpisodeInfo.asDaoModel(): Episode =
+    Episode(
+        uri = uri,
+        title = title,
+        subtitle = subTitle,
+        summary = summary,
+        author = author,
+        published = published,
+        duration = duration,
+        podcastUri = ""
```
riggaroo, line 56: Will this cause an issue? Should it be propagated from the Podcast info? Or maybe an episode needs a primary key link to a Podcast?

riggaroo (commented):
Looks good so far, would you be able to add some screen recordings when you have them?

Jetcaster/core/domain/src/main/java/com/example/jetcaster/core/model/EpisodeInfo.kt
simona-anomis: Changed

riggaroo (approved):
Lgtm! :)

yongsuk44:
With PR #1486 merged into the main branch, I have worked on updating the logic in this PR to use only Grid. Would it be alright to contribute these changes to the current branch?

- I have modified it so that the AnimationScope is not applied when the device is unfolded and the detail screen is open.
- I fixed an issue where SharedElementTransition was not applied in the library list.

**If you are working on this separately, feel free to skip this comment.**

https://github.com/user-attachments/assets/a2994e92-8fcd-47fb-965b-21c01c97272d

simona-anomis:
> With PR #1486 merged into the main branch, I have worked on updating the logic in this PR to use only Grid. Would it be alright to contribute these changes to the current branch?
> 
> * I have modified it so that the AnimationScope is not applied when the device is unfolded and the detail screen is open.
> * I fixed an issue where SharedElementTransition was not applied in the library list.
> 
> **If you are working on this separately, feel free to skip this comment.**
> 
>  fold.mp4

Hiya @yongsuk44  - thanks for your contributions! Just to clarify - have you already made these changes locally, but just haven't committed to this PR? If so, please let me know, as I was planning to do this today :) just so we don't duplicate the work. If you've already done the work, please feel free to commit to this branch. Thanks again!

yongsuk44:
@simona-anomis 
I've created a separate PR #1492 to merge into the 'jetcaster-ui-update' branch. 
Could you please take a look? 🙂

PR #1498, opened 2024-11-06
Jetcaster UI updates

simona-anomis:
Add to Jetcaster: shared elements, swipetodismiss, lazy list animations and removal of episode from queue


Jetcaster/mobile/src/main/java/com/example/jetcaster/ui/shared/EpisodeListItem.kt
```diff
@@ -66,35 +72,73 @@ fun EpisodeListItem(
...
+        }
+        when (dismissState.currentValue) {
+            SwipeToDismissBoxValue.EndToStart -> {
+                removeFromQueue(episode)
+            }
+
+            SwipeToDismissBoxValue.StartToEnd -> {
+            }
+
+            SwipeToDismissBoxValue.Settled -> {
```
mlykotom, line 141: if we're not caring about the other states, should we just use 
```
if (dismissState.currentValue == SwipeToDismissBoxValue.EndToStart) {
            removeFromQueue(episode)
        }
```

mlykotom (commented):
Some minor comments

mlykotom (approved):

Jetcaster/mobile/src/main/java/com/example/jetcaster/ui/player/PlayerScreen.kt
```diff
@@ -346,10 +354,16 @@ private fun PlayerContentRegular(
     onBackPress: () -> Unit,
     onAddToQueue: () -> Unit,
     playerControlActions: PlayerControlActions,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
 ) {
     val playerEpisode = uiState.episodePlayerState
     val currentEpisode = playerEpisode.currentEpisode ?: return
+
+    val sharedTransitionScope = LocalSharedTransitionScope.current
```
alexvanyo, line 362: Could this also be done in the non-regular `PlayerContent`s for getting the same shared element transition for tabletop mode? Or maybe this could be done within the `PlayerImage` itself to avoid duplication?

simona-anomis:
As discussed previously, this PR needs to wait for the 1.8 stable and shared elements grid support

PR #1504, opened 2024-11-12
Fix issue: tapping text input outside placeholder don't bring it to focus

magdalena-trivina:
Before
[tap on text input (before).webm](https://github.com/user-attachments/assets/6336821b-0abe-4090-ab06-a7dd8a756102)

After
[tap on text input (after).webm](https://github.com/user-attachments/assets/c70cebd5-b737-4e03-9925-bed61f14a3b9)


mlykotom (approved):
Thank you for your contribution!

PR #1505, opened 2024-11-12
[Jetchat] fix: white padding in profile screen in dark mode

magdalena-trivina:
There's a white padding in profile screen when in dark mode.

[Before]
White padding (before)

[After]
Padding (after)


mlykotom:
Pls update formatting :)

Jetchat/app/src/main/java/com/example/compose/jetchat/profile/Profile.kt
```diff
@@ -89,13 +88,13 @@ fun ProfileScreen(
...
             .nestedScroll(nestedScrollInteropConnection)
-            .systemBarsPadding()
     ) {
         Surface {
             Column(
                 modifier = Modifier
                     .fillMaxSize()
-                    .verticalScroll(scrollState),
+                    .verticalScroll(scrollState)
+                    .padding(16.dp),
```
riggaroo, line 97: This should continue to use systemBarsPadding() and not a hardcoded value.

PR #1510, opened 2024-11-14
[Jetcaster] Use one HomeScreen call inside SupportingPaneScaffold

### LINKED ISSUE #1508: [Bug]: Home screen scroll state lost when clicking on podcast details

### Sample app

Jetcaster

### What happened?

When opening a podcast details from the followed podcast list, the internal scroll state of the home screen is lost.

[Screen_recording_20241112_161132.webm](https://github.com/user-attachments/assets/88d6e6a1-ff39-41cd-ae6d-2b50b72c7e50)

This is likely due to the `HomeScreen` being called in two different places in the code, depending on this `if` statement:

https://github.com/android/compose-samples/blob/cb453e4cd91cffe3bf2acacc6256fb1397c86751/Jetcaster/mobile/src/main/java/com/example/jetcaster/ui/home/Home.kt#L325-L357

### Relevant logcat output

_No response_

### Code of Conduct

alexvanyo:
Fixes #1508 by only calling `HomeScreen` in a single source code location, in the main pane of `SupportingPaneScaffold`.

### DIFF
--- a/Jetcaster/mobile/src/main/java/com/example/jetcaster/ui/home/Home.kt
+++ b/Jetcaster/mobile/src/main/java/com/example/jetcaster/ui/home/Home.kt
@@ -324,6 +324,12 @@ private fun HomeScreenReady(
         SupportingPaneScaffold(
             value = navigator.scaffoldValue,
             directive = navigator.scaffoldDirective,
+            mainPane = {
+                HomeScreen(
+                    homeState = homeState,
+                    modifier = Modifier.fillMaxSize()
+                )
+            },
             supportingPane = {
                 val podcastUri = navigator.currentDestination?.content
                 if (!podcastUri.isNullOrEmpty()) {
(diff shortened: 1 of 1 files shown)


Jetcaster/mobile/src/main/java/com/example/jetcaster/ui/home/Home.kt
```diff
@@ -348,16 +343,16 @@ private fun HomeScreenReady(
...
-                        homeState = homeState,
-                        modifier = Modifier.fillMaxSize()
-                    )
-                },
-                modifier = Modifier.fillMaxSize()
-            )
-        }
+                }
+            },
+            mainPane = {
```
dturner, line 348: nit: Since you're refactoring this code, it would make more sense to move the `mainPane` parameter before `sortingPane`.

dturner (approved):

Jetcaster/mobile/src/main/java/com/example/jetcaster/ui/home/Home.kt
alexvanyo: Makes sense, re-ordered in https://github.com/android/compose-samples/pull/1510/commits/69ef8ced4ddad92635d7168a54173e269f79c730

PR #1525, opened 2025-01-08
Update and rename README.md to UPDATE README.md

zA3092:
Update realme.md


mlykotom (commented):
I believe this is a mistake PR? I'm closing for now, pls reopen in case you think it's a mistake from my end

PR #1540, opened 2025-03-18
[Reply] Enable predictive back 

trambui09:
Enable predictive back in the Reply sample


secondsun (approved):
LGTM

rebeccasg (approved):

alabiaga:
Once the targetSDK version becomes 36 then it'll be on by default right, we should have a clean up task to ensure we clean it from the manifest when that happens, a lint check will flag it in the manifest to remove as it'll be default on?

JolandaVerhoef (approved):

PR #1563, opened 2025-05-16
Fix spotless configuration and apply to all samples except Jetcaster

mlykotom:
Jetcaster is refactored in #1561, so the same config will be applied there, but I won't format it here, because it would create a million conflicts.


bentrengrove (approved):
Does this also fix the GitHub action not auto applying spotless?

mlykotom:
> Does this also fix the GitHub action not auto applying spotless?

Yeah it does. We had outdated configuration that didn't work already. The GH actions run properly `spotlessCheck` and the configuration is now correct in all samples.

PR #1571, opened 2025-05-26
Sajawal/jetchat/navigation

SajawalBashir:
**Migrated to Jetpack Compose Navigation**

### Summary
This PR refactors the app's navigation system to use **Jetpack Compose Navigation** instead of the previous View-based navigation approach.

### Changes Made
- Replaced legacy `NavController`/XML navigation with `androidx.navigation.compose.NavHost`
- Introduced `rememberNavController()` and route-based composable destinations
- Updated `MainActivity` and navigation logic to work with Compose-based screens
- Adjusted related ViewModel instantiations to match Compose lifecycle patterns

### Benefits
- Better integration with the existing Compose UI
- More concise and idiomatic navigation code
- Easier to test individual destinations and screen flows

### Notes
- Tested navigation flows between Conversation and Profile screens
- Removed unused legacy navigation artifacts (e.g., XML nav graphs)


bentrengrove:
Thanks for this contribution but unfortunately this sample is deliberately designed to show how to integrate Compose with Fragment navigation and changing this to Compose navigation would remove that. It is highly likely we will do this in the future as Nav3 becomes stable but I will leave it for now.

I encourage you to open an issue first before making large contributions to save yourself the trouble.
Thanks again.

PR #1590, opened 2025-07-04
[Jetlagged] Update icons and remove material icons dependency

JolandaVerhoef:
This PR replaces the Material Icons dependency with local drawable resources for all icons used in the JetLagged sample.
This involves:
- Adding new vector drawable XML files for each icon (`ic_bedtime.xml`, `ic_home.xml`, `ic_leaderboard.xml`, `ic_menu.xml`, `ic_settings.xml`, `ic_single_bed.xml`, `ic_watch.xml`).
- Updating composable functions in `HomeScreenCards.kt`, `JetLaggedHeader.kt`, and `JetLaggedDrawer.kt` to use `painterResource` for loading these local drawables instead of relying on the `Icons` object from the material-icons-extended library.
- Removing the `androidx.compose.material.iconsExtended` dependency from the app's `build.gradle.kts` file.
- Modifying the `Screen` enum in `JetLaggedDrawer.kt` to use `Int` (drawable resource ID) for the icon property instead of `ImageVector`.
- Changing the `icon` parameter type in `TwoLineInfoCard` from `ImageVector` to `Int`.


riggaroo (approved):

JetLagged/app/src/main/java/com/example/jetlagged/sleep/JetLaggedHeader.kt
jaya98777, line 1: 4fd993f6d67abf8f300fc1129e2266f87ed55d7e

PR #1596, opened 2025-07-17
Keep one CODEOWNERS

yrezgui:
There's currently two different CODEOWNERS. We're keeping only one


riggaroo (approved):

yrezgui:
I'm skipping one check as the build GitHub action expected but blocked as waiting

PR #1599, opened 2025-07-29
Fix drawer animation logic

cartland:
The drawer animation logic had a bug where it would not properly animate the drawer to the correct position.

This commit fixes the issue by adjusting the conditions that determine whether the drawer can reach the target position with decay. Specifically, it checks if `actualTargetX > 0f` instead of `targetDifference > 0f` and `actualTargetX == 0f` instead of `targetDifference < 0f`.


emreartan6-png:
#1599

PR #1646, opened 2026-02-05
Migrate JetNews to Navigation 3

claraf3:
Migration to Navigation 3 with simple deeplink implementation


claraf3:
@dturner Hi Don I wasn't sure how complicated you wanted the deepLink implementation to be. In a small app with only one deep link, I imagine the implementation would probably be similar to what I have right now - a small helper function to parse the postId.

JetNews/app/src/main/java/com/example/jetnews/ui/navigation/JetnewsNavigation.kt
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright 2021 The Android Open Source Project
```
dturner:
```suggestion
 * Copyright 2026 The Android Open Source Project
```

PR #1647, opened 2026-02-13
Migrate Jetsnack to Navigation 3

dturner:
I have migrated Jetsnack to use Navigation 3. 

Previously Jetsnack was using nested `NavHost`s. The outer `NavHost` held the Home and SnackDetail destinations and the Home destination had a `NavHost` containing the Feed, Search, Cart and Profile destinations. The Home destinations were all wrapped with a scaffold that displayed the bottom navigation bar. The SnackDetail destination was full screen without the bottom navigation bar. 

I have refactored this to use a single back stack and a single `NavDisplay` is used for all screens. This simplifies the code, but does introduce a caveat. 

In order to display the bottom navigation bar on the Home destinations the `NavDisplay` is wrapped with the scaffold. This means that the bottom bar is conditionally hidden when the SnackDetail navigation key is the last item in the back stack. This isn't particularly scalable. This may be better implemented as a scene decorator, although we're still figuring out best practices there.

Still to do: 

- When the moving between the Feed and Cart destinations the top bar disappears and reappears. It should stay put since it appears on both screens. 
- Shared element transitions between Cart and SnackDetail seem to be broken
- Implement deeplinks. 

Consider re-implementing this with nested NavDisplays.


Jetsnack/app/src/main/java/com/example/jetsnack/ui/components/Snacks.kt
```diff
@@ -268,10 +268,10 @@ private fun HighlightSnackItem(
 ) {
     val sharedTransitionScope = LocalSharedTransitionScope.current
         ?: throw IllegalStateException("No Scope found")
-    val animatedVisibilityScope = LocalNavAnimatedVisibilityScope.current
-        ?: throw IllegalStateException("No Scope found")
+    val animatedContentScope = LocalNavAnimatedContentScope.current
```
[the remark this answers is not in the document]
dturner: `LocalNavAnimatedContentScope.current` is non nullable

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -61,110 +66,101 @@ import com.example.jetsnack.ui.theme.JetsnackTheme
...
-                            onSnackSelected = jetsnackNavController::navigateToSnackDetail,
+                JetsnackScaffold(
+                    bottomBar = {
+                        val showBottomBar = backStack.last() !is SnackDetailKey
+
+                        AnimatedVisibility(visible = showBottomBar) {
+                            JetsnackBottomBar(
+                                tabs = HomeSections.entries.toTypedArray(),
+                                currentKey = backStack.findLast { it in HomeSections.entries.map { it.route } } ?: FeedKey,
+                                onItemClick = { navKey -> backStack.addHomeSection(navKey) },
```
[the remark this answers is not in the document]
dturner: Refactored

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -61,110 +67,101 @@ import com.example.jetsnack.ui.theme.JetsnackTheme
...
-                    composableWithCompositionLocal(
-                        route = MainDestinations.HOME_ROUTE,
-                    ) { backStackEntry ->
-                        MainContainer(
-                            onSnackSelected = jetsnackNavController::navigateToSnackDetail,
+                JetsnackScaffold(
+                    bottomBar = {
+                        val showBottomBar = backStack.last() !is SnackDetailKey
+
+                        AnimatedVisibility(visible = showBottomBar) {
```
riggaroo: This is changing the current animation of the BottomBar when the shared element transition is occuring, you need to tie it into the same transition that Nav uses in order to use the .animateEnterExit + renderInSharedTransitionScopeOverlay to happen with the current in progress shared element. 

Right now the top app bar is correctly transitioning, but the animation of the bottom one is not.  

If you can't access the NavScope here, then the animation itself should be fixed to slide up from the bottom and out from the bottom, so specify the enter/exit transiitons here on AnimatedVisibility, and remove the `animateEnterExit` modifier on the JetsnackBottomBar.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -61,110 +67,101 @@ import com.example.jetsnack.ui.theme.JetsnackTheme
...
+                            }
+                            entry<SnackDetailKey> { key ->
+                                SnackDetail(
+                                    key.snackId,
+                                    origin = key.origin,
+                                    upPress = { backStack.removeLastOrNull() },
+                                )
+                            }
+                        },
+                        transitionSpec = { transitionSpec },
```
riggaroo, line 157: I'm not sure why, but comparing between develop + this branch, the transition between the Feed and Cart composables looks a bit quicker, and doesn't seem like its performing a fadeIn or fadeOut. Doesn't look bad, I'm just wondering if there was a change missed here. 
Not blocking on this one :)

riggaroo (changes requested):
Overall looking good! Just the bottom bar animation needs fixing.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
dturner: Thanks, I definitely wouldn't have spotted this. I've updated the code to use the latter recommendation. We don't have access to LocalNavAnimatedContentScope here because we're not inside NavDisplay.

Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackApp.kt
```diff
@@ -61,110 +67,101 @@ import com.example.jetsnack.ui.theme.JetsnackTheme
...
+                    val transitionSpec = fadeIn(nonSpatialExpressiveSpring()) togetherWith
+                        fadeOut(nonSpatialExpressiveSpring())
+
+                    NavDisplay(
+                        backStack = backStack,
+                        onBack = { backStack.removeLastOrNull() },
+                        sharedTransitionScope = this@SharedTransitionLayout,
+                        entryProvider = entryProvider {
+                            entry<FeedKey> {
+                                Feed(
```
riggaroo, line 129: Within the Feed view, we should add a check on the TopAppBar to only slide in / slide out when its going towards snackdetail and coming back from snack detail, otherwise it now performs the animation between every screen transition, due to the new single instance of a navigation scope.

tiendz2003:
Any update here?

PR #1671, opened 2026-03-18
Add message reaction to chat - long press bubble to react emoji

shishupalbhati:


riggaroo:
Hello, you are free to build up these features on your own fork of the repository, however we are not looking for extra features on the apps at this point in time that don't demonstrate new concepts in Compose - as these samples are for learning, we like to keep them lean. Thanks a bunch and happy learning Compose!

PR #1675, opened 2026-03-20
Migrate Jetnews to Navigation 3

bsagmoe:
## Key Changes
* Created `NavKey` objects and classes to replace the `JetnewsDestinations` object.
* Refactored the home route to only be responsible for displaying the list of posts and moved the post detail UI and logic into its own route.
* Implemented a `ListDetailSceneStrategy` to display the post detail view alongside the list of posts if the window width size class is at least expanded.
* Added `NavigationState` and `Navigator` classes and helpers for storing and updating the app's navigation state. 
* Replaced `JetnewsNavGraph` with `JetnewsNavDisplay`.
* Removed Navigation 2 dependencies 
* Implemented support for deeplinks for the home page, post detail page, and interest page.
* Improved the `app_opensArticle` (now `app_opensPost`) test so that it now actually tests that the post content is displayed.


JetNews/app/src/androidTest/java/com/example/jetnews/JetnewsTests.kt
```diff
@@ -55,17 +54,10 @@ class JetnewsTests {
 
     @Test
     fun app_opensArticle() {
-
-        println(composeTestRule.onRoot().printToString())
-        composeTestRule.onAllNodes(hasText(manuel.name, substring = true))[0].performClick()
-
-        println(composeTestRule.onRoot().printToString())
-        try {
-            composeTestRule.onAllNodes(hasText("3 min read", substring = true))[0].assertExists()
```
bsagmoe, line 64: This text also shows up in the list screen, so the test didn't actually test that the article screen was opened to the correct article.

JetNews/app/src/main/java/com/example/jetnews/ui/post/PostScreen.kt
```diff
@@ -194,6 +227,25 @@ private fun TopAppBar(
...
+        modifier = modifier.padding(end = 16.dp),
+    ) {
+        Row(Modifier.padding(horizontal = 8.dp)) {
+            FavoriteButton(onClick = { /* Functionality not available */ })
+            BookmarkButton(isBookmarked = isFavorite, onClick = onToggleFavorite)
+            ShareButton(onClick = onSharePost)
+            TextSettingsButton(onClick = { /* Functionality not available */ })
+        }
+    }
+}
```
bsagmoe, line 359: Moved here from HomeScreens.kt

JetNews/app/src/main/java/com/example/jetnews/ui/MainActivity.kt
```diff
@@ -20,21 +20,41 @@ import android.os.Bundle
...
-            val widthSizeClass = calculateWindowSizeClass(this).widthSizeClass
-            JetnewsApp(appContainer, widthSizeClass)
+            val initialDeepLinkBackStack = remember { intent.data?.handleDeepLink() }
+            var isOpenedByDeepLink by rememberSaveable { mutableStateOf(initialDeepLinkBackStack != null) }
+            var initialBackStack by rememberInitialBackStack(initialDeepLinkBackStack ?: listOf(HomeKey))
+
+            NewIntentEffect { newIntent ->
+                newIntent.data?.handleDeepLink()?.let {
+                    isOpenedByDeepLink = true
+                    initialBackStack = it
```
dturner: Suggest using a named lambda parameter here instead of `it` such as `syntheticBackStack` so it's clear what `handleDeepLink` is returning.

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsNavDisplay.kt
```diff
@@ -0,0 +1,128 @@
...
+        )
+    }
+
+    val navEntries = navigationState.toDecoratedEntries(entryProvider = entryProvider)
+
+    val listDetailSceneStrategy = rememberListDetailSceneStrategy<NavKey>(isExpandedScreen)
+
+    SharedTransitionLayout {
+        Surface {
+            val sceneState = rememberSceneState(
```
dturner, line 91: What's the reason for using `rememberSceneState` instead of the `NaviDisplay` that takes an `entryProvider`?

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsNavDisplay.kt
```diff
@@ -0,0 +1,128 @@
...
+            )
+
+            NavDisplay(
+                sceneState = sceneState,
+                navigationEventState = navigationEventState,
+                predictivePopTransitionSpec = {
+                    unveilIn() togetherWith scaleOut(
+                        targetScale = .8f,
+                        transformOrigin = when (it) {
+                            NavigationEvent.EDGE_LEFT -> TransformOrigin(1 - PIVOT_FRACTION_OFFSET, .5f)
```
dturner, line 122: Oh, I see. If you know the edge where predictive back is being initiated from, you can move the outgoing screen away from where the user's finger is. This is really cool....but I'm wondering whether additional features should be added in this PR or in a separate PR if this is now best practice. Leave it up to you. If leaving in this PR then add a comment to explain this behavior as it took me a while to figure it out.

dturner: Additionally, if this is making the implementation more complex, I wonder whether this would be better placed in a recipe. Again, leave it up to you.

JetNews/app/src/main/java/com/example/jetnews/ui/components/AppNavRail.kt
```diff
@@ -49,14 +51,14 @@ fun AppNavRail(currentRoute: String, navigateToHome: () -> Unit, navigateToInter
     ) {
         Spacer(Modifier.weight(1f))
         NavigationRailItem(
-            selected = currentRoute == JetnewsDestinations.HOME_ROUTE,
+            selected = currentRoute is HomeKey,
```
dturner: You could reduce the duplication here by storing the top level navigation keys in a collection that associates each key with its metadata (icon etc). Example: https://github.com/android/nav3-recipes/blob/main/app/src/main/java/com/example/nav3recipes/multiplestacks/MultipleStacksActivity.kt#L59

JetNews/app/src/main/java/com/example/jetnews/ui/navigation/Navigator.kt
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package com.example.jetnews.ui.navigation
+
+import com.example.jetnews.ui.home.HomeKey
+import com.example.jetnews.ui.interests.InterestsKey
+import com.example.jetnews.ui.post.PostKey
+
+class Navigator(val state: NavigationState) {
+    fun toHome() {
```
dturner: Currently `Navigator` is tightly coupled with each of the features: 

- it needs a `to` method to navigate to that feature 
- it's dependent on that feature's keys. 

To loosen this, you could define the behavior based on whether the key is a top level destination or not. 
Very basic example that doesn't handle the case where you navigate to a key that's already in the back stack: https://github.com/android/nav3-recipes/blob/main/app/src/main/java/com/example/nav3recipes/multiplestacks/Navigator.kt#L25

JetNews/app/src/main/java/com/example/jetnews/ui/post/PostRoute.kt
```diff
@@ -0,0 +1,124 @@
...
+            onBack = onBack,
+            onToggleFavorite = postViewModel::toggleFavorite,
+            onScroll = postViewModel::onScroll,
+        )
+    }
+}
+
+@OptIn(FlowPreview::class)
+@Composable
+fun PostRoute(
```
dturner: nit: The term "route" is usually associated with Nav2 terminology. I'd change to `PostScreen` which then fits with the convention of having a stateful and stateless versions of a screen composable. If making this change, apply to other `Route`s in the codebase.

JetNews/app/src/main/java/com/example/jetnews/deeplink/JetnewsDeeplinks.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.example.jetnews.deeplink.util.DeepLinkMatcher
+import com.example.jetnews.deeplink.util.DeepLinkPattern
+import com.example.jetnews.deeplink.util.DeepLinkRequest
+import com.example.jetnews.deeplink.util.KeyDecoder
+import com.example.jetnews.ui.home.HomeKey
+import com.example.jetnews.ui.interests.InterestsKey
+import com.example.jetnews.ui.navigation.DeepLinkKey
+import com.example.jetnews.ui.post.PostKey
+
+val HomeDeepLinkPattern = DeepLinkPattern(
```
dturner: Does it make sense to have deeplink URIs defined in one place (here), or defined next to the nav keys?

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsNavDisplay.kt
bsagmoe: To properly handle back vs. up, you can't  (currently) use `NavDisplay`'s built in back handling, so you have to set up your own `NavigationBackHandler` and use the `NavDisplay` overlay that takes a `navigationEventState` parameter. Which just also requires that you pass a `sceneState`. I've got some 0th dev feedback to submit about making it easier to override the built-in back handling without having to use this lowest level `NavDisplay`.

JetNews/app/src/main/java/com/example/jetnews/ui/JetnewsNavDisplay.kt
bsagmoe: I feel like it's a small enough thing that also demonstrates a benefit of moving to Nav 3, so I think I'll keep it!

JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeScreens.kt
```diff
@@ -82,15 +84,104 @@ import androidx.compose.ui.text.style.TextAlign
...
+            homeViewModel = homeViewModel,
+            isExpandedScreen = isExpandedScreen(),
+            openDrawer = openDrawer,
+            navigateToPost = navigateToPost,
+        )
+    }
+}
+
+/**
+ * Displays the Home route.
```
dturner: nit: route -> screen

JetNews/app/src/main/java/com/example/jetnews/ui/interests/InterestsScreen.kt
```diff
@@ -96,6 +104,61 @@ enum class Sections(@StringRes val titleResId: Int) {
...
+        InterestsScreen(
+            interestsViewModel,
+            isExpandedScreen = isExpandedScreen(),
+            openDrawer = openDrawer,
+        )
+    }
+}
+
+/**
+ * Stateful composable that displays the Navigation route for the Interests screen.
```
dturner: nit: displays the Interests screen.

JetNews/app/src/main/java/com/example/jetnews/ui/navigation/NavigationState.kt
```diff
@@ -0,0 +1,108 @@
...
+fun rememberInitialBackStack(backStack: List<NavKey>): MutableState<List<NavKey>> {
+    return rememberSerializable(
+        serializer = MutableStateSerializer(ListSerializer(NavKeySerializer())),
+    ) {
+        mutableStateOf(backStack)
+    }
+}
+
+@Composable
+fun rememberNavigationState(mainTopLevelRoute: NavKey, topLevelRoutes: Set<NavKey>, initialBackStack: List<NavKey>): NavigationState {
```
dturner: Is there a way of deriving `mainTopLevelRoute` from `initialBackStack` to avoid the need for the parameter? If not, it might be better named `homeRoute` or `startRoute` as `main` doesn't tell me much.

dturner: nit: Route -> Key here and throughout

JetNews/app/src/main/java/com/example/jetnews/ui/navigation/NavigationState.kt
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    return remember(mainTopLevelRoute, topLevelRoute, topLevelRoutes, backStacks) {
+        NavigationState(mainTopLevelRoute, topLevelRoute, topLevelRoutes, backStacks)
+    }
+}
+
+class NavigationState(
+    val mainTopLevelRoute: NavKey,
```
dturner: same comments apply here

dturner:
Added a few nits. Overall LGTM.

JetNews/app/src/main/java/com/example/jetnews/ui/navigation/NavigationState.kt
bsagmoe: Renamed to `primaryTopLevelKey` and added some KDocs for the `NavigationState` class explaining its purpose.

JoseAlcerreca:
Sorry, wrong account 🤖

